PR #19, opened 2023-05-01
Refactor

cretz:
## What was changed

Refactored to support a bit more simplicity and room for growth. Specifically:

* Moved `/cmd/omes/main.go` to `/cmd/main.go` - no benefit to being in `omes` at this point
* Moved `/components/*` and `/scenario` and `/executors` packages all to one `/loadgen` package - from experience, sprawling these packages out in Go can cause many headaches later
* Moved `/runner` to `/loadgen/runner` - Keeping all non-main Go code under `/loadgen` keeps the top dir free from a bunch of separate packages
* Added `go.mod` to `/workers/go` - Workers need to be self-contained projects
* Moved `pyproject.toml` to `/workers/python` - Workers need to be self-contained projects
* Updated SDK and removed `/runner/devserver` - It is now in the latest Go SDK
* Removed cloud run that requires secret access from CI. We want to encourage forks on this repo.

After this refactor and once https://github.com/temporalio/features/pull/269 is merged, there will be another refactor to use it.

(note, this is the continuation of #18 which was made while this repo was still private and could not be updated)


cretz:
Ug, CI here is built to _require_ secrets which means you can't work off a forked branch. Putting in draft while I remove the cloud config and try https://github.com/temporalio/setup-temporal instead.

cretz:
Had to fix CI, now marked as ready for review again. I removed the cloud secret-access requirement for CI. IMO we should encourage people to use forks for their PRs, but requiring access to secrets disallows using forks.

(branches on this repo itself make for stale branches that stick around forever and/or have to be cleaned up...I'd disallow them if I could)

.github/workflows/ci.yml
```diff
@@ -17,30 +17,11 @@ jobs:
         with:
           go-version: "^1.19"
 
-      - run: go build ./cmd/omes
+      - run: go build -o temporal-omes ./cmd
```
bergundy, line 20: No need to customize the output now.

bergundy (approved):

.github/workflows/ci.yml
cretz: Yes you do, or it'd be called `cmd`

PR #20, opened 2023-05-02
Refactoring with prepared-worker support

cretz:
## What was changed

This is a followup to #20. Specifically:

* Added `prepare-worker` command
* Changed `run` command to `run-scenario`
* Changed `all-in-one` command to `run-scenario-with-worker`
* Changed `--start-dev-server` option to `--embedded-server`
* Moved the runner code that's only for CLI to `/cmd` and to the specific commands needed
* Move shared code that only supports CLIs and is not needed by scenario authors to `/cmd/cmdoptions`
* Fixed issues with duration errors reporting after context complete
* Fixed issues where interrupt was not gracefully stopping runs
* Downgraded required Python version in pyproject.toml to support sdkbuild's minimum Python version


cmd/prepare_worker.go
```diff
@@ -0,0 +1,161 @@
...
+		Use:   "prepare-worker",
+		Short: "Build worker ready to run",
+		Run: func(cmd *cobra.Command, args []string) {
+			if _, err := b.build(cmd.Context()); err != nil {
+				b.logger.Fatal(err)
+			}
+		},
+	}
+	b.addCLIFlags(cmd.Flags())
+	cmd.MarkFlagRequired("dir")
```
cretz, line 30:
```suggestion
	cmd.MarkFlagRequired("dir-name")
```

go.mod
```diff
@@ -33,18 +34,20 @@ require (
...
 	golang.org/x/text v0.8.0 // indirect
-	golang.org/x/time v0.1.0 // indirect
+	golang.org/x/time v0.3.0 // indirect
 	google.golang.org/genproto v0.0.0-20230322174352-cde4c949918d // indirect
 	google.golang.org/grpc v1.54.0 // indirect
 	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace go.temporal.io/features => github.com/temporalio/features v0.0.0-20230501123624-d8e75548b643
```
cretz, line 53: I wish we hadn't used a vanity name in features repo. The Go module name there should be `github.com/temporalio/features` instead of `go.temporal.io/features`. Not sure if I should open an issue or just ignore it.

loadgen/shared_iterations_executor.go
```diff
@@ -24,6 +24,8 @@ type SharedIterationsExecutor struct {
 	Duration time.Duration
 	// Function to execute a single iteration of this scenario.
 	Execute func(ctx context.Context, run *Run) error
+	// If true, will also include errors even after context complete/
```
cretz, line 27:
```suggestion
	// If true, will also include errors even after context complete.
```

workers/go/worker/worker.go
cretz, line 1: Need to delete `workers/go/main.go`

cmd/prepare_worker.go
cretz: Fixed

loadgen/shared_iterations_executor.go
cretz: Fixed

workers/go/worker/worker.go
cretz: Fixed

cmd/prepare_worker.go
```diff
@@ -0,0 +1,161 @@
...
+}`,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed preparing: %w", err)
+	}
+	return prog, nil
+}
+
+func (b *workerBuilder) buildPython(ctx context.Context, baseDir string) (sdkbuild.Program, error) {
+	// If version not provided, try to read it from pyproject.toml
```
bergundy, line 115: What's the rationale here? Why not require version?

cmd/prepare_worker.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return prog, nil
+}
+
+func rootDir() string {
+	_, currFile, _, _ := runtime.Caller(0)
+	return filepath.Dir(filepath.Dir(currFile))
+}
+
+func normalizeLangName(lang string) (string, error) {
```
bergundy, line 152: nit: Might as well put this as a helper in sdkbuild.

cmd/prepare_worker.go
cretz: We don't in the features repo and we don't for Go, so why require it here?

bergundy (approved):

cmd/prepare_worker.go
cretz: No, this is not the same implementation as features repo. Not only do we not support all langs here (yet), but we normalize here to the full name matching the worker folder whereas features repo normalizes to the file extension. They are different use cases.

cmd/prepare_worker.go
bergundy: Hmm.. I thought we did in both

go.mod
cretz: It will when you try to upgrade the `features` lib for updates in `sdkbuild`. This `replace` statement wouldn't have to exist if we made the module name the location where the module can be accessed at.

PR #21, opened 2023-05-03
Basic worker image building

cretz:
## What was changed

Added `build-worker-image` command. Intentionally don't have a bunch of `git` commands or support `push`ing at this time, though we can add later if we must (though we may find cloning a repo and pushing a docker image to be better suited for the environment and not the CLI).

### DIFF
--- a/cmd/build_worker_image.go
+++ b/cmd/build_worker_image.go
@@ -1,0 +1,177 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/temporalio/omes/cmd/cmdoptions"
+	"go.uber.org/zap"
+	"golang.org/x/mod/semver"
+)
+
+func buildWorkerImageCmd() *cobra.Command {
+	var b workerImageBuilder
+	cmd := &cobra.Command{
+		Use:   "build-worker-image",
+		Short: "Build an image",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := b.build(cmd.Context()); err != nil {
+				b.logger.Fatal(err)
+			}
+		},
+	}
+	b.addCLIFlags(cmd.Flags())
+	cmd.MarkFlagRequired("language")
+	cmd.MarkFlagRequired("version")
+	return cmd
+}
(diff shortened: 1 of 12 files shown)


cmd/build_worker_image.go
```diff
@@ -0,0 +1,177 @@
...
+		if len(b.tags) == 0 {
+			return fmt.Errorf("at least one tag required for path version")
+		} else if s, err := os.Stat(b.version); err != nil {
+			return fmt.Errorf("invalid path version: %w", err)
+		} else if !s.IsDir() {
+			return fmt.Errorf("invalid path version: must be dir")
+		} else if !filepath.IsLocal(b.version) {
+			return fmt.Errorf("invalid path version: must be beneath this dir")
+		}
+		// We have to copy the entire
```
Sushisource, line 83: Copy the entire!

Sushisource (approved):

cmd/build_worker_image.go
cretz: Yup, copy the entire everything except the entire comment out of my head

cmd/run_scenario.go
```diff
@@ -47,6 +49,7 @@ func (r *scenarioRunner) addCLIFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.runID, "run-id", "", "Run ID for this run")
 	fs.IntVar(&r.iterations, "iterations", 0, "Iterations for the scenario (cannot be provided with duration)")
 	fs.DurationVar(&r.duration, "duration", 0, "Duration for the scenario (cannot be provided with iteration)")
+	fs.DurationVar(&r.waitForServer, "wait-for-server", 0, "Duration to try to connect to server before failing")
```
bergundy, line 52: `connect-timeout` maybe? `wait-for-server` sounds to me like a condition (boolean) not a duration.
I'd accept another name where it's clear that this is a duration.

bergundy (approved):
Nice, super simple.
I wish we could move some of this into features but I can see that there's a bit less value in this case.

cmd/build_worker_image.go
cretz: :+1: Will fix in next PR coming shortly after this one.

cmd/run_scenario.go
cretz: :+1: Will fix in next PR coming shortly after this one.

PR #22, opened 2023-05-05
Many activity/child workflow

### LINKED ISSUE #10: Scenario request: Lots of workflows with lots of children/activities
### Describe the scenario you'd like

Run `<workflow-count>` workflows that each run either `<child-count>` children and/or `<activity-count>` activities.

The variables can be set based on good numbers that stress the environment.

cretz:
## What was changed

* Reworked execution concurrency to be simpler and more idiomatic
* Added support for kitchen sink executor
* Updated kitchen sink workflow to support action concurrency
* Added support for arbitrary user options for scenarios
* Added `workflow_with_many_actions` scenario that accepts activity and child workflow count options
* Added `list-scenarios` command
* Updated README

## Checklist

1. Closes #10


README.md
```diff
@@ -23,51 +23,37 @@ There's no need to install anything to use this, it's a self-contained Go projec
 
 ### Define a scenario
 
-Scenarios are defined using plain Go code. They are located in the [scenarios](./scenarios/) folder.
+Scenarios are defined using plain Go code. They are located in the [scenarios](./scenarios/) folder. There are already
```
bergundy, line 26: Long term we'll want this OSS repo to not contain any baked in scenarios, right?

cmd/list_scenarios.go
```diff
@@ -0,0 +1,43 @@
...
+				if iface, _ := scen.Executor.(loadgen.HasDefaultConfiguration); iface != nil {
+					config := iface.GetDefaultConfiguration()
+					config.ApplyDefaults()
+					defaultConfigDesc = "\n    Default configuration:"
+					if config.Iterations != 0 {
+						defaultConfigDesc += fmt.Sprintf("\n        Iterations: %v", config.Iterations)
+					}
+					if config.Duration != 0 {
+						defaultConfigDesc += fmt.Sprintf("\n        Duration: %v", config.Duration)
+					}
```
bergundy, line 29: Wouldn't it be better to put a Describe method on the executor interface?

cmd/run_scenario.go
```diff
@@ -33,23 +34,27 @@ func runScenarioCmd() *cobra.Command {
...
 
 func (r *scenarioRunner) addCLIFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.scenario, "scenario", "", "Scenario name to run")
 	fs.StringVar(&r.runID, "run-id", "", "Run ID for this run")
-	fs.IntVar(&r.iterations, "iterations", 0, "Iterations for the scenario (cannot be provided with duration)")
-	fs.DurationVar(&r.duration, "duration", 0, "Duration for the scenario (cannot be provided with iteration)")
-	fs.DurationVar(&r.waitForServer, "wait-for-server", 0, "Duration to try to connect to server before failing")
+	fs.IntVar(&r.iterations, "iterations", 0, "Override default iterations for the scenario (cannot be provided with duration)")
+	fs.DurationVar(&r.duration, "duration", 0, "Override duration for the scenario (cannot be provided with iteration)")
+	fs.IntVar(&r.maxConcurrent, "max-concurrent", 0, "Override max-concurrent for the scenario")
```
bergundy, line 55: If we really want to make the executor generic, these should be supplied via `--option`, I never got around to restructuring this.

README.md
cretz: Not from my POV. This should have all scenarios we want to run as a company. This is a tool not a library IMO; we don't want the compatibility guarantees that come with the latter.

loadgen/generic_executor.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.temporal.io/sdk/client"
+	"go.uber.org/zap"
+)
+
+type GenericExecutor struct {
```
bergundy, line 12: Not loving this name, "shared iterations" is more descriptive than "generic" and we will likely want different executors with different behaviors.

cretz: I don't think the common one needs to have a descriptive name. That is not very Go like. When we get to that other one, we can rename.

loadgen/generic_executor.go
```diff
@@ -0,0 +1,136 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	if g.config.Duration > 0 {
+		ctx, cancel = context.WithTimeout(ctx, g.config.Duration)
+	}
+	defer cancel()
+
+	startTime := time.Now()
+	var runErr error
+	doneCh := make(chan error)
+	var currentlyRunning int
```
bergundy, line 77: nit: I'd name this `numCurrentlyRunning` to clarify this is a count not a toggle state.

loadgen/generic_executor.go
```diff
@@ -0,0 +1,136 @@
...
+			waitOne()
+			// Exit loop if error
+			if runErr != nil || ctx.Err() != nil {
+				break
+			}
+		}
+		// Run concurrently
+		g.logger.Debugf("Running iteration %v", i)
+		currentlyRunning++
+		run := &Run{
```
bergundy, line 102: I think the name `Run` came from my original code but I don't love it. Got any better ideas?

loadgen/generic_executor.go
```diff
@@ -0,0 +1,136 @@
...
+			ScenarioName:    g.options.ScenarioName,
+			IterationInTest: i + 1,
+			Logger:          g.logger.With("iteration", i),
+			ID:              g.options.RunID,
+		}
+		go func() {
+			startTime := time.Now()
+			err := g.executor.Execute(ctx, run)
+			// Only log/wrap/send to channel if context is not done
+			if ctx.Err() == nil {
```
bergundy, line 113: I'd switch to eager return and remove the nesting

loadgen/generic_executor.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, fmt.Errorf("invalid scenario: iterations and duration are mutually exclusive")
+	}
+
+	return run, nil
+}
+
+// Run a scenario.
+// Spins up coroutines according to the scenario configuration.
+// Each coroutine runs the scenario Execute method in a loop until the scenario duration or max iterations is reached.
+func (g *genericRun) Run(ctx context.Context) error {
```
bergundy, line 67: Nice, this is much cleaner than what I wrote

loadgen/kitchensink/kitchen_sink.go
```diff
@@ -2,21 +2,53 @@ package kitchensink
...
+	Sleep                *SleepAction           `json:"sleep"`
+	QueryHandler         *QueryHandlerAction    `json:"query_handler"`
+	Signal               *SignalAction          `json:"signal"`
+	ExecuteActivity      *ExecuteActivityAction `json:"execute_activity"`
+	ExecuteChildWorkflow *ExecuteChildWorkflow  `json:"execute_child_workflow"`
+	NestedActionSet      *ActionSet             `json:"nested_action_set"`
 }
 
+// NopActivityName is the name of the noop activity
+const NopActivityName = "noop"
```
bergundy, line 47:
```suggestion
const NoopActivityName = "noop"
```

loadgen/kitchensink/kitchen_sink.go
```diff
@@ -2,21 +2,53 @@ package kitchensink
...
+
+// NopActionExecuteActivity is an action that executes a noop activity
+var NopActionExecuteActivity = &Action{ExecuteActivity: &ExecuteActivityAction{Name: NopActivityName}}
```
bergundy, line 50:
```suggestion
var NoopActionExecuteActivity = &Action{ExecuteActivity: &ExecuteActivityAction{Name: NoopActivityName}}
```

loadgen/generic_executor.go
cretz: Not what a Go programmer would usually say, heh. You're lucky it's not `c` :-)

scenarios/workflow_with_many_actions.go
```diff
@@ -0,0 +1,46 @@
...
+		Description: "Each iteration executes a single workflow with a number of child workflows and/or activities. " +
+			"Additional options: children-per-workflow (default 30), activities-per-workflow (default 30).",
+		Executor: loadgen.KitchenSinkExecutor{
+			PrepareWorkflowParams: func(ctx context.Context, opts loadgen.RunOptions, params *kitchensink.WorkflowParams) error {
+				// We want these executed concurrently
+				params.ActionSet.Concurrent = true
+
+				// Get options
+				children := opts.ScenarioOptionInt("children-per-workflow", 30)
+				activities := opts.ScenarioOptionInt("activities-per-workflow", 30)
```
bergundy, line 21: Would be nice if the options were more structured so they can be shown to a user describing the scenario.

loadgen/generic_executor.go
cretz: Hardly enough of a difference to even comment on much less change, heh

loadgen/kitchensink/kitchen_sink.go
cretz: No, e.g. https://pkg.go.dev/io#NopCloser

loadgen/kitchensink/kitchen_sink.go
cretz: No, e.g. https://pkg.go.dev/io#NopCloser

workers/python/kitchen_sink.py
```diff
@@ -81,38 +99,64 @@ def workflow_query(arg: str) -> str:
...
-            while not self._pending_actions.empty():
-                action = await self._pending_actions.get()
-                should_return, return_value = await self.handle_action(input, action)
+        if input.action_set_signal:
+            action_set_queue: asyncio.Queue[ActionSet] = asyncio.Queue()
+
+            def enqueue(item: ActionSet) -> None:
+                action_set_queue.put_nowait(item)
+
+            workflow.set_signal_handler(input.action_set_signal, enqueue)
```
bergundy, line 120: Could rewrite to use a lambda

bergundy: Scratch that, it'd be hard to type it

workers/python/kitchen_sink.py
```diff
@@ -81,38 +99,64 @@ def workflow_query(arg: str) -> str:
...
+
+    async def handle_action_set(
+        self, input: WorkflowParams, action_set: ActionSet
+    ) -> Tuple[bool, Any]:
+        should_return = False
+        return_value = None
+        # If these are non-concurrent, just execute and return if requested
+        if not action_set.concurrent:
+            for action in action_set.actions:
+                should_return, return_value = await self.handle_action(input, action)
```
bergundy, line 139: Doesn't seem like there that much value in having the actions return anything.
Thinking of just dropping this functionality.
Also would be nice to be able to handle a stream of signals instead of batching but I think this is good enough for now.

cretz: Return value is fine, but we can remove later if we need. We do handle a stream of signals unless I misunderstand.

workers/python/kitchen_sink.py
```diff
@@ -140,25 +184,44 @@ async def handle_action(
...
-            )
-            await queue.get()
-            queue.task_done()
+            signal_event = asyncio.Event()
+
+            def signal_handler(arg: Optional[Any] = None) -> None:
+                signal_event.set()
+
+            workflow.set_signal_handler(action.signal.name, signal_handler)
+            await signal_event.wait()
```
bergundy, line 193: Is there a reason you don't return in this branch?

cretz: Same reason I don't in sleep and we shouldn't in many other places we do. No need to repeat `return (False, None)` in each arm.

workers/python/kitchen_sink.py
bergundy: We block until the action or set is complete before processing the next signal, streaming would allow receiving more work while processing is in progress.

loadgen/kitchensink/kitchen_sink.go
bergundy: First time I'm seeing no-op shortened like that

README.md
bergundy: Hmm.. when we last spoke I thought we agreed otherwise.
But I'm totally fine starting out like this and breaking it out to be extensible eventually.

scenarios/workflow_with_many_actions.go
bergundy: Not blocking

loadgen/generic_executor.go
bergundy: 🤷

loadgen/generic_executor.go
bergundy: Definitely lucky it's not C

loadgen/generic_executor.go
bergundy: Sure, can postpone AFAIC

cmd/run_scenario.go
bergundy: Can wait for when there's an executor that doesn't take these params.

workers/python/kitchen_sink.py
cretz: Ah, I meant the `action_set_signal`. But yes, when we need that functionality we can add it.

We need to be very careful guessing at things we might need. This needs to be an evolving system, not one based on guesses. The only reason this has the features it does now is because I needed them for integration testing the Python SDK, no more.

loadgen/kitchensink/kitchen_sink.go
cretz: It's pretty common in assembly representations

README.md
cretz: I probably contradicted myself. But at this point, we really need to shoot for simplicity and necessity and just be willing to evolve.

PR #23, opened 2023-05-17
Multi-task-queue Scenario

### LINKED ISSUE #13: Scenario request: Namespace with lots of task queues
### Describe the scenario you'd like

Create `<workflow-count>` workflows on `<task-queue-count>` separate task queues (and workers).

The variables can be set based on good numbers that stress the environment.

cretz:
## What was changed

Added a new `workflow_on_many_task_queues` scenario and support for running many workers in a single worker process.

For example to run scenario with 500 workflows across 20 task queues in Go using embedded server and running worker at the same time, you'd run

```
go run ./cmd run-scenario-with-worker --scenario workflow_on_many_task_queues --language go --embedded-server --iterations 500 --option task-queue-count=20 --task-queue-suffix-index-end 19
```

Of course this could be broken out to just, say, 2 workers running 10 each against existing server:

```
go run ./cmd run-worker --server-address 1.2.3.4:7233 --scenario workflow_on_many_task_queues --language go --run-id my-run --task-queue-suffix-index-end 9
```
```
go run ./cmd run-worker --server-address 1.2.3.4:7233 --scenario workflow_on_many_task_queues --language go --run-id my-run --task-queue-suffix-index-start 10 --task-queue-suffix-index-end 19
```
```
go run ./cmd run-scenario --scenario workflow_on_many_task_queues --iterations 500 --option task-queue-count=20
```

Also works for Python.

## Checklist

1. Closes #13

### DIFF
--- a/scenarios/workflow_on_many_task_queues.go
+++ b/scenarios/workflow_on_many_task_queues.go
@@ -1,0 +1,33 @@
+package scenarios
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/temporalio/omes/loadgen"
+	"github.com/temporalio/omes/loadgen/kitchensink"
+)
+
+func init() {
+	loadgen.MustRegisterScenario(loadgen.Scenario{
+		Description: "Each iteration executes a single workflow on one of the task queues. " +
+			"Workers must be started with --task-queue-suffix-index-end as one less than task queue count here. " +
+			"Additional options: task-queue-count (required).",
+		Executor: loadgen.KitchenSinkExecutor{
+			WorkflowParams: kitchensink.NewWorkflowParams(kitchensink.NopActionExecuteActivity),
+			PrepareWorkflowParams: func(ctx context.Context, opts loadgen.RunOptions, params *kitchensink.WorkflowParams) error {
+				// Require task queue count
+				if opts.ScenarioOptionInt("task-queue-count", 0) == 0 {
+					return fmt.Errorf("task-queue-count option required")
+				}
+				return nil
+			},
+			UpdateWorkflowOptions: func(ctx context.Context, run *loadgen.Run, options *loadgen.KitchenSinkWorkflowOptions) error {
+				// Add suffix to the task queue based on modulus of iteration
+				options.StartOptions.TaskQueue +=
+					fmt.Sprintf("-%v", run.IterationInTest%run.RunOptions.ScenarioOptionInt("task-queue-count", 0))
+				return nil
+			},
+		},
+	})
+}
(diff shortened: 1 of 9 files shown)


scenarios/workflow_on_many_task_queues.go
```diff
@@ -0,0 +1,33 @@
...
+		Executor: loadgen.KitchenSinkExecutor{
+			WorkflowParams: kitchensink.NewWorkflowParams(kitchensink.NopActionExecuteActivity),
+			PrepareWorkflowParams: func(ctx context.Context, opts loadgen.RunOptions, params *kitchensink.WorkflowParams) error {
+				// Require task queue count
+				if opts.ScenarioOptionInt("task-queue-count", 0) == 0 {
+					return fmt.Errorf("task-queue-count option required")
+				}
+				return nil
+			},
+			UpdateWorkflowOptions: func(ctx context.Context, run *loadgen.Run, options *loadgen.KitchenSinkWorkflowOptions) error {
```
bergundy, line 25: This is roundabout way to do this, I'd use the generic executor and have all of the logic inline instead of mutating the params every iteration.

scenarios/workflow_on_many_task_queues.go
```diff
@@ -0,0 +1,33 @@
...
+func init() {
+	loadgen.MustRegisterScenario(loadgen.Scenario{
+		Description: "Each iteration executes a single workflow on one of the task queues. " +
+			"Workers must be started with --task-queue-suffix-index-end as one less than task queue count here. " +
+			"Additional options: task-queue-count (required).",
+		Executor: loadgen.KitchenSinkExecutor{
+			WorkflowParams: kitchensink.NewWorkflowParams(kitchensink.NopActionExecuteActivity),
+			PrepareWorkflowParams: func(ctx context.Context, opts loadgen.RunOptions, params *kitchensink.WorkflowParams) error {
+				// Require task queue count
+				if opts.ScenarioOptionInt("task-queue-count", 0) == 0 {
```
bergundy, line 20: This works, though I wish it was more tightly coupled to the run options and less to reduce the chance of human error.

bergundy (approved):

scenarios/workflow_on_many_task_queues.go
cretz: But this comes with all the defaults set if you want to just alter them. But the good part about this approach and similar is you can do either (and we can add other callbacks in the future if we want them).

scenarios/workflow_on_many_task_queues.go
cretz: Yeah, I could use https://pkg.go.dev/github.com/spf13/pflag on a per scenario basis, but it's hard to use with the command because you have to know the scenario first. You can tell it "all flags are valid" and just parse unparsed ones after you've resolved the scenario, but that's a bit ugly. Yes, I could make a full argument thing on my own with data types and such, but it seemed like overkill though I wouldn't be against it.

PR #24, opened 2023-05-17
Add Docker image build & publish GHA

mindaugasrukas:
## What was changed
Added Docker image build & publish GHA

## Why?
Have pre-built images for testing.

## Checklist

1. Closes 
No
2. How was this tested:
GHA

3. Any docs updates needed?
No

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,58 @@
+name: Build & publish Docker image
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  publish_to_docker_hub:
+    runs-on: ubuntu-latest
+
+    env:
+      DOCKER_TAG: ${{ github.ref_name }}
+
+    steps:
+      - name: Dump GitHub context
+        id: github_context_step
+        run: echo '${{ toJSON(github) }}'
+
+      - name: Dump Input context
+        id: input_context_step
+        run: echo '${{ toJSON(inputs) }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+        run: echo '${{ toJSON(github) }}'
+
+      - name: Dump Input context
+        id: input_context_step
+        run: echo '${{ toJSON(inputs) }}'
+
+      - name: Checkout
+        uses: actions/checkout@v3
+      
+      - name: Set up QEMU
```
bergundy, line 28: Why is this needed?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,58 @@
...
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to DockerHub
```
bergundy, line 34: There’s a command that builds the image, this isn’t the way it’s supposed to be built

bergundy (commented):
Let’s sync on what we want here, we had a different idea of how this image would be built

.github/workflows/docker.yaml
mindaugasrukas: For having ARM images to run on macOS M1.

Sushisource:
Closing in favor of https://github.com/temporalio/omes/pull/25

PR #26, opened 2023-08-25
Docker image publishing working

Sushisource:
Works as seen in https://github.com/temporalio/omes/actions/runs/5978821365/job/16221730147


cmd/push_images.go
```diff
@@ -0,0 +1,76 @@
+// TODO: Should be de-duped with similar code in features repo
```
cretz, line 1: FWIW, this is why this repo depends on the features repo - I wanted to dedupe the whole build worker on any-version of any-SDK.

cretz (approved):
LGTM. Do we want to update the README with information here?

cmd/push_images.go
Sushisource: Yeah for sure, I think we'll want to eventually figure out what we want the separation to be and where everything goes per our standup discussion at some point, probably after this initial CICD effort is running properly.

PR #27, opened 2023-08-28
Naive steady state transition scenario

cretz:
## What was changed

Created new steady state transition scenario. This requires duration and requires a `state-transitions-per-second` scenario option. This is a naive implementation that just runs a basic workflow once to get the count, and then runs that as many times as necessary to run that count. There is no backpressure nor is there any logic in checking whether server state transitions is actually matching this.

To try out with embedded server/worker, clone branch and run:

    go run ./cmd run-scenario-with-worker --scenario state_transitions_steady --language go --embedded-server --duration 30s --option state-transitions-per-second=200

This will output logs like:

```
2023/08/28 10:47:58 INFO  Starting DevServer ExePath C:\Users\user\AppData\Local\Temp\temporal-cli-go-sdk-1.22.1.exe Args [server start-dev --ip 127.0.0.1 --port 62795 --namespace default --dynamic-config-value frontend.enableServerVersionCheck=false --log-level error --headless]
2023/08/28 10:47:59 INFO  DevServer ready
2023-08-28T10:47:59.242-0500    INFO    cmd/run_worker.go:112   Started embedded local server at: 127.0.0.1:62795
2023-08-28T10:47:59.243-0500    INFO    cmd/prepare_worker.go:65        Building go program at c:\work\tem\omes-state-transition\omes\workers\go\omes-temp-2640346882
2023-08-28T10:48:05.810-0500    INFO    cmd/run_worker.go:173   Starting worker with command: [./program.exe --task-queue state_transitions_steady:yjpvisxa --server-address 127.0.0.1:62795 --namespace default --prom-handler-path /metrics --log-level info --log-encoding console]
2023-08-28T10:48:07.021-0500    INFO    cmdoptions/client.go:68 Client connected to 127.0.0.1:62795, namespace: default
2023-08-28T10:48:07.021-0500    INFO    scenarios/state_transitions_steady.go:56        State transitions per second is 200, which means 5ms between every state transition
2023-08-28T10:48:07.072-0500    INFO    cmdoptions/client.go:68 Client connected to 127.0.0.1:62795, namespace: default
2023-08-28T10:48:07.127-0500    INFO    internal/internal_worker.go:965 Started Worker  {"Namespace": "default", "TaskQueue": "state_transitions_steady:yjpvisxa", "WorkerID": "58352@cretz-laptop@"}
2023-08-28T10:48:07.148-0500    INFO    scenarios/state_transitions_steady.go:83        Simple workflow takes 7 state transitions, which means we need to start a workflow every 35ms. Running for 30s now...
2023-08-28T10:48:37.150-0500    INFO    scenarios/state_transitions_steady.go:134       Run complete, ran 1715 iterations
2023-08-28T10:48:37.151-0500    INFO    cmd/run_worker.go:192   Sending interrupt to worker, PID: 58352
```

I do not know whether this even works or solves the needs.

### DIFF
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -25,6 +25,12 @@ type Executor interface {
 	Run(context.Context, RunOptions) error
 }
 
+// ExecutorFunc is an [Executor] implementation for a function
+type ExecutorFunc func(context.Context, RunOptions) error
+
+// Run implements [Executor.Run].
+func (e ExecutorFunc) Run(ctx context.Context, opts RunOptions) error { return e(ctx, opts) }
+
 // HasDefaultConfiguration is an interface executors can implement to show their
 // default configuration.
 type HasDefaultConfiguration interface {
@@ -66,7 +72,8 @@ func GetScenario(name string) *Scenario {
 type RunOptions struct {
 	// Name of the scenario (inferred from the file name)
 	ScenarioName string
-	// Run ID of the current scenario run, used to generate a unique task queue and workflow ID prefix.
+	// Run ID of the current scenario run, used to generate a unique task queue
+	// and workflow ID prefix. This is a single value for the whole scenario.
 	RunID string
 	// Metrics component for registering new metrics.
 	MetricsHandler client.MetricsHandler
@@ -115,20 +122,21 @@ func (r *RunConfiguration) ApplyDefaults() {
 	}
 }
 
-// Run represents a scenario run.
+// Run represents a scenario individual run (many may be in a scenario).
 type Run struct {
-	// Used for Workflow ID and task queue name generation.
-	ID string
-	// The name of the scenario used for the run.
-	ScenarioName string
-	// Temporal client to use for executing workflows, etc.
-	Client client.Client
-	// Each call to the Execute method gets a distinct `IterationInTest`. This
-	// will never be 0.
-	IterationInTest int
-	Logger          *zap.SugaredLogger
-	// Do not mutate this
-	RunOptions *RunOptions
+	// Do not mutate this, this is shared across the entire scenario
+	*RunOptions
+	// Each run should have a unique iteration.
+	Iteration int
+	Logger    *zap.SugaredLogger
+}
+
+func (r *RunOptions) NewRun(iteration int) *Run {
+	return &Run{
+		RunOptions: r,
+		Iteration:  iteration,
+		Logger:     r.Logger.With("iteration", iteration),
+	}
 }
 
 // TaskQueueForRun returns a default task queue name for the given scenario name and run ID.
@@ -136,14 +144,16 @@ func TaskQueueForRun(scenarioName, runID string) string {
 	return fmt.Sprintf("%s:%s", scenarioName, runID)
 }
 
-func (r *Run) DefaultKitchenSinkWorkflowOptions() KitchenSinkWorkflowOptions {
-	return KitchenSinkWorkflowOptions{
-		StartOptions: client.StartWorkflowOptions{
-			TaskQueue:                                TaskQueueForRun(r.ScenarioName, r.ID),
-			ID:                                       fmt.Sprintf("w-%s-%d", r.ID, r.IterationInTest),
-			WorkflowExecutionErrorWhenAlreadyStarted: true,
-		},
+func (r *Run) DefaultStartWorkflowOptions() client.StartWorkflowOptions {
+	return client.StartWorkflowOptions{
+		TaskQueue:                                TaskQueueForRun(r.ScenarioName, r.RunID),
+		ID:                                       fmt.Sprintf("w-%s-%d", r.RunID, r.Iteration),
+		WorkflowExecutionErrorWhenAlreadyStarted: true,
 	}
+}
+
+func (r *Run) DefaultKitchenSinkWorkflowOptions() KitchenSinkWorkflowOptions {
+	return KitchenSinkWorkflowOptions{StartOptions: r.DefaultStartWorkflowOptions()}
 }
 
 type KitchenSinkWorkflowOptions struct {
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -122,7 +122,7 @@ func (r *RunConfiguration) ApplyDefaults() {
 	}
 }
 
-// Run represents a scenario individual run (many may be in a scenario).
+// Run represents an individual scenario run (many may be in a scenario).
 type Run struct {
 	// Do not mutate this, this is shared across the entire scenario
 	*RunOptions
@@ -131,6 +131,7 @@ type Run struct {
 	Logger    *zap.SugaredLogger
 }
 
+// NewRun creates a new run.
 func (r *RunOptions) NewRun(iteration int) *Run {
 	return &Run{
 		RunOptions: r,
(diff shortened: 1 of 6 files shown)


scenarios/state_transitions_steady.go
```diff
@@ -0,0 +1,123 @@
...
+				default:
+				}
+			}(iter)
+		}
+	}
+	// We don't want to wait to let workflows complete here. In dev cases the
+	// local server will just shutdown, in non-dev cases, the server admin can
+	// wait. But we don't want to introduce delays for those that may want to run
+	// this quickly/frequently.
+	s.Logger.Infof("Run complete, ran %v iterations", iter)
```
bergundy, line 121: I would wait.
Getting passed the loop doesn't even mean we started the workflows.
Each omes scenario should ensure completion of all of the work that it started.

scenarios/state_transitions_steady.go
```diff
@@ -0,0 +1,123 @@
...
+	// Start a workflow every X interval until duration reached or there are N
+	// start failures in a row
+
+	const maxConsecutiveErrors = 5
+	errCh := make(chan error, 10000)
+	ticker := time.NewTicker(workflowStartInterval)
+	defer ticker.Stop()
+	var consecutiveErrCount int
+	iter := 1
+	for begin := time.Now(); time.Since(begin) < s.Configuration.Duration; iter++ {
```
bergundy, line 87: In the future you might want to make a generic executor that uses intervals instead of concurrency for spawning but since this is the only use case, it's fine to refactor later.

loadgen/scenario.go
```diff
@@ -115,37 +122,40 @@ func (r *RunConfiguration) ApplyDefaults() {
...
-		},
+func (r *Run) DefaultStartWorkflowOptions() client.StartWorkflowOptions {
+	return client.StartWorkflowOptions{
+		TaskQueue:                                TaskQueueForRun(r.ScenarioName, r.RunID),
+		ID:                                       fmt.Sprintf("w-%s-%d", r.RunID, r.Iteration),
+		WorkflowExecutionErrorWhenAlreadyStarted: true,
 	}
 }
 
+func (r *Run) DefaultKitchenSinkWorkflowOptions() KitchenSinkWorkflowOptions {
```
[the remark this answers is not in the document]
cretz: Multiple public things here are undocumented, not sure we need to overdo it here, but ok will document these

scenarios/state_transitions_steady.go
```diff
@@ -0,0 +1,123 @@
...
+		workflowStartInterval,
+		s.Configuration.Duration,
+	)
+
+	// Start a workflow every X interval until duration reached or there are N
+	// start failures in a row
+
+	const maxConsecutiveErrors = 5
+	errCh := make(chan error, 10000)
+	ticker := time.NewTicker(workflowStartInterval)
```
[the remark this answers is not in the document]
cretz: > the solution would need to also look at a metric and apply some control loop on top of that.

I intentionally didn't want to alter anything based on the metric because that's a bigger deal. I don't think the mandate is to maintain a steady count in the face of failures.

scenarios/state_transitions_steady.go
```diff
@@ -0,0 +1,123 @@
...
+			"Running for %v now...",
+		stateTransitionsPerWorkflow,
+		workflowStartInterval,
+		s.Configuration.Duration,
+	)
+
+	// Start a workflow every X interval until duration reached or there are N
+	// start failures in a row
+
+	const maxConsecutiveErrors = 5
```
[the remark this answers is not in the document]
cretz: No, not until it's needed

scenarios/state_transitions_steady.go
cretz: This is meant to run a certain duration steadily, not have some graceful unsteady close period.

> Each omes scenario should ensure completion of all of the work that it started.

I am not sure this should be true.

scenarios/state_transitions_steady.go
cretz: > it's fine to refactor later

Right! This is an important approach to take if we want to keep this stuff simple.

loadgen/scenario.go
```diff
@@ -25,6 +25,12 @@ type Executor interface {
 	Run(context.Context, RunOptions) error
 }
 
+// ExecutorFunc is an [Executor] implementation for a function
```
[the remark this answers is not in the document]
cretz: Fixed

loadgen/scenario.go
```diff
@@ -115,37 +122,40 @@ func (r *RunConfiguration) ApplyDefaults() {
 	}
 }
 
-// Run represents a scenario run.
+// Run represents a scenario individual run (many may be in a scenario).
```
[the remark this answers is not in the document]
cretz: Fixed

loadgen/scenario.go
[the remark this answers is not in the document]
cretz: Documented these. If we want to require documentation on everything though, it needs to be a separate PR to go fix that (there are several things, like `KitchenSinkWorkflowOptions` just below this, that were ok not to be documented).

scenarios/state_transitions_steady.go
bergundy: In this case it's not even fire and forget, you may exit before starting all of the workflows if the duration is short enough and the spawn rate is fast.

> I am not sure this should be true.

I would like an operator to know by the end of a scenario run that the scenario is complete and whether it is successful or not.

cretz: > In this case it's not even fire and forget, you may exit before starting all of the workflows if the duration is short enough and the spawn rate is fast.

I think we're ok with that

> that the scenario is complete and whether it is successful or not.

The software will complete with a successful exit code

bergundy: > The software will complete with a successful exit code

Yes, but in this case the scenario isn't complete yet (at least the way that we've defined completion in other scenarios), not sure why you want to change the definition of completion here. I'm especially concerned when someone runs a small amount of iterations to test this scenario and they may not realize that the scenario exits prematurely.

Sushisource: I do think we need to do this before merging. This needs to be easy to run and orphaning workflows places operational burden on the setup that we don't want.

cretz: The completion is when we stopped starting workflows. Heck, the workflows don't even have to succeed. I don't think we should define general completion across scenarios. Heck, some scenarios may just load test creating a bunch of namespaces and not even have workflows. That's why they're scenarios, they define what completion means to them.

Having said, that I will add a wait for max of a minute or so here to confirm workflow completion to satisfy this separate definition of completion for this scenario.

cretz: Note, we need to be prepared when users ask why from beginning to end of the scenario the state transition count was not steady, despite the name, we need to say we have a stop period where the scenario is still alive but not creating workflows anymore.

cretz: Added wait

scenarios/state_transitions_steady.go
```diff
@@ -114,10 +119,29 @@ func (s *stateTransitionsSteady) run(ctx context.Context) error {
...
+	// confirm all workflows on the task queue are no longer running.
+	s.Logger.Infof("Run complete, ran %v iterations, waiting on all workflows to complete", iter)
+	// First, wait for all starts to have started (they are done in goroutine)
+	startWG.Wait()
+	// Now check every so often
+	start := time.Now()
+	const interval = 3 * time.Second
+	const maxWait = time.Minute
+	for {
+		resp, err := s.Client.CountWorkflow(ctx, &workflowservice.CountWorkflowExecutionsRequest{
```
Sushisource, line 134: Heh, so, like I mentioned in the other PR, already a use-case for re-using that function.

Sushisource (approved):

scenarios/state_transitions_steady.go
cretz: Maybe a more general "EnsureNoRunningWorkflows"

PR #28, opened 2023-08-28
Add recreation of bench-go Throughputstress scenario

Sushisource:
This recreates the throughputstress scenario in bench-go. It is not precisely 1:1, but does basically all the same stuff. You can configure how many iterations of the steps *within* the workflow to run, as well as how many outer iterations (or duration) of the scenario to execute & how many to run concurrently (which are general omes-wide things).

TODO: 

- [x] Visibility check at the end, working on that now.
- [x] Make knobs configurable via passed-in inputs on CLI


workers/go/workflows/throughput_stress.go
```diff
@@ -0,0 +1,181 @@
...
+	UpdateSleep         = "sleep"
+	UpdateActivity      = "activity"
+	UpdateLocalActivity = "localActivity"
+	ASignal             = "someSignal"
+)
+
+// ThroughputStressWorkflow is meant to mimic the throughputstress scenario from bench-go of days
+// past, but in a less-opaque way. We do not ask why it is the way it is, it is not our place to
+// question the inscrutable ways of the old code.
+func ThroughputStressWorkflow(ctx workflow.Context, params *througput_stress.WorkflowParams) (string, error) {
```
cretz, line 22: Bit surprised you couldn't get this with the kitchen sink one. It has concurrent nested action sets and such, but yeah it'd take additions for more capabilities.

workers/go/activities/activities.go
```diff
@@ -1,7 +1,90 @@
...
+			WorkflowID: wid,
+			QueryType:  queryType,
+		},
+	)
+
+	if resp != nil && resp.QueryRejected != nil {
+		activity.GetLogger(ctx).Info("failed to query workflow", "queryRejected", *resp.QueryRejected)
+	}
+
+	if err != nil {
```
cretz, line 58: Move the error check above the resp check and you won't need `resp != nil` above (usually common pattern to check `err` first)

workers/go/activities/activities.go
```diff
@@ -1,7 +1,90 @@
 package activities
```
cretz, line 1: I think we didn't organize this well when omes was first developed. I think activities and workflows that work together for a purpose should be together in a package. In other words, packages should be by purpose of thing (e.g. `throughputstress` package) not by type like activity or workflow.

workers/go/activities/activities.go
```diff
@@ -1,7 +1,90 @@
...
+func (a *Activities) SelfDescribe(ctx context.Context) error {
+	info := activity.GetInfo(ctx)
+	wid := info.WorkflowExecution.ID
+
+	_, err := a.Client.DescribeWorkflowExecution(ctx, wid, "")
+	if err != nil {
+		activity.GetLogger(ctx).Info("failed to describe workflow", "Error", err)
+		return err
+	}
+	return nil
```
cretz, line 75:
```suggestion
	return err
```

A bit inconsistent with whether we log errors or not in these activities (I like not doing it personally)

workers/go/worker/worker.go
```diff
@@ -53,12 +52,17 @@ func (a *App) Run(cmd *cobra.Command, args []string) {
 
 func runWorkers(client client.Client, taskQueues []string) error {
 	errCh := make(chan error, len(taskQueues))
+	activityS := activities.Activities{
```
cretz, line 55:
```suggestion
	activityInstance := activities.Activities{
```

 (nitpick, can ignore)

workers/go/workflows/throughput_stress.go
```diff
@@ -0,0 +1,181 @@
...
+				return workflow.ExecuteActivity(actCtx, "SelfUpdate", UpdateLocalActivity).Get(ctx, nil)
+			},
+		)
+		if err != nil {
+			return "", err
+		}
+		// Possibly continue as new
+		if workflow.GetInfo(ctx).GetCurrentHistoryLength() >= params.ContinueAsNewAfterEventNumber {
+			if i == 0 {
+				return "", fmt.Errorf("trying to coninue as new on first iteration, workflow will never end")
```
cretz, line 142:
```suggestion
				return "", fmt.Errorf("trying to continue as new on first iteration, workflow will never end")
```

loadgen/througput_stress/throughput_stress.go
```diff
@@ -0,0 +1,12 @@
+package througput_stress
```
kevinkyyro, line 1:
```suggestion
package throughput_stress
```

workers/go/workflows/utils.go
```diff
@@ -0,0 +1,29 @@
...
+
+// RunConcurrently runs the given functions concurrently, waiting for all to complete.
+// Returns early if any of the functions return an error.
+func RunConcurrently(ctx workflow.Context, funcs ...func(workflow.Context) error) error {
+	completed := 0
+	var innerErr error
+	for _, f := range funcs {
+		f := f
+		workflow.Go(ctx, func(ctx workflow.Context) {
+			innerErr = f(ctx)
```
cretz, line 16: This could swallow an error.

```suggestion
			if err := f(ctx); err != nil {
				innerErr = err
			}
```

workers/go/workflows/utils.go
```diff
@@ -0,0 +1,29 @@
...
+			completed++
+		})
+	}
+
+	awaitErr := workflow.Await(ctx, func() bool { return completed == len(funcs) || innerErr != nil })
```
cretz, line 21: FWIW, there is a https://pkg.go.dev/go.temporal.io/sdk/workflow#WaitGroup for this use case, though what you have here is fine too

workers/go/workflows/throughput_stress.go
```diff
@@ -0,0 +1,181 @@
...
+		err = workflow.ExecuteLocalActivity(
+			localActCtx, "Payload", activities.MakePayloadInput(0, 256)).Get(ctx, nil)
+		if err != nil {
+			return "", err
+		}
+
+		// Do some stuff in parallel
+		err = RunConcurrently(ctx,
+			func(ctx workflow.Context) error {
+				child := workflow.ExecuteChildWorkflow(ctx, "throughputStressChild")
```
kevinkyyro, line 93: The existing throughputstress workflows do not cancel appropriately, which is a bit annoying, but I think it would be a really nice non-breaking departure to handle that here with like `PARENT_CLOSE_POLICY_TERMINATE` or something. What do you think? 

Same might apply to activity cancellation as well if any are long running, though it doesn't look like they are at a glance.

Sushisource: Cancels should propagate to all of these by default and work as expected. I'm not sure I want to change the close policy to terminate b/c all the children should conclude rapidly as well and I don't want to cover up something getting stuck with that. I'll be adding visibility checks to that effect.

workers/go/activities/activities.go
Sushisource: Yeah, I meant to get rid of this one thanks

workers/go/workflows/throughput_stress.go
Sushisource: Honestly, it was more about making it obvious than anything else. KitchenSink is kinda like bench go and as discussed I'm not sure I really like that much abstraction.

I think it's useful for fuzz testing or something, for sure, but that's not quite the purpose of this one.

loadgen/scenario.go
```diff
@@ -151,15 +156,22 @@ type KitchenSinkWorkflowOptions struct {
...
+// completion ignoring its result.
 func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSinkWorkflowOptions) error {
-	r.Logger.Debugf("Executing workflow with options: %v", options.StartOptions)
-	execution, err := r.Client.ExecuteWorkflow(ctx, options.StartOptions, "kitchenSink", options.Params)
+	return r.ExecuteAnyWorkflow(ctx, options.StartOptions, "kitchenSink", options.Params)
+}
+
+// ExecuteAnyWorkflow wraps calls to the client executing workflows to include some logging,
+// returning an error if the execution fails.
+func (r *Run) ExecuteAnyWorkflow(ctx context.Context, options client.StartWorkflowOptions, workflow interface{}, args ...interface{}) error {
```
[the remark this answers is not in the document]
Sushisource: I just like the fact that we're always logging when workflows are getting started here. It's a small but nonzero value.

loadgen/kitchensink/kitchen_sink.go
```diff
@@ -44,7 +44,7 @@ type Action struct {
 }
 
 // NopActivityName is the name of the noop activity
-const NopActivityName = "noop"
+const NopActivityName = "Noop"
```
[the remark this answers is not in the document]
Sushisource: It's because I switched to stateful activities.

loadgen/througput_stress/throughput_stress.go
```diff
@@ -0,0 +1,12 @@
...
+
+// WorkflowParams is the single input for the throughput stress workflow.
+type WorkflowParams struct {
+	// Number of times we should loop through the steps in the workflow.
+	Iterations int
+	// What iteration to start on. If we have continued-as-new, we might be starting at a nonzero
+	// number.
+	InitialIteration int
+	// If nonzero, we will continue as new after history has grown to be at least this many events.
+	ContinueAsNewAfterEventNumber int
```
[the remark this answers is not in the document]
Sushisource: Yeah, agreed. I don't think every scenario needs to be in every language. I added the json tags in any case.

workers/go/activities/activities.go
Sushisource: Yeah, agreed. I didn't change it just to avoid changing too much, but, I might as well do it now.

workers/go/activities/activities.go
Sushisource: Technically they could both be nil, if something was wrong inside the querying, but sure.

workers/go/workflows/utils.go
Sushisource: Thx

workers/go/workflows/utils.go
Sushisource: Figured it was unnecessary since we aren't really running in parallel.

workers/go/workflows/throughput_stress.go
cretz: Works for me, but I do think throughputstress stuff needs to be together and not separated

workers/go/workflows/throughput_stress.go
kevinkyyro: Cool, as long as cancellations propagate, I'm happy!

alexshtin (approved):
Throughput scenario looks good to me. I didn't review scaffolding code changes as I don't know the project. Which I really need to fix and get myself familiar with omes.

workers/go/workflows/throughput_stress.go
Sushisource: Done

loadgen/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	expectedCount int,
+	waitAtMost time.Duration,
+) error {
+	deadline := time.Now().Add(waitAtMost)
+	for {
+		visibilityCount, err := client.CountWorkflow(ctx, request)
+		if err != nil {
+			return err
+		}
+		if visibilityCount.Count == int64(expectedCount) {
```
cretz, line 25:
```suggestion
		if visibilityCount.Count >= int64(expectedCount) {
```

Otherwise, this function may be less useful as a general helper

loadgen/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
+)
+
+func VisibilityCountIsEventually(
```
cretz, line 12: In the other PR, Roey was requesting documentation on all these public functions

cretz, line 12: Arguably this is a tiny little thing only needed by one test and doesn't deserve it's own file/function. This is overabstraction IMO and will lead to sprawl.

loadgen/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	ctx context.Context,
+	client client.Client,
+	request *workflowservice.CountWorkflowExecutionsRequest,
+	expectedCount int,
+	waitAtMost time.Duration,
+) error {
+	deadline := time.Now().Add(waitAtMost)
+	for {
```
cretz, line 20: Often more clearly implemented as a time.After + time.Ticker instead of a naive sleep so you don't have to wait the interval

loadgen/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+		visibilityCount, err := client.CountWorkflow(ctx, request)
+		if err != nil {
+			return err
+		}
+		if visibilityCount.Count == int64(expectedCount) {
+			return nil
+		}
+		if time.Now().After(deadline) {
```
cretz, line 28: Often in Go you'll see start captured and this expressed as

```go
		if time.Since(start) >= waitAtMost {
```

Some even put it in a loop, e.g.

```go
for start := time.Now(); time.Since(start) < waitAtMost; {
```

(or similar)

loadgen/kitchen_sink_executor.go
```diff
@@ -30,18 +30,30 @@ func (k KitchenSinkExecutor) Run(ctx context.Context, options RunOptions) error
...
+				}
 			}
 			return run.ExecuteKitchenSinkWorkflow(ctx, &options)
 		},
-	}.Run(ctx, options)
+	}
+	return ge.Run(ctx, options)
+}
+
+func (k KitchenSinkExecutor) PreScenario(_ context.Context, _ ScenarioInfo) error {
```
cretz, line 51: I don't think we should make users implement pre/post scenario. I think they can have a run, and if they want to do something at the beginning or end, it's their code to do so as they please

loadgen/scenario.go
```diff
@@ -78,10 +88,15 @@ type RunOptions struct {
 	Configuration RunConfiguration
 	// ScenarioOptions are options passed from the command line. Do not mutate these.
 	ScenarioOptions map[string]string
+	// The namespace that was used when connecting the client.
+	Namespace string
+	// Unique salt for this scenario run. Use to provide values that are unique but stable for
+	// the duration of the scenario run.
+	Salt uuid.UUID
```
cretz, line 95:
```suggestion
	ScenarioID string
```

I think this would be clearer to users, or maybe `LocalRunID` or maybe `ExecutionID` or something. Salt is a confusing term for this use IMO.

But really, is this needed? Can't users just create a UUID if they need it?

scenarios/throughput_stress.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/operatorservice/v1"
+
+	"github.com/temporalio/omes/loadgen"
+	"github.com/temporalio/omes/loadgen/throughputstress"
+	"go.temporal.io/sdk/client"
+)
+
+// --option flags
```
cretz, line 21: Not sure these are "flags", but doesn't matter much

scenarios/throughput_stress.go
```diff
@@ -0,0 +1,105 @@
...
+			IterFlag, CANEventFlag),
+		Executor: &loadgen.GenericExecutor[ThroughputStressStorage]{
+			DefaultConfiguration: loadgen.RunConfiguration{
+				Iterations:    20,
+				MaxConcurrent: 5,
+			},
+			PreScenarioImpl: func(ctx context.Context, info loadgen.ScenarioInfo, storage *ThroughputStressStorage) error {
+				storage = &ThroughputStressStorage{
+					workflowCount: atomic.Uint64{},
+				}
```
cretz, line 46: This doesn't do anything meaningful, it only reassigns a local variable

scenarios/throughput_stress.go
```diff
@@ -0,0 +1,105 @@
...
+						ContinueAsNewAfterEventCount: continueAsNewCount,
+					})
+				// The 1 is for the final workflow run
+				storage.workflowCount.Add(uint64(result.TimesContinued + result.ChildrenSpawned + 1))
+				return err
+			},
+			PostScenarioImpl: func(ctx context.Context, info loadgen.ScenarioInfo, storage *ThroughputStressStorage) error {
+				totalWorkflowCount := storage.workflowCount.Load()
+				info.Logger.Info("Total workflows executed: ", totalWorkflowCount)
+				return loadgen.VisibilityCountIsEventually(
```
cretz, line 91: If `ExecuteAnyWorkflow` waits for workflow completed anyways, what is the value of this assertion? You're just confirming that you ran as many workflows as you said you did? I could understand if `ExecuteAnyWorkflow` just _started_ workflows and this waited for them all to complete, but otherwise I am not seeing what could fail here unless you're intentionally just integration testing our count feature.

workers/go/workflow_utils/workflow_utils.go
cretz, line 1: I think you can just put this in your workflow file for that workflows needs and we can extract later if it needs to be shared. I don't think we need a whole `workflow_utils` package for this simple, single-use function.

Sushisource: The point is just discoverability. It seems highly likely more things will be shared at some point, so the file is likely to exist, and then someone can easily discover this helper later on.

scenarios/throughput_stress.go
Sushisource: It is making sure that visibility counts match what is expected. So yeah, it is testing that. It matters and has found issues before.

loadgen/helpers.go
Sushisource: Like the other thing, there's almost certainly going to be things like this in the future, I can't imagine we're never going to share any common stuff ever

loadgen/helpers.go
Sushisource: The point here is for them to be exactly equal

loadgen/generic_executor.go
```diff
@@ -9,35 +9,56 @@ import (
 	"go.uber.org/zap"
 )
 
-type GenericExecutor struct {
+type GenericExecutor[STORAGE any] struct {
+	// Used for Executor.PreScenario
+	PreScenarioImpl func(context.Context, ScenarioInfo, *STORAGE) error
```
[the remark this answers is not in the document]
Sushisource: It's to avoid name clashes with the interface methods

loadgen/generic_executor.go
```diff
@@ -9,35 +9,56 @@ import (
...
 	// Function to execute a single iteration of this scenario
-	Execute func(ctx context.Context, run *Run) error
+	Execute func(context.Context, *Run, *STORAGE) error
+	// Used for Executor.PostScenario
+	PostScenarioImpl func(context.Context, ScenarioInfo, *STORAGE) error
 	// Default configuration if any.
 	DefaultConfiguration RunConfiguration
+	// Storage for the scenario
+	ScenarioStorage STORAGE
```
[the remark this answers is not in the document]
Sushisource: Doesn't really seem to impose much burden. Places not using it just say `any` and can ignore everything else. For things that do need it, you get to avoid having to do ugly casts. Pre/post is a pretty common thing to want to do too.

If you'd rather have it be it's own new executor though, then fine I can just do that.

loadgen/scenario.go
```diff
@@ -21,8 +23,14 @@ type Scenario struct {
 
 // Executor for a scenario.
 type Executor interface {
+	// PreScenario is called once before the scenario begins execution. Use it for one-time setup,
+	// like adding a custom search attribute.
+	PreScenario(context.Context, ScenarioInfo) error
```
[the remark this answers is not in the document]
Sushisource: They could be put inside of only `GenericExecutor`. Problem being that it's definition of `Run` invokes the user defined run once per iteration, and these need to be called only once each. Doesn't seem like a tremendous complication to me and is generally something that comes up quite a bit, but I can get rid of it from `Executor`.

workers/go/workflow_utils/workflow_utils.go
cretz: I am not convinced it justifies premature abstraction/utilities and a new package, but it's subjective so if y'all think so, ok.

scenarios/throughput_stress.go
Sushisource: It's really not a benchmarking tool is the thing. It's a loadgen tool, and it wants to verify some correctness properties about the results of that load. You can definitely argue that maybe that's not the point of a "pure" loadgen tool and I wouldn't really disagree, but, I think the practical reality of what we need is something that will supersede bench-go and ensuring stuff was correct as some basic level is indeed an important part of that.

loadgen/scenario.go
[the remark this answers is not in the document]
cretz: To me:

```go
Executor: ExecutorFunc(func(...) {
  doPreThing()
  defer doPostThing()
  &GenericExecutor{...}.Run()
})
```

makes sense and doesn't infect people that don't need pre/post. At the least I think we should get rid of the generic. I will say I may be less likely to use the generic executor if it looks less generic, but now I'm starting to wonder if we need a generic executor instead of just utilities to run for iterations-or-duration.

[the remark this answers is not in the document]
Sushisource: bro, generics are _more generic_ :wink: 

I would be fine with utilities instead of one generic executor. I'm not going to bother at the moment, because I want to finish this. I'll use a new one for this purpose. I want the storage to live on an actual executor instance so it's not just floating off in global variable space and that seems useful.

At the moment, this is all a question of style. I'd say the generic parameter doesn't really have much of any material impact, it's just a question of taste. Fine, no big deal.

Ideally, these different features you might want in executors could be easily composed. I don't know exactly what that looks like. We can figure it out later.

loadgen/generic_executor.go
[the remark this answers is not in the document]
cretz: Yeah, I think I probably will be inclined to use my own executor. We should consider, for those wanting their own executor, helpers for executing iterations/durations.

loadgen/generic_executor.go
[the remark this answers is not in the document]
cretz: At that point, generic executor is not giving me much just turning my function field into an interface function. I think this shows you can keep the pre/post stuff here but remove from the executor interface.

loadgen/helpers.go
cretz: We will share when we share, but this is a bit too premature IMO

scenarios/throughput_stress.go
cretz: It can be hard to know what to assert against I guess. I guess here this scenario is an e2e test for the visibility count API, but many scenarios may only assert their more primary goals. Technically harmless to assert anything of course. You could test that history is right, that metrics are accurate, etc. I don't mind it here, just seems like an unnecessary step that should never fail if the server tests work properly in the server repo.

loadgen/generic_executor.go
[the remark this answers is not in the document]
Sushisource: undone

workers/go/workflow_utils/workflow_utils.go
```diff
@@ -0,0 +1,32 @@
+package workflow_utils
```
cretz, line 1:
```suggestion
package workflowutils
```

(and change dir, if we have to keep this package)

workers/go/activities/activities.go
cretz, line 1: Would totally support just moving this under kitchen sink or into `worker`

cretz (approved):
I am approving, but I disagree with:

* Requiring a type parameter on the generic executor
* Requiring that the general user data be called "storage"
* Having small packages

loadgen/generic_executor.go
```diff
@@ -11,32 +11,33 @@ import (
...
 
-func (g GenericExecutor) GetDefaultConfiguration() RunConfiguration { return g.DefaultConfiguration }
+func (g *GenericExecutor) GetDefaultConfiguration() RunConfiguration {
+	return g.DefaultConfiguration
+}
 
 type genericRun struct {
 	executor *GenericExecutor
-	options  RunOptions
+	options  ScenarioInfo
```
bergundy, line 25: If you're renaming the type, please rename the field on this struct too.

loadgen/generic_executor.go
```diff
@@ -63,7 +64,8 @@ func (g *GenericExecutor) newRun(options RunOptions) (*genericRun, error) {
 
 // Run a scenario.
 // Spins up coroutines according to the scenario configuration.
-// Each coroutine runs the scenario Execute method in a loop until the scenario duration or max iterations is reached.
+// Each coroutine runs the scenario Execute method in a loop until the scenario duration or max
+// iterations is reached.
```
bergundy, line 68: I think line length of 120 is standard these days, can you make sure your editor supports that?

loadgen/scenario.go
```diff
@@ -164,15 +183,22 @@ type KitchenSinkWorkflowOptions struct {
...
 func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSinkWorkflowOptions) error {
-	r.Logger.Debugf("Executing workflow with options: %v", options.StartOptions)
-	execution, err := r.Client.ExecuteWorkflow(ctx, options.StartOptions, "kitchenSink", options.Params)
+	return r.ExecuteAnyWorkflow(ctx, options.StartOptions, "kitchenSink", nil, options.Params)
+}
+
+// ExecuteAnyWorkflow wraps calls to the client executing workflows to include some logging,
+// returning an error if the execution fails.
+func (r *Run) ExecuteAnyWorkflow(ctx context.Context, options client.StartWorkflowOptions, workflow interface{}, valuePtr interface{}, args ...interface{}) error {
+	r.Logger.Debugf("Executing workflow %s with options: %v", workflow, options)
```
bergundy, line 195: nit: This is noisy since it logs for every workflow spawned, I'd rather keep debug level a bit cleaner than this.

scenarios/throughput_stress.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	genericExec := &loadgen.GenericExecutor{
+		DefaultConfiguration: loadgen.RunConfiguration{
+			Iterations:    20,
+			MaxConcurrent: 5,
+		},
+		Execute: func(ctx context.Context, run *loadgen.Run) error {
+			internalIterations := run.ScenarioInfo.ScenarioOptionInt(IterFlag, 5)
+			continueAsNewCount := run.ScenarioInfo.ScenarioOptionInt(CANEventFlag, 100)
```
bergundy, line 54: I'm wondering why we need continue-as-new and child workflows to test throughput.
It's not _wrong_ but omes exposes metrics for workflow e2e latency and throughput that we won't be able to leverage with this approach.

loadgen/generic_executor.go
Sushisource: It's not a matter of editor support, it's a matter of nicely fitting multiple buffers side by side. It's nice, I like it, and it's harmless. If it's something anyone cares about, stuff it in editorconfig or w/e. Shouldn't really be a thing anyone ever needs to comment on in a review.

loadgen/generic_executor.go
Sushisource: Will do

loadgen/scenario.go
Sushisource: The reason is to avoid counting things when multiple runs have been made on the same cluster. This is very easy to have happen right now since run-id is provided on the CLI and may very easily take the same value repeatedly.

loadgen/scenario.go
Sushisource: That was already the case?

scenarios/throughput_stress.go
Sushisource: I don't disagree. I'm just recreating the scenario. Really, the name `throughputstress` is probably just not precise.

loadgen/scenario.go
```diff
@@ -99,6 +108,12 @@ func (r *RunOptions) ScenarioOptionInt(name string, defaultValue int) int {
 	return i
 }
 
+// UniqueRunID returns a unique RunID for this scenario run. Since RunID is user-specified and may
+// be non-unique, you can use this for a stable (for the run) and guaranteed unique ID.
+func (s *ScenarioInfo) UniqueRunID() string {
+	return fmt.Sprintf("%s-%s", s.RunID, s.Salt.String())
+}
```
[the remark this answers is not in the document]
Sushisource: OK, I fixed this by just asserting no workflows already exist with that Run ID when starting

scenarios/throughput_stress.go
kevinkyyro: Without looking at the specific assertion here, just want to say that server teams would definitely get value out of the loadgen failing on invariant violations. I wouldn't ask the loadgen tool to exhaustively check invariants, but I think "if the server tests work properly" is not a safe assumption in reality...

PR #29, opened 2023-09-07
Re-do throughputstress executor as a workflow

Sushisource:
Still a couple of TODOs left but no big changes

## What was changed
Re-did the throughputstress scenario to be driven from a workflow

## Why?
Makes it much easier to invoke by simply starting a workflow in CI/CD pipelines

## Checklist

1. Closes 

2. How was this tested:
Ran directly

3. Any docs updates needed?

### DIFF
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -75,7 +75,7 @@ type ScenarioInfo struct {
 	ScenarioName string
 	// Run ID of the current scenario run, used to generate a unique task queue
 	// and workflow ID prefix. This is a single value for the whole scenario, and
-	// not a Workflow RunId.
+	// not a Workflow RunID.
 	RunID string
 	// Metrics component for registering new metrics.
 	MetricsHandler client.MetricsHandler
@@ -101,6 +101,22 @@ func (s *ScenarioInfo) ScenarioOptionInt(name string, defaultValue int) int {
 		panic(err)
 	}
 	return i
+}
+
+func (s *ScenarioInfo) ScenarioOptionDuration(name string, defaultValue time.Duration) time.Duration {
+	v := s.ScenarioOptions[name]
+	if v == "" {
+		return defaultValue
+	}
+	d, err := time.ParseDuration(v)
+	if err != nil {
+		panic(err)
+	}
+	return d
+}
+
+func (s *ScenarioInfo) TaskQueue() string {
+	return TaskQueueForRun(s.ScenarioName, s.RunID)
 }
 
 const DefaultIterations = 10
@@ -147,10 +163,6 @@ func (s *ScenarioInfo) NewRun(iteration int) *Run {
 // TaskQueueForRun returns a default task queue name for the given scenario name and run ID.
 func TaskQueueForRun(scenarioName, runID string) string {
 	return fmt.Sprintf("%s:%s", scenarioName, runID)
-}
-
-func (r *Run) TaskQueue() string {
-	return TaskQueueForRun(r.ScenarioName, r.RunID)
 }
 
 // DefaultStartWorkflowOptions gets default start workflow info.
--- a/loadgen/workflow_executor.go
+++ b/loadgen/workflow_executor.go
@@ -1,0 +1,41 @@
+package loadgen
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.temporal.io/sdk/client"
+)
+
+// WorkflowExecutor is a very dumb executor that does nothing but run the specified workflow.
+// Scenarios using it are expected to implement all their driver logic as a workflow.
+type WorkflowExecutor struct {
+	WorkflowType       string
+	WorkflowArgCreator func(info ScenarioInfo) []interface{}
+}
+
+func (w WorkflowExecutor) Run(ctx context.Context, info ScenarioInfo) error {
+	// TODO: ??
+	//executorTimeout := info.ScenarioOptionDuration(scenarios.ExecutorTimeoutFlag, 1*time.Hour)
+	executorTimeout := 1 * time.Hour
+	workflowOpts := client.StartWorkflowOptions{
+		ID:                       fmt.Sprintf("%s-driver", info.TaskQueue()),
+		TaskQueue:                info.TaskQueue(), // TODO: Should be special tq for driver?
+		WorkflowExecutionTimeout: executorTimeout,
+	}
+	run, err := info.Client.ExecuteWorkflow(
+		ctx,
+		workflowOpts,
+		w.WorkflowType,
+		w.WorkflowArgCreator(info)...,
+	)
+	if err != nil {
+		return err
+	}
+	err = run.Get(ctx, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
--- a/workers/go/throughputstress/activities.go
+++ b/workers/go/throughputstress/activities.go
@@ -2,9 +2,17 @@ package throughputstress
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 
+	"go.temporal.io/api/workflowservice/v1"
+
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/operatorservice/v1"
+	"go.temporal.io/api/serviceerror"
+
+	"github.com/temporalio/omes/loadgen/throughputstress"
 	"go.temporal.io/sdk/activity"
 	"go.temporal.io/sdk/client"
 )
@@ -83,4 +91,47 @@ func (a *Activities) SelfUpdate(ctx context.Context, updateName string) error {
 func (a *Activities) SelfSignal(ctx context.Context, signalName string) error {
 	we := activity.GetInfo(ctx).WorkflowExecution
 	return a.Client.SignalWorkflow(ctx, we.ID, we.RunID, signalName, nil)
+}
+
+func (a *Activities) EnsureSearchAttributeRegistered(ctx context.Context, namespace string) error {
+	attribMap := map[string]enums.IndexedValueType{
+		throughputstress.ThroughputStressScenarioIdSearchAttribute: enums.INDEXED_VALUE_TYPE_KEYWORD,
+	}
+	_, err := a.Client.OperatorService().AddSearchAttributes(ctx,
+		&operatorservice.AddSearchAttributesRequest{
+			Namespace:        namespace,
+			SearchAttributes: attribMap,
+		})
+	var svcErr *serviceerror.AlreadyExists
+	if !errors.As(err, &svcErr) {
+		return err
+	}
+
+	return nil
+}
+
+func (a *Activities) VisibilityCount(ctx context.Context, query string) (int64, error) {
+	visibilityCount, err := a.Client.CountWorkflow(ctx, &workflowservice.CountWorkflowExecutionsRequest{
+		Query: query,
+	})
+	if err != nil {
+		return -1, err
+	}
+	return visibilityCount.Count, nil
+}
+
+type VisibilityCountMatchesInput struct {
+	Query    string
+	Expected int
+}
+
+func (a *Activities) VisibilityCountMatches(ctx context.Context, input *VisibilityCountMatchesInput) error {
+	count, err := a.VisibilityCount(ctx, input.Query)
+	if err != nil {
+		return err
+	}
+	if count != int64(input.Expected) {
+		return fmt.Errorf("expected %d workflows in visibility, got %d", input.Expected, count)
+	}
+	return nil
 }
--- a/workers/go/throughputstress/workflow.go
+++ b/workers/go/throughputstress/workflow.go
@@ -2,11 +2,10 @@ package throughputstress
 
 import (
 	"fmt"
-	"github.com/temporalio/omes/workers/go/workflowutils"
 	"time"
 
 	"github.com/temporalio/omes/loadgen/throughputstress"
-	"github.com/temporalio/omes/scenarios"
+	"github.com/temporalio/omes/workers/go/workflowutils"
 	"go.temporal.io/sdk/temporal"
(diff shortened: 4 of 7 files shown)


loadgen/workflow_executor.go
```diff
@@ -0,0 +1,41 @@
...
+// WorkflowExecutor is a very dumb executor that does nothing but run the specified workflow.
+// Scenarios using it are expected to implement all their driver logic as a workflow.
+type WorkflowExecutor struct {
+	WorkflowType       string
+	WorkflowArgCreator func(info ScenarioInfo) []interface{}
+}
+
+func (w WorkflowExecutor) Run(ctx context.Context, info ScenarioInfo) error {
+	// TODO: ??
+	//executorTimeout := info.ScenarioOptionDuration(scenarios.ExecutorTimeoutFlag, 1*time.Hour)
```
Sushisource, line 20: Need to decide how to fix this. It's not really a scenario option, but an executor-specific option.

loadgen/scenario.go
```diff
@@ -103,6 +103,22 @@ func (s *ScenarioInfo) ScenarioOptionInt(name string, defaultValue int) int {
 	return i
 }
 
+func (s *ScenarioInfo) ScenarioOptionDuration(name string, defaultValue time.Duration) time.Duration {
```
cretz, line 106: I would like to see a project shortly that puts scenario options as cobra flags

loadgen/workflow_executor.go
cretz: I think we have duration on the main args right, can't you reuse that?

loadgen/workflow_executor.go
```diff
@@ -0,0 +1,41 @@
...
+	executorTimeout := 1 * time.Hour
+	workflowOpts := client.StartWorkflowOptions{
+		ID:                       fmt.Sprintf("%s-driver", info.TaskQueue()),
+		TaskQueue:                info.TaskQueue(), // TODO: Should be special tq for driver?
```
cretz, line 24: Cleanup is done by task queue, so I like the idea that everything shares a task queue where we can, though there will surely be cross-task-queue scenarios in the future

loadgen/workflow_executor.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"go.temporal.io/sdk/client"
+)
+
+// WorkflowExecutor is a very dumb executor that does nothing but run the specified workflow.
+// Scenarios using it are expected to implement all their driver logic as a workflow.
+type WorkflowExecutor struct {
```
cretz, line 13: I thought you were going to make it even more generic where a workflow would execute omes with its args as an activity if `--in-workflow` was given to the scenario or something. With this way, arguably this executor has no big value since anyone can just make an executor that makes a simple client call to start a single workflow.

workers/go/throughputstress/activities.go
```diff
@@ -2,9 +2,17 @@ package throughputstress
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 
+	"go.temporal.io/api/workflowservice/v1"
+
```
cretz, line 10: Not a big fan of all the blank lines in the imports. I know server and SDK do it here and there for legacy reasons, but I would encourage Go authors to just have two blocks: stdlib and non-stdlib.

cretz (approved):

loadgen/workflow_executor.go
Sushisource: The problem with that is you don't end up getting any durability guarantees which is important. I figure as we add more CI type scenarios like this one they can abstract out the shared bits of useful workflow logic

workers/go/throughputstress/activities.go
Sushisource: I dunno why GoLand refuses to do things the goimports way. I keep telling it to.

cretz: I think the goimports way is not to mess with existing import blocks. So it'll sort, and IDEs will auto-add to the bottom, but if you introduce a blank line, it'll retain it.

loadgen/workflow_executor.go
cretz: My fear is scenarios are now gonna have to choose whether they are wrapped as a workflow or not, which really feels more like a operator concern than a scenario author one.

workers/go/throughputstress/activities.go
Sushisource: I've never touched them manually ever, so it's something the IDE is doing :shrug:

loadgen/workflow_executor.go
cretz: > If we agree with that, though, we can open an item in here?

I don't agree here. I think a _required_ extra layer of indirection from the caller to the scenario is a bit much. It's not whether we know Temporal will always be present, it's that a scenario is "any code" as opposed to "any workflow code". Scenario authors need a Temporal client, which is not available in a workflow. So it'll get confusing because some scenarios want workflow code and others want normal Go code.

> we should have a layer on top of every scenario that allows it to be run as a workflow.

The problem here is, like Spencer said, you can't do the things in workflow to ensure it's durable. Just a wrapper workflow that is a single max-attempt-1 activity that just runs the thing while heartbeating doesn't give you many workflow benefits.

Maybe we can have a middleground here. Maybe we leave default as it is today for most scenarios. And there is a "wrap everything in workflow" CLI option whose workflow by default is the single-max-attempt-1-activity. For scenarios that must be run as a workflow because that have in-overarching-workflow code to write in addition to normal Temporal client code, they can set a different overarching workflow type and _require_ that CLI option be given for its scenario.

Sushisource: > I feel like this PR is taking the project in the wrong direction. This isn't what the original vision was.

This is I think maybe a bit not helpful, because we identified a need that we can't really meet with the existing setup, so we have to evolve to address that.

And, then yeah, we've covered why just slapping a workflow layer on top isn't really sufficient.

What I'm curious to hear is why that original vision matters, or why
> a required extra layer of indirection from the caller

Is even necessarily a bad thing. Not that I definitely disagree, it's just worth questioning. Conversely, there's clearly quite a bit of value gained by having the drivers be workflows in the form of durability. It could even work by just using embedded server specifically for the driver workflow, when a scenario isn't targeting a temporal instance (which, I still question if that's something worth caring about, we're testing temporal after all). So, what really is the "vision", and why is it important?

> there is a "wrap everything in workflow" CLI option whose workflow by default is the single-max-attempt-1-activity that just runs omes

This I think doesn't really provide much value - because the main point is that consumers who want to operate in workflow mode want to do so by using the `temporal` CLI or a client or whatever. I only made the workflow executor to just have a really easy way to run the scenario locally, and allow it to be executable using the normal `omes` CLI flags.

> For scenarios that must be run as a workflow .. they can set a different overarching workflow type and require that CLI option be given for its scenario.

This is kinda what I did, no? Except there's no flag, it just happens, that's the way this scenario runs.

loadgen/workflow_executor.go
Sushisource: I could, it's a little unspecific though. Since, it's a timeout now, really not a duration.

loadgen/workflow_executor.go
Sushisource: Kept it the same

loadgen/workflow_executor.go
bergundy: I'm not sure what needs to be durable and can't be without a workflow.

Looks like there are a couple of steps here:
1. Launch a bunch of workflows with a concurrency limit.
2. Query visibility to verify proper scenario completion.

For (1), using child workflows lets you resume from crashes without losing any progress but adds significant overhead to every start workflow call. If we were to take a different approach allowing omes to be runnable as an activity, we could use heartbeats for checkpoints without the child workflow overhead.

For (2), We could add a `validate` step that can be launched by an external coordinator for any scenario, or use the same heartbeat details to resume at that step.

I would consider not running directly as an activity, and keeping as a separate command and using the client to heartbeat and checkpoint if given an activity task token at startup.

The reason this approach takes the project away from the original vision is that one of the goals for omes was to automatically instrument workflow E2E latency and throughput, putting everything in a single workflow makes this impossible.

If you still want the child workflow approach, I would integrate it more tightly into the framework, e.g. create a `GenericChildWorkflowExecutor` and capture the required metrics.

Also your approach requires every workflow based scenario to be registered with the worker, which beats the point of having scenarios that are self contained.

workers/go/throughputstress/workflow.go
```diff
@@ -20,6 +19,110 @@ const (
...
+				completedWorkflowCount += 1 + result.ChildrenSpawned + result.TimesContinued
+				if completedTopLevelWorkflowCount == params.Iterations {
+					allWorkflowsDoneSet.SetValue(nil)
+				}
+			}
+		})
+	}
+
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		for i := 0; i < params.Iterations; i++ {
```
bergundy, line 93: If iterations are greater than some threshold you'll want to split this into multiple child workflows each executing some of the workflows. For a very large number of iterations, you might want a multi-level tree structure to fanout.

loadgen/workflow_executor.go
cretz: > This I think doesn't really provide much value

Only in that unfortunately some external systems (CICD) require a workflow to run anything. Other than that, I agree there's no value.

> This is kinda what I did, no? Except there's no flag, it just happens, that's the way this scenario runs.

Yup, this works for me.

My whole concern is that it's a bit confusing for scenario authors to have to concern themselves with whether where their scenario runs requires an overarching workflow. But given the lack of "any scenario can be a workflow if the caller _must_ have one", just a workflow for this one makes sense.

Sushisource: @bergundy Your suggestions can work, but I have two concerns:

1. It doesn't solve the original issue, which is that the pipeline(s) want an easy way to trigger a scenario by simply starting a workflow, rather than needing a copy of the omes binary/image.
2. This is all specific to this scenario. We can do durability the hard(er) way, or we can use our own product, and the latter seems preferable.

If the only things stopping us from doing that are 
> child workflow overhead.

Well, just use a local activity to start one.
And
> automatically instrument workflow E2E latency and throughput

It seems we can also solve that by simply pulling that info from visibility and/or metrics. I'm not sure I follow why that doesn't work with a workflow driver.

Lastly, and probably most importantly, right now we have a tool which _isn't used_. These changes are motivated by _wanting to use it_. I want to be a little less concerned with philosophy and a little more concerned with actually running something and iterating on it.

bergundy: > It doesn't solve the original issue, which is that the pipeline(s) want an easy way to trigger a scenario by simply starting a workflow, rather than needing a copy of the omes binary/image.

Why can't it launch omes as a container from an activity? We're trying to come up with a standard way to launch any scenario, our pipelines should be able to run any omes scenario, not just this one.

> This is all specific to this scenario. We can do durability the hard(er) way, or we can use our own product, and the latter seems preferable. There are likely to be other concerns that your specific fixes may not address.

Is this really a requirement or are we just trying to stick to what we had? Can we come up with a solution that works for any omes scenario and worker combination?

> It seems we can also solve that by simply pulling that info from visibility and/or metrics. I'm not sure I follow why that doesn't work with a workflow driver. And also isn't actually something we really need at the moment, since we need a loadgen tool, and don't actually have a big need for a genuine benchmarking tool at the moment - not that we never will of course.

I don't see a point in a loadgen tool that doesn't do any measurements TBH, for me this is one of the top goals of this project. Sure you can find different ways to measure, but I think there's value in having the default mode measured automatically and not have to invent custom ways to measure each scenario.

> Lastly, and probably most importantly, right now we have a tool which isn't used. These changes are motivated by wanting to use it. I want to be a little less concerned with philosophy and a little more concerned with actually running something and iterating on it.

I fear we'll end up building the same tool that we're trying to replace if we're not opinionated in the way that we structure scenarios.

I think if you just want to run a single workflow, why even put it in omes?

I'm not opposed to having a workflow driven executor but I would implement it as an executor that can run arbitrary workflows just like the generic executor and it should come with OOTB support for metrics.

workers/go/throughputstress/workflow.go
Sushisource: I'll add a comment about this technique so it can be added if/when this gets used that way. Not immediately clear it will be.

loadgen/workflow_executor.go
Sushisource: Yes, it was a real requirement, from the folks who are actually going to be running this in pipelines, and they want it to be trigger-able as a workflow - and that makes good sense.

If the only thing we're "losing" is this as-yet-unused execution histogram, I'm just not seeing how that's a big issue. We are going to be tracking status using our existing metrics that Temporal emits

>not have to invent custom ways to measure each scenario.

That's exactly it, this additional histogram is a new, custom way, whereas we already have the expertise and dashboards built up around our existing metrics. We get exactly what we need with `workflow_endtoend_latency` right now.

> I would implement it as an executor that can run arbitrary workflows just like the generic executor and it should come with OOTB support for metrics.

It is? That's `workflow_executor`, there's just not much to it. I've even added the histogram metric to it, it's just maybe not particularly interesting in this case since it'll be very long times.

Sushisource:
Closing this because Roey and I agreed that the nature of Omes as a load generation tool is that it intends to exert fine-grained control over load generated against the server, and hence generating all that load from *within one workflow* is somewhat synthetic, and also limited by the scalability of one workflow (despite the fact that it does bring the benefit of easy durability). We would prefer to have the omes binary/client itself control the launching of many workflows in order to accomplish that goal. This means that usages of omes from existing infrastructure which would like to use workflows should consider using a driver workflow with activities that run omes, and those activities can use heartbeats to maintain some state and be resumable in the event of failures.

PR #30, opened 2023-09-13
Introduce --tls flag

aromanovich:
## What was changed

`--tls` flag was added.

## Why?

To allow connecting to Temporal clusters that use TLS but not require client certificates.

### DIFF
--- a/cmd/cmdoptions/client.go
+++ b/cmd/cmdoptions/client.go
@@ -34,10 +34,11 @@ func (c *ClientOptions) loadTLSConfig() (*tls.Config, error) {
 			return nil, fmt.Errorf("failed to load certs: %s", err)
 		}
 		return &tls.Config{Certificates: []tls.Certificate{cert}}, nil
-	} else if c.EnableTLS {
-		return &tls.Config{}, nil
 	} else if c.ClientKeyPath != "" {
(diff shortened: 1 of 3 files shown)


cmd/cmdoptions/client.go
```diff
@@ -33,6 +34,8 @@ func (c *ClientOptions) loadTLSConfig() (*tls.Config, error) {
 			return nil, fmt.Errorf("failed to load certs: %s", err)
 		}
 		return &tls.Config{Certificates: []tls.Certificate{cert}}, nil
+	} else if c.EnableTLS {
+		return &tls.Config{}, nil
 	} else if c.ClientKeyPath != "" {
 		return nil, errors.New("got TLS key with no cert")
 	}
```
bergundy, line 41: I'd rather still fail if only the key is provided.

```suggestion
	} else if c.ClientKeyPath != "" {
		return nil, errors.New("got TLS key with no cert")
	} else if c.EnableTLS {
		return &tls.Config{}, nil
	}
```

bergundy (commented):
Thanks for this, please also add to the Python worker or I can do it for you.

cretz (approved):
LGTM (agree with @bergundy's caveat though)

cretz:
Thanks!

PR #36, opened 2023-11-02
Initial proto definitions for fuzzy workflow

Sushisource:
These are the protos that will be used for the upgraded version of kitchen sink, which will be capable of executing arbitrarily generated workflow definitions

### DIFF
--- a/workers/proto/kitchen_sink/kitchen_sink.proto
+++ b/workers/proto/kitchen_sink/kitchen_sink.proto
@@ -28,10 +28,10 @@ message ClientSequence {
   repeated ClientActionSet action_sets = 1;
 }
 
-// A set of client actions to execute. Client actions are always executed concurrently, since
-// sequential execution can be represented by a series of 1-size action sets.
+// A set of client actions to execute.
 message ClientActionSet {
   repeated ClientAction actions = 1;
+  bool concurrent = 2;
 }
 
 message ClientAction {
@@ -39,6 +39,7 @@ message ClientAction {
     DoSignal do_signal = 1;
     DoQuery do_query = 2;
     DoUpdate do_update = 3;
+    ClientActionSet nested_actions = 4;
   }
 }
 
@@ -66,15 +67,14 @@ message DoUpdate {
   oneof variant {
     // An update must exist named `do_actions_update` which does the same thing the signal handler
     // would do when given the provided action set. The handler should return the `WorkflowState`
-    // when done.
+    // when done. You may also include a `ReturnErrorAction` or `ContinueAsNewAction` in the set
+    // to exit the handler in those ways.
     ActionSet do_actions = 1;
-    // This update request must be rejected by the workflow using a handler named `always_reject`
+    // This update request must be rejected (at validation time) by the workflow using a handler
+    // named `always_reject`
     google.protobuf.Empty reject_me = 2;
-    // This update request must be accepted and then failed by the workflow using a handler named
-    // `always_fail`
-    temporal.api.failure.v1.Failure fail_update = 3;
     // Send an arbitrary update request
-    HandlerInvocation custom = 4;
+    HandlerInvocation custom = 3;
   }
 }
 
@@ -108,9 +108,9 @@ message Action {
     TimerAction timer = 1;
     ExecuteActivityAction exec_activity = 2;
     ExecuteChildWorkflowAction exec_child_workflow = 3;
-    AwaitSignalAction await_signal = 4;
-    AwaitUpdateAction await_update = 5;
-    SendSignalAction send_signal = 6;
+    AwaitWorkflowState await_workflow_state = 4;
+    SendSignalAction send_signal = 5;
+    CancelWorkflowAction cancel_workflow = 6;
     SetPatchMarkerAction set_patch_marker = 7;
     UpsertSearchAttributesAction upsert_search_attributes = 8;
     UpsertMemoAction upsert_memo = 9;
@@ -128,11 +128,18 @@ message Action {
 // commands, but some care is taken to remove things that would be uninteresting for these tests.
 
 // All cancellable commands will have this available as a field. If it is set, the command
-// should be cancelled at the specified juncture.
+// should be cancelled at the specified juncture (if possible, not all command types will be
+// cancellable at all stages. Is is up to the generator to produce valid conditions).
 message CancelCondition {
   oneof condition {
+    // Cancel the command before it's begun - IE: Cancel it immediately after starting it with
+    // no await, within the same workflow task.
     google.protobuf.Empty before_started = 1;
+    // Cancel the command after it's been started. Not all SDKs will know when a command is started
+    // and in those cases they should issue the cancellation in the next workflow task after
+    // creating the command.
     google.protobuf.Empty after_started = 2;
+    // Cancel the command after it's already completed.
     google.protobuf.Empty after_completed = 3;
   }
 }
@@ -189,7 +196,6 @@ message ExecuteChildWorkflowAction {
   google.protobuf.Duration workflow_task_timeout = 9;
   // Default: PARENT_CLOSE_POLICY_TERMINATE.
   ParentClosePolicy parent_close_policy = 10;
-  // string control = 11; (unused from StartChildWorkflowExecutionCommandAttributes)
   // Default: WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE.
   temporal.api.enums.v1.WorkflowIdReusePolicy workflow_id_reuse_policy = 12;
   temporal.api.common.v1.RetryPolicy retry_policy = 13;
@@ -204,14 +210,14 @@ message ExecuteChildWorkflowAction {
   temporal.omes.kitchen_sink.child_workflow.ChildWorkflowCancellationType cancellation_type = 18;
   // Whether this child should run on a worker with a compatible build id or not.
   VersioningIntent versioning_intent = 19;
+  // If set, when to cancel the child
+  CancelCondition cancel_condition = 20;
 }
 
-message AwaitSignalAction {
-  string signal_name = 1;
-}
-
-message AwaitUpdateAction {
-  string update_name = 1;
+// Wait for the workflow state to have a matching k/v entry
+message AwaitWorkflowState {
+  string key = 1;
+  string value = 2;
 }
 
 message SendSignalAction {
@@ -224,6 +230,14 @@ message SendSignalAction {
   repeated temporal.api.common.v1.Payload args = 4;
   // Headers to attach to the signal
   map<string, temporal.api.common.v1.Payload> headers = 5;
+  // If set, when to cancel the sending of the signal (only before-sent is valid)
+  CancelCondition cancel_condition = 6;
+}
+
+// Cancel an external workflow (may be a child)
+message CancelWorkflowAction {
+  string workflow_id = 1;
+  string run_id = 2;
 }
 
 // patched or getVersion API
--- a/workers/proto/kitchen_sink/kitchen_sink.proto
+++ b/workers/proto/kitchen_sink/kitchen_sink.proto
@@ -127,25 +127,31 @@ message Action {
 // The following actions and the messages they depend on are largely duplicative of Core/Server's
 // commands, but some care is taken to remove things that would be uninteresting for these tests.
 
-// All cancellable commands will have this available as a field. If it is set, the command
-// should be cancelled at the specified juncture (if possible, not all command types will be
-// cancellable at all stages. Is is up to the generator to produce valid conditions).
-message CancelCondition {
+// All await commands will have this available as a field. If it is set, the command
+// should be either awaited upon, cancelled, or abandoned at the specified juncture (if possible,
+// not all command types will be cancellable at all stages. Is is up to the generator to produce
+// valid conditions).
+message AwaitableChoice {
   oneof condition {
+    // Wait for the command to finish. This is the default.
+    google.protobuf.Empty wait_finish = 1;
+    // Start the command, but don't bother waiting on it after that.
+    google.protobuf.Empty abandon_after_start = 2;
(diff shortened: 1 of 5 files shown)


workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  VersioningIntent versioning_intent = 19;
+}
+
+message AwaitSignalAction {
+  string signal_name = 1;
+}
+
+message AwaitUpdateAction {
+  string update_name = 1;
+}
```
Sushisource, line 215: I'm thinking actually I could replace both of these with `AwaitWorkflowStateCondition` and a key/value pair.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  // always retryable. Either this or schedule_to_close_timeout must be specified.
+  google.protobuf.Duration start_to_close_timeout = 8;
+  // Maximum time allowed between successful worker heartbeats.
+  google.protobuf.Duration heartbeat_timeout = 9;
+  // Activities are provided by a default retry policy controlled through the service dynamic
+  // configuration. Retries are happening up to schedule_to_close_timeout. To disable retries set
+  // retry_policy.maximum_attempts to 1.
+  temporal.api.common.v1.RetryPolicy retry_policy = 10;
+  // Whether or not this activity will be a local activity
+  oneof locality {
```
Quinn-With-Two-Ns, line 170: Given that normal activities have some options local activities do not does it make sense for them to be a separate action?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+    google.protobuf.Empty after_started = 2;
+    google.protobuf.Empty after_completed = 3;
+  }
+}
+
+message TimerAction {
+  uint64 milliseconds = 1;
+}
+
+message ExecuteActivityAction {
```
Quinn-With-Two-Ns, line 144: Should we have version fields?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  oneof locality {
+    google.protobuf.Empty is_local = 11;
+    RemoteActivityOptions remote = 12;
+  }
+  // If set, when to cancel the activity
+  CancelCondition cancel_condition = 13;
+}
+
+message ExecuteChildWorkflowAction {
+  string namespace = 2;
```
Quinn-With-Two-Ns, line 179: Worth including given this is not supported in some SDKs and cloud?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+// between each of those sets, which may not be desired.
+//
+// All actions are handled before proceeding to the next action set, unless one of those actions
+// would cause the workflow to complete/fail/CAN.
+message ActionSet {
+  repeated Action actions = 1;
+  bool concurrent = 2;
+}
+
+message Action {
```
Quinn-With-Two-Ns, line 106: In this system how would I model activities or signal handlers that depend on each other or the main workflow thread?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  oneof variant {
+    // An update must exist named `do_actions_update` which does the same thing the signal handler
+    // would do when given the provided action set. The handler should return the `WorkflowState`
+    // when done.
+    ActionSet do_actions = 1;
+    // This update request must be rejected by the workflow using a handler named `always_reject`
+    google.protobuf.Empty reject_me = 2;
+    // This update request must be accepted and then failed by the workflow using a handler named
+    // `always_fail`
+    temporal.api.failure.v1.Failure fail_update = 3;
```
Quinn-With-Two-Ns, line 75: Why do we need this, shouldn't the updates action set result in a return or error?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+    ActionSet nested_action_set = 14;
+  }
+}
+
+// The following actions and the messages they depend on are largely duplicative of Core/Server's
+// commands, but some care is taken to remove things that would be uninteresting for these tests.
+
+// All cancellable commands will have this available as a field. If it is set, the command
+// should be cancelled at the specified juncture.
+message CancelCondition {
```
Quinn-With-Two-Ns, line 132: I see this was only done for activities, not the other commands like child workflows or timers. Is there any reason why?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  oneof condition {
+    google.protobuf.Empty before_started = 1;
+    google.protobuf.Empty after_started = 2;
```
Quinn-With-Two-Ns, line 135: How does  this work with activities since at least Go and Java do not expose an activity has started  to users.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+    AwaitUpdateAction await_update = 5;
+    SendSignalAction send_signal = 6;
+    SetPatchMarkerAction set_patch_marker = 7;
+    UpsertSearchAttributesAction upsert_search_attributes = 8;
+    UpsertMemoAction upsert_memo = 9;
+    WorkflowState set_workflow_state = 10;
+
+    ReturnResultAction return_result = 11;
+    ReturnErrorAction return_error = 12;
+    ContinueAsNewAction continue_as_new = 13;
```
Quinn-With-Two-Ns, line 121: We probably need a cancel action to for a  workflow to cancel another workflow or itself

workers/proto/kitchen_sink/kitchen_sink.proto
cretz: What does "depend on each other" mean exactly? A signal can set state, but not sure what the dependent signal is expecting. Are you asking for a wait-for-state?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+}
+
+message DoUpdate {
+  oneof variant {
+    // An update must exist named `do_actions_update` which does the same thing the signal handler
+    // would do when given the provided action set. The handler should return the `WorkflowState`
+    // when done.
+    ActionSet do_actions = 1;
+    // This update request must be rejected by the workflow using a handler named `always_reject`
+    google.protobuf.Empty reject_me = 2;
```
cretz, line 72: Why empty instead of bool? I would like to propose no backwards compatibility requirements on these protos. The protos only need to be consumable with the same code in the repo at the same commit. Backwards incompatible changes can/should be allowed in the proto code as with all code in this repository. Also has added benefit that we don't have to overthink now, can build something and come back to improve.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: We'd have to dupe everything else in the message which is why I went with that.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: It's in the remote options. Can't have versioning intent for a local activity.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Yeah, it's not going to get used initially but fine to have in here.

workers/proto/kitchen_sink/kitchen_sink.proto
Quinn-With-Two-Ns: Yeah I am arguing your not duplicating everything since not everything in a activity applies to a local activity. I think it would require less validation in the implementations to split them up, but not a big deal

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: This is up to the generator. The generator can decide to make things depend on other things and that doesn't necessarily need to be expressible in the DSL - the generator keeps it in memory and may make use of the knowledge while the test runs.

Quinn-With-Two-Ns: for example, signal A does not want to run past a certain point until the main workflow has run past a certain point.

I think wait for state would solve my ask.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: It could work that way, but that wasn't my plan. My thinking was to make the action set handler always return the state, and then be able to explore the other update conditions explicitly with these. I don't necessarily have any issue doing it that way though.

Quinn-With-Two-Ns:
The only hard blocker for me right now is how the `DoUpdate` works and a `wait-for-state` action

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Yeah, I forgot :). Will add.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Good question - I will add a comment. The idea is that the SDK will do this by distinguishing between cancelling _immediately_ (same WFT), or cancelling it in a subsequent WFT but before completion. A bit of thought might be needed here on how to make that work consistently from the generator / workflow impls.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+message ClientActionSet {
+  repeated ClientAction actions = 1;
+}
+
+message ClientAction {
+  oneof variant {
+    DoSignal do_signal = 1;
+    DoQuery do_query = 2;
+    DoUpdate do_update = 3;
+  }
```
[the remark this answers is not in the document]
Sushisource: Sure, that's resonable

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: I don't care about compat either. Same reason I gave before:

> The reason is that it’s a variant of a oneof. For normal fields, sure, but for oneofs the point is merely to have specified which variant. Especially since a lot of the languages don’t have true ADTs, this makes it really obvious that variant was explicitly chosen

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Yeah, wait for state will allow the generator to do arbitrary synchronization

workers/proto/kitchen_sink/kitchen_sink.proto
Quinn-With-Two-Ns: It  is important for update testing I can run multiple actions and then return a value or continue as new

workers/proto/kitchen_sink/kitchen_sink.proto
Quinn-With-Two-Ns: Ah missed that

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -0,0 +1,328 @@
...
+  repeated ActionSet initial_actions = 1;
+}
+
+// A set of actions to execute concurrently or sequentially. It is necessary to be able to represent
+// sequential execution without multiple 1-size action sets, as that implies the receipt of a signal
+// between each of those sets, which may not be desired.
+//
+// All actions are handled before proceeding to the next action set, unless one of those actions
+// would cause the workflow to complete/fail/CAN.
+message ActionSet {
```
Quinn-With-Two-Ns, line 101: Also a more general question on this DSL,  how  do I launch actions I don't wait on? ie launch a local activity async. and complete the workflow?

workers/proto/kitchen_sink/kitchen_sink.proto
cretz: Ah, ok. I don't really agree, I think a bool as a oneof makes plenty of sense and all languages have a way to see which variant was chosen. But I guess not a big deal.

workers/proto/kitchen_sink/kitchen_sink.proto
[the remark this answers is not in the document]
Sushisource: Actually, I changed my mind here. It needs to be the way it is in workflows to be able to express certain patterns without a signal in between - that isn't necessary here, and because we have to generate this I'd rather be inconsistent than have a technique that I know I'm just not going to bother using in the generator b/c of the added complexity. The desired patterns are representable without adding the concurrent flag or nested set, so fewer ways for the generator is better.

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Good Q. I changed the cancel condition to a more general awaitable choice. LMK what you think about that.

cretz:
> upgraded version of kitchen sink

I assume this means we can replace the kitchen sink with this instead of keeping both around?

Sushisource:
> > upgraded version of kitchen sink
> 
> I assume this means we can replace the kitchen sink with this instead of keeping both around (once the workflows are written of course)?

That's the idea

workers/proto/kitchen_sink/kitchen_sink.proto
Quinn-With-Two-Ns: Yeah I think that will work

Quinn-With-Two-Ns (approved):

PR #37, opened 2023-11-08
Fuzzy workflow generation

Sushisource:
Create a fuzzy workflow input generator & revamp the kitchen sink workflow definitions to work with it

All the features expressible in the DSL are not yet implemented in the generator or the workflows, but subsequent PRs will flesh that out


workers/python/kitchen_sink.py
```diff
@@ -1,269 +1,173 @@
...
+
+        # Run all initial input actions
+        if input and input.initial_actions:
+            for action_set in input.initial_actions:
+                return_value = await self.handle_action_set(action_set)
+                if return_value is not None:
                     return return_value
 
-        return return_value
+        # Run all actions from signals
```
Quinn-With-Two-Ns, line 39: Not directly related to your PR, but I think this approach of handling signals is not as valuable as handling the actions in the handler itself. There have been quite a few bugs in Java where something is broken in a signal handler, but not in the main workflow thread.

Quinn-With-Two-Ns:
Implementing the Java side, for `AwaitableChoice` I think we also need an `abandon` option.

Sushisource:
> Implementing the Java side, for `AwaitableChoice` I think we also need an `abandon` option. Also cancel after `x` duration.

There is "abandon after start". I don't think it really applies to any other time. We can add cancel after duration, but I'm not sure what new path it unlocks compared to cancel after start.

Quinn-With-Two-Ns:
>There is "abandon after start"
abandon after start is different then abandon though I thought?

Quinn-With-Two-Ns:
Fine punting on "cancel after duration"   until we come up with a case we need it. i just  suggested it because the old kitchen sink supported  it

Sushisource:
> > There is "abandon after start"
> 
> abandon after start is different then abandon though I thought?

There's abandon as in the abandon cancellation mode for activities and children (but that can already be chosen by the generator as the cancel mode), is that what you mean?

loadgen/kitchensink/helpers.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+type ClientActionsExecutor struct {
+	Client     client.Client
+	WorkflowID string
+	RunID      string
+}
+
+func (e *ClientActionsExecutor) ExecuteClientSequence(ctx context.Context, clientSeq *ClientSequence) error {
+	println("!!!!!!!! Number of action sets ", len(clientSeq.ActionSets))
```
Quinn-With-Two-Ns, line 46: Do we need these `println` here still? Can we convert them to proper them to use the same logger the rest of the Go code uses?

loadgen/kitchensink/helpers.go
```diff
@@ -0,0 +1,139 @@
...
+			_, err = e.Client.UpdateWorkflow(ctx, e.WorkflowID, e.RunID, handler.Name, handler.Args)
+		} else {
+			return fmt.Errorf("do_update must recognizable variant")
+		}
+		if update.FailureExpected {
+			err = nil
+		}
+	} else if query := action.GetDoQuery(); query != nil {
+		if query.GetReportState() != nil {
+			// TODO: Use args
```
Quinn-With-Two-Ns, line 123: Do you want an issue for this?

loadgen/scenario.go
```diff
@@ -168,14 +170,55 @@ func (r *Run) DefaultKitchenSinkWorkflowOptions() KitchenSinkWorkflowOptions {
...
+	if clientSeq != nil && len(clientSeq.ActionSets) > 0 {
+		executor := &kitchensink.ClientActionsExecutor{
+			Client:     r.Client,
+			WorkflowID: handle.GetID(),
+			RunID:      handle.GetRunID(),
+		}
+		go func() {
+			clientActionsErr = executor.ExecuteClientSequence(cancelCtx, clientSeq)
+			if clientActionsErr != nil {
+				fmt.Printf("Client actions failed: %v\n", clientActionsErr)
```
Quinn-With-Two-Ns, line 201: Can we use the logger?

loadgen/scenario.go
```diff
@@ -168,14 +170,55 @@ func (r *Run) DefaultKitchenSinkWorkflowOptions() KitchenSinkWorkflowOptions {
...
+				if err != nil {
+					return
+				}
+				cancel()
+			}
+		}()
+	}
+
+	executeErr := handle.Get(cancelCtx, nil)
+	cancel()
```
Quinn-With-Two-Ns, line 214: Can we  just `defer` this?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -74,141 +76,90 @@ func handleActionSet(
...
-	case action.Sleep != nil:
-		if err := workflow.Sleep(ctx, time.Duration(action.Sleep.Millis)*time.Millisecond); err != nil {
+	if rr := action.GetReturnResult(); rr != nil {
+		return true, rr.ReturnThis, nil
+	} else if re := action.GetReturnError(); re != nil {
+		return true, nil, temporal.NewApplicationError(re.Failure.Message, "")
+	} else if can := action.GetContinueAsNew(); can != nil {
+		return true, nil, workflow.NewContinueAsNewError(ctx, KitchenSinkWorkflow, can.Arguments)
+	} else if timer := action.GetTimer(); timer != nil {
+		if err := workflow.Sleep(ctx, time.Duration(timer.Milliseconds)*time.Millisecond); err != nil {
```
Quinn-With-Two-Ns, line 88: Timer action has an `AwaitableChoice` that is ignored

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -7,47 +7,47 @@ import (
 	"time"
 
 	"github.com/temporalio/omes/loadgen/kitchensink"
+	"go.temporal.io/api/common/v1"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/workflow"
 )
 
-func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowParams) (interface{}, error) {
+func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (interface{}, error) {
```
Quinn-With-Two-Ns, line 15: I think  we discussed before, but can the return values all be `payloads`?

workers/go/kitchensink/kitchen_sink.go
Quinn-With-Two-Ns: Oh do we never handle  `AwaitableChoice` in the Go workflow?

loadgen/kitchensink/helpers.go
Sushisource: I'll just get rid of these, had forgotten to clean them up

workers/go/go.mod
```diff
@@ -36,16 +37,22 @@ require (
...
+	google.golang.org/genproto/googleapis/api v0.0.0-20230822172742-b8732ec3820d // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230822172742-b8732ec3820d // indirect
 	google.golang.org/grpc v1.57.0 // indirect
 	google.golang.org/protobuf v1.31.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
 replace github.com/temporalio/omes => ../../
+replace (
+	github.com/temporalio/features/features => github.com/temporalio/features/features v0.0.0-20231117211247-ca7959c1fe2c
```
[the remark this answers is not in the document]
Sushisource: GoLand loves to choke hard on the worker code if these aren't in there for some reason, which is just lovely. I don't think they're necessary for compile though no, I can remove them.

workers/go/kitchensink/kitchen_sink.go
Sushisource: It's not handled in the python one either yet. This is one of the things I'll add in another PR.

loadgen/kitchensink/helpers.go
Sushisource: Nah, I'll just come in one of the follow up PRs where I add more features

workers/go/kitchensink/kitchen_sink.go
Quinn-With-Two-Ns: Works for me, to make my life easier can you mark the stuff you plan to do in follow up PRs with `TODO` or something? It is hard to tell if something is planned to be done but just hasn't, was forgotten, or I misunderstood a feature.

Sushisource: That's going to be quite a lot of stuff - essentially everything in the DSL that isn't currently emitted anywhere by the generator.

If something's in the generator, the workflows need to handle it since it could be generated. I think the generator is reasonably easy to read. Does that work?

Quinn-With-Two-Ns: I understand, yeah I can work with that

workers/python/kitchen_sink.py
```diff
@@ -1,269 +1,195 @@
...
-            action_set_queue: asyncio.Queue[ActionSet] = asyncio.Queue()
-
-            def enqueue(item: ActionSet) -> None:
-                action_set_queue.put_nowait(item)
+    action_set_queue: asyncio.Queue[ActionSet] = asyncio.Queue()
+    workflow_state = WorkflowState()
+
+    @workflow.signal
+    async def do_actions_signal(self, action_set: ActionSet) -> None:
+        self.action_set_queue.put_nowait(action_set)
```
Quinn-With-Two-Ns, line 30: Didn't  we say we want to treat signal like update and allow for the action set to be run in the signal handler?

Sushisource: Yes, that's right, I forgot to go back and add that. We talked about the DSL having an option to either pipe back to main to to run in-handler.

Quinn-With-Two-Ns (approved):

PR #39, opened 2023-11-22
Introduce --auth-header option along with TEMPORAL_OMES_AUTH_HEADER env var

aromanovich:
## What was changed
This PR adds the `--authorization-env-var` option.

## Why?
It can be used as `--authorization-env-var TEMPORAL_CLI_AUTH` to authenticate Omes with a token stored in the `TEMPORAL_CLI_AUTH` env var, which would be passed in `Authorization` gRPC metadata.

## Checklist

1. How was this tested:
I ran some benchmarks using `run-scenario` command.

2. Any docs updates needed?
Probably not. I suppose the description in `--help` output is enough.

### DIFF
--- a/cmd/cmdoptions/client.go
+++ b/cmd/cmdoptions/client.go
@@ -1,13 +1,30 @@
 package cmdoptions
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"os"
+
 	"github.com/spf13/pflag"
 	"go.temporal.io/sdk/client"
 	"go.uber.org/zap"
 )
+
+const AUTH_HEADER_ENV_VAR = "TEMPORAL_OMES_AUTH_HEADER"
+
+type authHeaderProvider struct {
+	value string
+}
+
+func (p *authHeaderProvider) GetHeaders(ctx context.Context) (map[string]string, error) {
+	return map[string]string{"Authorization": p.value}, nil
+}
+
+func newAuthHeaderProvider(value string) *authHeaderProvider {
+	return &authHeaderProvider{value: value}
+}
 
 // Options for creating a Temporal client.
 type ClientOptions struct {
@@ -21,6 +38,8 @@ type ClientOptions struct {
 	ClientCertPath string
 	// TLS client private key
 	ClientKeyPath string
+	// Authorization header value
+	AuthHeader string
 }
 
 // loadTLSConfig inits a TLS config from the provided cert and key files.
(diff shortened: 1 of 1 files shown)


cmd/cmdoptions/client.go
```diff
@@ -1,14 +1,29 @@
...
 	"go.temporal.io/sdk/client"
 	"go.uber.org/zap"
 )
 
+type authTokenProvider struct {
+	token string
+}
+
+func (p *authTokenProvider) GetHeaders(ctx context.Context) (map[string]string, error) {
+	return map[string]string{"Authorization": p.token}, nil
```
Sushisource, line 20: You might want to move the getenv call to be in here so, if necessary, the value can be changed without a restart

cretz:
Can you help me understand why the authorization header cannot be provided to the CLI directly? If you need to use an env var in your shell for that value you can.

Or even better, accept a way to provide arbitrary headers to the CLI.

aromanovich:
It was my impression that using an env var is a somewhat conventional way to pass the authorization header to Temporal-related CLIs:
* tctl v1 supported `TEMPORAL_CLI_AUTH`
* tctl2's [tctl-authorization-plugin](https://github.com/temporalio/tctl/blob/a145061f0eb3b2bdaaf075cbd96c747753f112c5/cmd/plugins/tctl-authorization-plugin/main.go#L50) supported `TEMPORAL_CLI_AUTHORIZATION_TOKEN`;
*  there is an issue to support it in new CLI as well (but the name is not decided yet): https://github.com/temporalio/cli/issues/342.

Passing a token as a CLI argument leaks the token to places such as `ps` output and shell history.
Using environment variables is [far from perfect](https://smallstep.com/blog/command-line-secrets/#environment-variables) either, but still provides more security in some environments. In our case the env var is set by a trusted supervisor which communicates to a secret manager, which leaves no traces of the token in world readable places.

cretz:
> It was my impression that using an env var is a somewhat conventional way to pass the authorization header to Temporal-related CLIs

Nah, that's why those env vars are very `TEMPORAL_CLI_` specific, so they don't clash. But I guess if there is concern of using the CLI params, we could have general support for env vars. For this specific use case, `TEMPORAL_OMES_AUTH` env var can be implicitly supported (no arg needed), but we might as well also add a `--auth-header` arg for those that don't want to be forced to use env vars (and don't share your shall concerns, e.g. in CI runs).

cmd/cmdoptions/client.go
```diff
@@ -64,6 +83,11 @@ func (c *ClientOptions) Dial(metrics *Metrics, logger *zap.SugaredLogger) (clien
 	clientOptions.ConnectionOptions.TLS = tlsCfg
 	clientOptions.Logger = NewZapAdapter(logger.Desugar())
 	clientOptions.MetricsHandler = metrics.NewHandler()
+	authHeader := os.Getenv(AUTH_HEADER_ENV_VAR)
+	if c.AuthHeader != "" {
+		authHeader = c.AuthHeader // CLI argument value overrides env var value
+	}
+	clientOptions.HeadersProvider = newAuthHeaderProvider(authHeader)
```
cretz, line 90: We should only set a header provider if there is an auth header.

Granted we'll probably be coming back and changing this anyways if/when we ever want any other header than just this one special one. Arguably you should just have a `headerProvider` that has an entire map of headers to set instead of assuming this is the only use of header provider we'll ever need.

cretz (commented):
LGTM mostly, just one more note.

cretz (approved):
Thanks!

PR #41, opened 2023-12-01
More Fuzzy Features

Sushisource:
## What was changed
Implement more fuzzy workflow features

Unfortunately we need https://github.com/temporalio/sdk-python/pull/440 released or need to temporarily depend on an unreleased version for this to work.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz:
> Unfortunately we need https://github.com/temporalio/sdk-python/pull/440 released or need to temporarily depend on an unreleased version for this to work.

Is it possible to add a scenario+SDK skip somehow?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -116,25 +124,61 @@ func handleAction(
...
 		return nil, fmt.Errorf("unrecognized action")
 	}
 	return nil, nil
 }
 
 func launchActivity(ctx workflow.Context, act *kitchensink.ExecuteActivityAction) error {
+	args := act.GetArguments()
+	if len(args) == 0 {
+		args = nil
+	}
```
cretz, line 171: Is this needed?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -152,8 +196,45 @@ func launchActivity(ctx workflow.Context, act *kitchensink.ExecuteActivityAction
...
 		actCtx := workflow.WithActivityOptions(ctx, opts)
-		return workflow.ExecuteActivity(actCtx, act.ActivityType, act.Arguments).Get(actCtx, nil)
+		return withAwaitableChoice(actCtx, func(ctx workflow.Context) workflow.Future {
+			return workflow.ExecuteActivity(ctx, act.ActivityType, args)
+		}, act.GetAwaitableChoice())
+	}
+}
+
+func withAwaitableChoice(ctx workflow.Context, starter func(workflow.Context) workflow.Future,
+	awaitChoice *kitchensink.AwaitableChoice) error {
```
cretz, line 206:
```suggestion
func withAwaitableChoice(
	ctx workflow.Context,
	starter func(workflow.Context) workflow.Future,
	awaitChoice *kitchensink.AwaitableChoice,
) error {
```

A bit more idiomatic to do it like this when spanning multiple lines in a signature

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -152,8 +196,45 @@ func launchActivity(ctx workflow.Context, act *kitchensink.ExecuteActivityAction
...
+	var err error
+	didCancel := false
+	if awaitChoice.GetAbandon() != nil {
+		return nil
+	} else if awaitChoice.GetCancelBeforeStarted() != nil {
+		cancel()
+		didCancel = true
+		err = fut.Get(ctx, nil)
+	} else if awaitChoice.GetCancelAfterStarted() != nil {
+		_ = workflow.Sleep(ctx, time.Duration(1))
```
cretz, line 218:
```suggestion
		_ = workflow.Sleep(ctx, 1)
```

workers/python/kitchen_sink.py
```diff
@@ -104,16 +111,55 @@ async def handle_action(self, action: Action) -> Optional[Payload]:
...
+                lambda: self.workflow_state.kvs.get(action.await_workflow_state.key)
+                == action.await_workflow_state.value
+            )
+        elif action.HasField("upsert_memo"):
+            pass  # Python doesn't have memo upserting
+        elif action.HasField("upsert_search_attributes"):
+            updates: list[SearchAttributeUpdate[Any]] = []
+            # TODO: Use RawValue after https://github.com/temporalio/sdk-python/issues/438
+            #  and avoid checking key by name
+            for k, v in action.upsert_search_attributes.search_attributes.items():
```
cretz, line 155: We need to include the search attribute type with each search attribute item in this action IMO. Nowadays, search attribute setting needs to have key, value, _and_ type every time to avoid ambiguity only the server can resolve.

workers/python/kitchen_sink.py
```diff
@@ -168,10 +214,44 @@ def launch_activity(execute_activity: ExecuteActivityAction) -> Awaitable:
                 execute_activity.remote.cancellation_type
             ),
         )
-    # TODO: Handle cancels
+
     return activity_task
 
 
+async def handle_awaitable_choice(
+    awaitable: Union[Coroutine | asyncio.Task], choice: AwaitableChoice
```
cretz, line 222:
```suggestion
    awaitable: Coroutine | asyncio.Task, choice: AwaitableChoice
```

or

```suggestion
    awaitable: Union[Coroutine, asyncio.Task], choice: AwaitableChoice
```

Not both. Though really, I think `Awaitable[Any]` is best

workers/python/kitchen_sink.py
```diff
@@ -168,10 +214,44 @@ def launch_activity(execute_activity: ExecuteActivityAction) -> Awaitable:
...
+            await task
+        elif choice.HasField("cancel_after_completed"):
+            await task
+            task.cancel()
+            did_cancel = True
+        else:
+            await task
+    except asyncio.CancelledError:
+        if not did_cancel:
+            raise exceptions.ApplicationError("Unexpected cancellation")
```
cretz, line 252:
```suggestion
            raise
```

Might as well just re-raise the same thing like Go

Sushisource:
> Is it possible to add a scenario+SDK skip somehow?

Turns out it doesn't matter anyway since we're not running full fuzzer in CI yet.

workers/go/kitchensink/kitchen_sink.go
Sushisource: Not anymore. It was while the pass through converter I added wasn't working right.

workers/python/kitchen_sink.py
Sushisource: You mean add the type to the proto? Could do that. Testing all those isn't really that important here though, which is why I wanted to just pass raw payload through, since we're not asserting anything equals anything else really, just sometimes generating one of these commands.

cretz: One of the primary reasons we did typed search attributes is because nowadays, we require more than just key+value to set a search attribute. We require key+type+value.

workers/python/kitchen_sink.py
```diff
@@ -104,16 +111,55 @@ async def handle_action(self, action: Action) -> Optional[Payload]:
...
-            await workflow.execute_child_workflow(
-                child, id=child_action.workflow_id, args=args
+            await handle_awaitable_choice(
+                workflow.start_child_workflow(
+                    child, id=child_action.workflow_id, args=args
+                ),
+                child_action.awaitable_choice,
             )
+        elif action.HasField("set_patch_marker"):
+            if action.set_patch_marker.deprecated:
```
Quinn-With-Two-Ns, line 134: How is this supposed to work with Go and Java?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -116,25 +124,61 @@ func handleAction(
...
+		err := withAwaitableChoice(ctx, func(ctx workflow.Context) workflow.Future {
+			return workflow.ExecuteChildWorkflow(ctx, childType, child.GetInput()[0])
+		}, child.AwaitableChoice)
+		return nil, err
+	} else if patch := action.GetSetPatchMarker(); patch != nil {
+		if workflow.GetVersion(ctx, patch.GetPatchId(), workflow.DefaultVersion, 1) == 1 {
+			return handleAction(ctx, workflowState, patch.GetInnerAction())
+		}
+	} else if setWfState := action.GetSetWorkflowState(); setWfState != nil {
+		workflowState = setWfState
```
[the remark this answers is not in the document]
cretz: That's in an `else` clause. This local variable assignment (even though the variable is defined in parameter) is effectively the last thing run in this function. If you want this assignment to apply to what the pointer is pointing to, I think you need:

```suggestion
		*workflowState = *setWfState
```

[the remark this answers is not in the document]
Sushisource: Yeah, duh, that makes sense. I assumed it must've somehow been doing the right thing still because nothing was hanging but I guess I somehow never tripped a test that had a set and then await.

Deref on both sides complains about copying a mutex inside the proto thing. So I'll have to clone it (which is also annoying because of the type casting) or have a double pointer so I can just reassign the real pointer.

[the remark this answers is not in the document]
Sushisource: Double pointer is the simplest option.

workers/python/kitchen_sink.py
Sushisource: Just going to keep using the string check since the attrs have to be predefined and doing this is essentially the same amount of work & the generator/dsl doesn't actually care about different types.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -116,25 +124,58 @@ func handleAction(
 		if child.WorkflowType != "" {
 			childType = child.WorkflowType
 		}
-		err := workflow.ExecuteChildWorkflow(ctx, childType, child.GetInput()).Get(ctx, nil)
+		err := withAwaitableChoice(ctx, func(ctx workflow.Context) workflow.Future {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I would just do it properly rather then try to rely on timing to cover all the states

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -96,18 +98,24 @@ func handleActionSet(
...
 		return nil, temporal.NewApplicationError(re.Failure.Message, "")
 	} else if can := action.GetContinueAsNew(); can != nil {
 		return nil, workflow.NewContinueAsNewError(ctx, KitchenSinkWorkflow, can.Arguments)
 	} else if timer := action.GetTimer(); timer != nil {
-		if err := workflow.Sleep(ctx, time.Duration(timer.Milliseconds)*time.Millisecond); err != nil {
-			return nil, err
-		}
+		return nil, withAwaitableChoice(ctx, func(ctx workflow.Context) workflow.Future {
+			fut, setter := workflow.NewFuture(ctx)
+			workflow.Go(ctx, func(ctx workflow.Context) {
```
Quinn-With-Two-Ns, line 113: Why not just use a timer?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -152,8 +193,48 @@ func launchActivity(ctx workflow.Context, act *kitchensink.ExecuteActivityAction
...
+		cancel()
+		didCancel = true
+		err = fut.Get(ctx, nil)
+	} else if awaitChoice.GetCancelAfterStarted() != nil {
+		_ = workflow.Sleep(ctx, 1)
+		cancel()
+		didCancel = true
+		err = fut.Get(ctx, nil)
+	} else if awaitChoice.GetCancelAfterCompleted() != nil {
+		res := fut.Get(ctx, nil)
```
Quinn-With-Two-Ns, line 223: This logic seems wrong. If we are canceling after getting the result I would not expect the error to be a `CanceledError ` it should be nil.

workers/python/kitchen_sink.py
Sushisource: I really don't think it's an issue - keep in mind these tests are not about verifying behavioral specifics. All that matters here is that we generate the patch/versioning command (marker).

workers/go/kitchensink/kitchen_sink.go
Sushisource: Instead of what? The timer needs to be cancellable somehow and the only way to do that in conjunction with the different awaitable choices is to put it in another routine

workers/go/kitchensink/kitchen_sink.go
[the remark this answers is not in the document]
Sushisource: I guess? :man_shrugging: I'm honestly not sure the extra code is really worth it. The point of fuzzing tests is to generate wide coverage probabilistically - that's gonna happen this way. For the other command types we have no choice anyway, so we only get to be specific for child workflows.

workers/go/kitchensink/kitchen_sink.go
Quinn-With-Two-Ns: Instead of this sleep in a goroutine, it is very confusing to read when you can just create a timer and treat it's  future like an activity future.

workers/go/kitchensink/kitchen_sink.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah but we can test the proper scenario with child workflows so we should so we know we are testing those state transitions rather then relying on timing.

cretz (approved):

workers/go/kitchensink/kitchen_sink.go
Sushisource: The timer doesn't have a future is the problem

workers/go/kitchensink/kitchen_sink.go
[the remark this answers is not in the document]
Sushisource: Yeah, I considered changing to the struct, I should just do that.

Quinn-With-Two-Ns (approved):

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -330,13 +436,53 @@ impl<'a> Arbitrary<'a> for ActionSet {
...
 impl<'a> Arbitrary<'a> for Action {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
-        // TODO: Adjustable ratio of choice?
-        // TODO: The rest of the kinds of actions
-        let action_kind = u.int_in_range(0..=1)?;
-        let variant = match action_kind {
-            0 => action::Variant::Timer(u.arbitrary()?),
-            1 => action::Variant::ExecActivity(u.arbitrary()?),
-            _ => unreachable!(),
+        let action_kind = u.int_in_range(0..=100)? as f32;
```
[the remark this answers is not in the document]
Sushisource: Good points, though neither is I think is hugely important. I'll fix the clamping thing - the choosing thing maybe not just b/c the performance isn't an issue.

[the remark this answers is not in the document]
dandavison: Yes, the probability clamping should be fixed I think, but agree that from a performance POV the other change wouldn't affect any workloads we have planned. It should allow us to get rid of the macro-generated methods though, which is always a bit of a readability boost.

PR #42, opened 2023-12-04
Java worker kitchen sink

Quinn-With-Two-Ns:
Add Java omes worker and kitchen sink workflow


workers/java/io/temporal/omes/Main.java
```diff
@@ -0,0 +1,250 @@
...
+        throw new RuntimeException("Error loading certs", e);
+      }
+
+    } else if (StringUtils.isNotEmpty(clientKeyPath) && StringUtils.isEmpty(clientCertPath)) {
+      throw new RuntimeException("Client cert path must be specified since key path is");
+    } else if (isTlsEnabled) {
+      try {
+        sslContext = SimpleSslContextBuilder.noKeyOrCertChain().build();
+      } catch (SSLException e) {
+        // TODO Auto-generated catch block
```
Sushisource, line 145: ?

loadgen/kitchensink/kitchen_sink.pb.go
```diff
@@ -3393,11 +3393,12 @@ var file_kitchen_sink_proto_rawDesc = []byte{
...
-	0x2e, 0x67, 0x69, 0x74, 0x68, 0x75, 0x62, 0x2e, 0x63, 0x6f, 0x6d, 0x2f, 0x74, 0x65, 0x6d, 0x70,
-	0x6f, 0x72, 0x61, 0x6c, 0x69, 0x6f, 0x2f, 0x6f, 0x6d, 0x65, 0x73, 0x2f, 0x6c, 0x6f, 0x61, 0x64,
-	0x67, 0x65, 0x6e, 0x2f, 0x6b, 0x69, 0x74, 0x63, 0x68, 0x65, 0x6e, 0x73, 0x69, 0x6e, 0x6b, 0x62,
-	0x06, 0x70, 0x72, 0x6f, 0x74, 0x6f, 0x33,
+	0x12, 0x0b, 0x0a, 0x07, 0x41, 0x42, 0x41, 0x4e, 0x44, 0x4f, 0x4e, 0x10, 0x02, 0x42, 0x42, 0x0a,
+	0x10, 0x69, 0x6f, 0x2e, 0x74, 0x65, 0x6d, 0x70, 0x6f, 0x72, 0x61, 0x6c, 0x2e, 0x6f, 0x6d, 0x65,
+	0x73, 0x5a, 0x2e, 0x67, 0x69, 0x74, 0x68, 0x75, 0x62, 0x2e, 0x63, 0x6f, 0x6d, 0x2f, 0x74, 0x65,
+	0x6d, 0x70, 0x6f, 0x72, 0x61, 0x6c, 0x69, 0x6f, 0x2f, 0x6f, 0x6d, 0x65, 0x73, 0x2f, 0x6c, 0x6f,
+	0x61, 0x64, 0x67, 0x65, 0x6e, 0x2f, 0x6b, 0x69, 0x74, 0x63, 0x68, 0x65, 0x6e, 0x73, 0x69, 0x6e,
+	0x6b, 0x62, 0x06, 0x70, 0x72, 0x6f, 0x74, 0x6f, 0x33,
```
dandavison, line 3401: Just curious what this change is -- why does adding a Java worker cause a modification to the KS proto generated Go code?

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
```diff
@@ -0,0 +1,457 @@
...
+
+public class KitchenSinkWorkflowImpl implements KitchenSinkWorkflow {
+  public static final Logger log = Workflow.getLogger(KitchenSinkWorkflowImpl.class);
+  Map<String, String> state = new HashMap<>();
+  WorkflowQueue<KitchenSink.ActionSet> signalActionQueue = Workflow.newWorkflowQueue(100);
+
+  @Override
+  public Payload execute(KitchenSink.WorkflowInput input) {
+    // Run all initial input actions
+    if (input != null && input.getInitialActionsList().size() > 0) {
```
dandavison, line 33: nit: no need to check size, just let the for loop run over empty container?

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
```diff
@@ -0,0 +1,457 @@
...
+          action.getUpsertSearchAttributes();
+      Workflow.upsertSearchAttributes(upsertSearchAttributes.getSearchAttributes());
+    } else if (action.hasSetPatchMarker()) {
+      KitchenSink.SetPatchMarkerAction patchMarker = action.getSetPatchMarker();
+      if (Workflow.getVersion(patchMarker.getPatchId(), -1, 1) == 1) {
+        handleAction(patchMarker.getInnerAction());
+      }
+    } else if (action.hasUpsertMemo()) {
+      // TODO(https://github.com/temporalio/sdk-java/issues/623) Java does not support upsert memo.
+      log.warn("Java does not support upsert memo");
```
dandavison, line 202: I wonder whether this warning is going to be more annoying than helpful? We're intentionally submitting upsert-memo actions to all langs, so the statement that sdk-java doesn't support them isn't really a warning that something is wrong with the omes run.

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
```diff
@@ -0,0 +1,457 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+
+public class KitchenSinkWorkflowImpl implements KitchenSinkWorkflow {
+  public static final Logger log = Workflow.getLogger(KitchenSinkWorkflowImpl.class);
+  Map<String, String> state = new HashMap<>();
+  WorkflowQueue<KitchenSink.ActionSet> signalActionQueue = Workflow.newWorkflowQueue(100);
```
dandavison, line 28: How was 100 chosen and what are the consequences of this value?

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
Quinn-With-Two-Ns: Yeah. have no issue lowering the log level

loadgen/kitchensink/kitchen_sink.pb.go
Quinn-With-Two-Ns: This is the proto file description and the proto file did change, it had a Java option added, so I assume that is reflected here.

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
Quinn-With-Two-Ns: The only consequence if we hit the queue limit is the signal handler threads will block on inserting

Sushisource (approved):

PR #49, opened 2023-12-14
Typescript Support & KS Workflow

Sushisource:
Still draft while merging required `features` changes & also hooking up to CI


workers/typescript/src/payload-converter.ts
```diff
@@ -0,0 +1,58 @@
...
+} from '@temporalio/common';
+import { decode, encode } from '@temporalio/common/lib/encoding';
+import Payload = temporal.api.common.v1.Payload;
+
+export class PassThroughPayload implements PayloadConverterWithEncoding {
+
+  public toPayload(value: any): Payload | undefined {
+    if (!value || value.metadata === undefined || value.data === undefined) {
+      return undefined;
+    }
```
Sushisource, line 22: If there's a less stupid way to do this, please tell me. `instanceof` doesn't work. `Payload.verify` might be best, but we're not codgen-ing that right now.

workers/typescript/src/activities.ts
```diff
@@ -0,0 +1,7 @@
+export async function noop() {
+  return undefined;
+}
+
+export async function delay(delay_for_ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, delay_for_ms));
+}
```
bergundy, line 7:
```suggestion
export { sleep as delay } from '@temporalio/activity'
```

Or if you don't want the activity to respond to cancelation:

```ts
export { setTimeout as delay } from 'timers/promises'
```

workers/typescript/src/omes.ts
```diff
@@ -0,0 +1,98 @@
+import { Command } from 'commander';
```
bergundy, line 1: nit: [arg](https://www.npmjs.com/package/arg) is much lighter and IMHO nicer to use.

bergundy: Looks like it doesn't support help strings, so I'm not pushing here.

workers/typescript/src/omes.ts
```diff
@@ -0,0 +1,98 @@
...
+    dataConverter: {
+      payloadConverterPath: require.resolve('./payload-converter')
+    }
+  });
+
+  await worker.run();
+}
+
+run().then(() => {
+  process.exit(0);
```
bergundy, line 94: wondering why you need this, is the process not existing when `run` resolves?

workers/typescript/src/payload-converter.ts
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      asPayload = Payload.fromObject(value as any);
+    } catch (e) {
+      throw new ValueError('PassThroughPayload can only convert Payloads');
+    }
+    const asBytes = Payload.encode(asPayload).finish();
+    return Payload.create({
+      metadata: {
+        encoding: encode(this.encodingType)
+      }, data: asBytes
```
bergundy, line 33: can you add a prettier check to verify that code is properly formatted (same goes for linter).

workers/typescript/src/protos/json-module.js
bergundy, line 1: We don't commit generated code in the TS repo, I prefer not committing it but I'm okay with whatever you choose.

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -0,0 +1,295 @@
...
+
+const reportStateQuery = defineQuery<IWorkflowState, [Payload]>('report_state');
+const actionsSignal = defineSignal<[DoSignalActions]>('do_actions_signal');
+const actionsUpdate = defineUpdate<IPayload | undefined, [DoActionsUpdate]>('do_actions_update');
+
+export async function kitchenSink(input: WorkflowInput | undefined): Promise<IPayload | undefined> {
+  let workflowState: IWorkflowState = WorkflowState.create();
+  const actionsQueue = new Array<IActionSet>();
+
+  async function handleActionSet(actions: IActionSet): Promise<IPayload | null> {
```
bergundy, line 53: nit: prefer `undefined` over `null`, `undefined` payload converter is the equivalent of nil payload converter in sane languages that have somewhat better null representation.

It also will make it compatible with `actionsUpdate`'s return value.

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+      }));
+    }
+    const allComplete = Promise.all(promises);
+    await Promise.any([allComplete, condition(() => rval !== undefined)]);
+
+    return rval;
+  }
+
+  async function handleAction(action: IAction): Promise<IPayload | null | undefined> {
```
bergundy, line 81: Not sure I understand why you need both null and undefined here.

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -0,0 +1,295 @@
...
+        await sleep(1);
+      },
+      afterCompleted: (_: Promise<PRR | void>) => Promise<void> = async (task) => {
+        await task;
+      }
+    ) {
+      const cancelScope = new CancellationScope();
+      let didCancel = false;
+
+      const cancellablePromise = cancelScope.run(() => promise).catch((err) => {
```
bergundy, line 96: The promise should be created in the cancellation scope for it to be attached to it, otherwise the command that the promise represents will not be cancelled.

workers/typescript/src/omes.ts
Sushisource: It was not no

workers/typescript/src/protos/json-module.js
Sushisource: I prefer not committing either, but I did for the other langs since Rust generates their protos and I didn't want you to need Rust just to build and run fuzzer repros. That doesn't apply here, so yeah I can avoid checking it in.

workers/typescript/src/payload-converter.ts
Sushisource: Linter is already linting but I'll add a prettier check

workers/typescript/src/workflows/kitchen_sink.ts
Sushisource: Because I return things straight from proto and that's how they're typed there

Quinn-With-Two-Ns (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -25,9 +25,9 @@ jobs:
       - name: Run local scenario with worker
         run: ./temporal-omes run-scenario-with-worker --scenario workflow_with_single_noop_activity --log-level debug --language go --embedded-server --iterations 5
       - name: Build worker image
-        run: ./temporal-omes build-worker-image --language go --version v1.24.0 --tag-as-latest
+        run: ./temporal-omes build-worker-image --language go --version v1.25.2-0.20231129171107-288a04f72145 --tag-as-latest
```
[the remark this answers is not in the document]
Sushisource: I needed to make changes to `features` to build properly, but `features` is already updated to use the new (de-gogoproto'd) API, so I needed to update the Go stuff to work with that.

PR #53, opened 2023-12-18
Expose RegisterScenarios()

dandavison:
Two alterations made in order to use omes as a library in our CI/CD pipelines (i.e. when using omes as the starter, as opposed to the worker):

1. ~~Expose the `ScenarioRunner` struct and its `Run` method.~~
2. Expose a function that a library can call to register all scenarios.

Here is the (PR not open yet) CI/CD code making use of these changes:

https://github.com/temporalio/saas-cicd/blob/sdk-1421-omes/oss-cicd/e2e/workflows/test_workload/activities/omes.go#L38-L54

Also, some README edits.


scenarios/register.go
```diff
@@ -0,0 +1,7 @@
+package scenarios
+
+// RegisterScenarios registers all scenarios. It is intended to be called by Go
+// applications using omes as a library to run scenarios.
+func RegisterScenarios() {
+	// Scenarios are registered automatically on import of this package.
+}
```
Sushisource, line 7: Rather than exposing this and needing the caller to call it, maybe it'd make more sense to do a `NewScenarioRunner` which accepts the options, calls this, and returns the runner?

dandavison: The scenarios are currently loaded automatically for the CLI caller by Go's package `init()` mechanism, so it's only library callers that need a way to load them. I did it this way because I thought that the authors of the library might not appreciate a redundant call being placed in a code path when loading was already handled by the `init` mechanism. WDYT?

dandavison: Also I think the import graph might be showing that this would go against the design: `scenarios` imports `loadgen`, so I don't think there's anywhere we can import `scenarios` in the CLI-start-up sequence.

cmd/cmdoptions/scenariorunner.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+	"time"
+
+	"github.com/spf13/pflag"
+	"github.com/temporalio/omes/loadgen"
+	"go.temporal.io/sdk/client"
+	"go.uber.org/zap"
+)
+
+type ScenarioRunner struct {
```
cretz, line 17: I do not think we should use CLI code outside of the CLI

cmd/cmdoptions/scenariorunner.go
```diff
@@ -0,0 +1,117 @@
+package cmdoptions
```
cretz, line 1: I think the existing `loadgen` package should be the package used programmatically, we don't need a whole new package

cmd/cmdoptions/scenariorunner.go
```diff
@@ -0,0 +1,117 @@
...
+	fs.IntVar(&r.MaxConcurrent, "max-concurrent", 0, "Override max-concurrent for the scenario")
+	fs.StringSliceVar(&r.ScenarioOptions, "option", nil, "Additional options for the scenario, in key=value format")
+	fs.DurationVar(&r.ConnectTimeout, "connect-timeout", 0, "Duration to try to connect to server before failing")
+	r.ClientOptions.AddCLIFlags(fs)
+	r.MetricsOptions.AddCLIFlags(fs, "")
+	r.LoggingOptions.AddCLIFlags(fs)
+}
+
+// Run runs the scenario.
+func (r *ScenarioRunner) Run(ctx context.Context) error {
```
cretz, line 46: I know coming from other languages what's happening here may _look_ like a lot of logic, but it's not really. Just `loadgen.GetScenario` + `loadgen.ScenarioInfo` creation + `scenario.Executor.Run`. There's not need to pull all of this out from where it was and try to reuse it.

Can you show the code that will use this so we can discuss alternatives?

scenarios/register.go
cretz, line 1: This is an unnecessary file with a no-op function. If this function is going to get filled in later, don't commit it until that's the case

cmd/cmdoptions/scenariorunner.go
dandavison: > Can you show the code that will use this so we can discuss alternatives?

Please see the PR introduction.

scenarios/register.go
dandavison: The function is called by the library caller here: https://github.com/temporalio/saas-cicd/blob/ce7b279dc71951086c307f41eb9c82e8f1deba08/oss-cicd/e2e/workflows/test_workload/activities/omes.go#L38?

The purpose of the function is documented in the docstring, and the fact that its body is empty is explained in the comment in the code. See also the discussion below in the PR regarding how library callers should ensure that scenarios are registered.

cmd/cmdoptions/scenariorunner.go
dandavison: That's the purpose of this PR -- this PR changes the role of `ScenarioRunner` so that it can be used by library callers. So, when reviewing the PR, we can consider whether that's a good move or not, but it cannot be a premise of the review that it is not :)

cmd/cmdoptions/scenariorunner.go
dandavison: Let me make an alternative version of the `sass-cicd` caller that does not rely on exposing `ScenarioRunner` in omes, to see what it looks like.

dandavison: I agree with you @cretz. Duplicating the 50 lines or so between omes CLI and the cicd runner is the right decision, rather than exposing new abstractions in omes that there's not an urgent need for. I've done that here https://github.com/dandavison/temporalio-saas-cicd/blob/sdk-1421-omes/oss-cicd/e2e/workflows/test_workload/activities/omes.go and reverted that part of this PR.

scenarios/register.go
dandavison: I believe that it is necessary to expose the `RegisterScenarios()` function as this PR is doing. The only alternative that I see as things stand would be as below, but we can't require library callers to do this:

```go
_ "github.com/temporalio/omes/scenarios" // register scenarios by side-effect
```

scenarios/register.go
Sushisource: Bleh, Go imports. This is fine, it's no big deal. It seems a touch unnecessary, so if we could make it happen automatically that'd be great (I don't think having a redundant call matters, it's idempotent).

Sushisource (approved):

scenarios/register.go
```diff
@@ -0,0 +1,7 @@
+package scenarios
+
+// RegisterScenarios registers all scenarios. It is intended to be called by Go
+// applications using omes as a library to run scenarios.
+func RegisterScenarios() {
```
cretz, line 5: Can this method be removed? If including a Go package has side effects, it can be documented at the package level and users can import it for those side effects.

dandavison: So, the options (as seen from the library user's point of view) are:

### A
```go
import (
    "github.com/temporalio/omes/scenarios"
)
scenarios.RegisterScenarios()
```

and

### B
```go
import (
    _ "github.com/temporalio/omes/scenarios" // register scenarios by side-effect
)
```

IMO **A** is better library design for two reasons:
1. It's natural and explicit
2. It hides the implementation detail and allows us to change the way that scenarios are registered in the future. I.e., the fact that scenario registration happens as a side-effect of import should be considered an implementation detail of the library; we don't want changing that to be a breaking change for users.

cretz: > IMO A is better library design for two reasons:

But it's not really about personal preference on these Go things. In Go, imports with side effects are usually done via a side-effecting import, not a no-op method. (even if it was a method, it would be named something that it doesn't do, "register" is inaccurate because it is done whether you call the method or not)

But I would totally support changing the scenarios to not register on `init()` but instead only register in `RegisterScenarios` and just use `init()` to build up a slice of them to be registered in that function. Just don't forget to call `RegisterScenarios` for the CLI too.

dandavison: > But it's not really about personal preference on these Go things. In Go, imports with side effects are usually done via a side-effecting import, not a no-op method. (even if it was a method, it would be named something that it doesn't do, "register" is inaccurate because it is done whether you call the method or not)

I think you're misunderstanding slightly -- all that is deliberate. My proposal here was to deliberately introduce a no-op method that says it is registering scenarios when in fact it is doing nothing and the scenarios were registered already. `RegisterScenarios` was an accurate name in a forwards-compatibility sense because it was papering over a bug in the library that should be fixed in the future -- i.e. the registration should be moved into the function.

The reason I was proposing that we do this is that a library should never require users of the library to perform an import for side-effects only. It's fine to use Go's import side effect feature internally, but a library telling users that, in order to use the library, they must import something and not use it, is something that should be avoided.

That said, I agree that it would be better to make the public interface match up with the implementation, instead of going through a transition period in which a public function lies about what it is doing. Furthermore there's no real need to address this now, since the only known user of omes as a library is Temporal's CI/CD framework, which is written by the authors of omes. Or put another way, this defect in omes-as-a-library has been released to the public for a while and there's nothing about current work on internal CI/CD that makes it urgent to address the library design issue. So I'll revert this change and merge the PR with the random docstring/README edits only.

cretz: :+1: Definitely don't want a no-op for something that an import does. If we need to move from side-effecting import to function-based registration we can (and if you want to wait until some future time, no prob)

> It's fine to use Go's import side effect feature internally, but a library telling users that, in order to use the library, they must import something and not use it, is something that should be avoided.

This is not necessarily true in Go. Quite often side-effecting imports are used in libraries (e.g. k8s, protobuf, etc). Having said that, I have no problem with the more explicit approach.

> this defect in omes-as-a-library has been released to the public

Not really a "defect" and we should not have any backwards compatibility guarantees in omes libraries in any way. Omes should be able to be completely changed/refactored at will to whatever is best as needs arise.

> So I'll revert this change and merge the PR with the random docstring/README edits only.

:+1: (though it should be super easy to just move to explicit registration)

cretz (approved):

PR #54, opened 2023-12-18
TS docker fixes

Sushisource:
Make sure protos get built for the docker file


dockerfiles/typescript.Dockerfile
```diff
@@ -22,6 +22,8 @@ COPY go.mod go.sum ./
 
 # Build the CLI
 RUN CGO_ENABLED=0 /usr/local/go/bin/go build -o temporal-omes ./cmd
+# Build typescript proto files
+RUN cd workers/typescript && npm install && npm run proto-gen
```
Sushisource, line 26: We could also instead do this in the workflow rather than the docker image (which is why it works in the per-ci build)

dandavison (approved):
LGTM!

PR #58, opened 2024-01-11
Use https URL for submodule

dandavison:
This is convenient when building a docker image. Specifically, this PR helps Omes CICD work https://github.com/temporalio/saas-cicd/pull/558.


cretz (approved):
Completely agree that HTTP for submodules is way easier to use

PR #59, opened 2024-01-11
Fix state_transitions_steady workflow input

MichaelSnowden:
## What was changed

Use the raw WorkflowInput instead of TestInput wrapper for the params to the sdk client when executing the state transistion steady state scenario.

## Why?

This wrapper is unecessary and was causing the scenario to fail.

## Checklist

1. Closes 

2. How was this tested:

```
go run ./cmd run-scenario-with-worker --language go --scenario state_transitions_steady --run-id local-test-run-4 --duration 10s --option state-transitions-per-second=10
```

3. Any docs updates needed?


dandavison (approved):
LGTM!

PR #65, opened 2024-01-19
Add rate.Limiter to RunConfiguration

aromanovich:
## What was changed
`RunConfiguration` now accepts `*rate.Limiter`.

## Why?
To allow spreading out the load surge caused by going from 0 to `MaxConcurrent` runs in the beginning of the benchmark.

## Checklist

1. No issues to close.

2. How was this tested:
I ran benchmarks.

3. Any docs updates needed?
No.

### DIFF
--- a/loadgen/generic_executor.go
+++ b/loadgen/generic_executor.go
@@ -7,6 +7,7 @@ import (
 
 	"go.temporal.io/sdk/client"
 	"go.uber.org/zap"
+	"golang.org/x/time/rate"
 )
 
 type GenericExecutor struct {
@@ -25,6 +26,7 @@ type genericRun struct {
 	info     ScenarioInfo
 	config   RunConfiguration
 	logger   *zap.SugaredLogger
+	limiter  *rate.Limiter
 	// Timer capturing E2E execution of each scenario run iteration.
 	executeTimer client.MetricsTimer
 }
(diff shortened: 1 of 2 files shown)


loadgen/generic_executor.go
```diff
@@ -25,6 +26,7 @@ type genericRun struct {
 	info     ScenarioInfo
 	config   RunConfiguration
 	logger   *zap.SugaredLogger
+	limiter  *rate.Limiter
```
Sushisource, line 29: Why is this repeated here when it's stored on `config`? In fact, it seems this field is read but never written - I'm not sure anything's actually happening yet.

aromanovich: 🤦 You're totally right. This is a leftover from my initial draft hardcoded implementation.

I've fixed it, now the construction like this works as expected:
```golang
loadgen.MustRegisterScenario(loadgen.Scenario{
	Description: "Each iteration executes a single workflow with a noop activity.",
	Executor: loadgen.KitchenSinkExecutor{
		DefaultConfiguration: loadgen.RunConfiguration{
			Limiter: rate.NewLimiter(rate.Every(time.Second * 5), 1),
		},
		TestInput: &kitchensink.TestInput{
			WorkflowInput: &kitchensink.WorkflowInput{
				InitialActions: []*kitchensink.ActionSet{
					kitchensink.NoOpSingleActivityActionSet(),
				},
			},
		},
	},
})
```

cretz (approved):

loadgen/generic_executor.go
```diff
@@ -54,6 +54,9 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
 	if run.config.MaxConcurrent == 0 {
 		run.config.MaxConcurrent = g.DefaultConfiguration.MaxConcurrent
 	}
+	if run.config.Limiter == nil {
+		run.config.Limiter = g.DefaultConfiguration.Limiter
+	}
```
cretz, line 59:
```suggestion
```

Seems a nil limiter is supported at runtime

cretz (changes requested):
(sorry, removing accidental approval from earlier)

Issue #63, opened 2024-01-17
[Bug] Scenario.Run should return success iff the run succeeded for appropriate definition of success


dandavison:
We've noticed that `Scenario.Run` returns a non-error result in circumstances when it should clearly report an error: for example no worker was running and none of the workflows ever completed.

Define success appropriately (perhaps with CLI arguments influencing the definition) and ensure that   the return value honors the definition.

bergundy:
This only happens when you run a scenario with a duration parameter, not with an iteration count.
`GenericExecutor` doesn't not count run errors if they're emitted after the scenario deadline is exceeded.
Some ideas:
1. Put an optional deadline the context for each iteration - this works if the iteration deadline is shorter than the scenario deadline
2. Add a scenario option to require at least X successes

PR #66, opened 2024-01-23
Wait for completion of runs when using duration flag and add overall timeout

Sushisource:
## What was changed
Previously once the duration flag time was hit, the program would simply exit regardless of what actually happened. Now, we'll wait for any in-progress executions to finish. Since they may not, an overall timeout flag has been added which works with both iteration and duration mode

## Why?
Needed to make CI return pass/fail reliably

## Checklist

1. Closes https://github.com/temporalio/omes/issues/63

2. How was this tested:
Manually and unit tests

3. Any docs updates needed?


loadgen/generic_executor.go
```diff
@@ -106,24 +127,30 @@ func (g *genericRun) Run(ctx context.Context) error {
...
+		return fmt.Errorf("no runs were started. Iterations should be >0 or duration should be >0")
+	}
+	// Wait for all to be done or an error to occur. We will wait past the overall duration for
+	// executions to complete. It is expected that whatever is running omes may choose to enforce
+	// a hard timeout if waiting for started executions to complete exceeds a certain threshold.
+	for runErr == nil && currentlyRunning > 0 && ctx.Err() == nil {
+		waitOne(false)
+	}
+	if ctx.Err() != nil {
+		return fmt.Errorf("timed out while waiting for runs to complete: %w", ctx.Err())
```
dandavison, line 153: nit: move this termination into the loop? (Technically there's a tiny race as it is)

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -18,6 +18,10 @@ type KSWorkflowState struct {
 
 func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (*common.Payload, error) {
 	workflow.GetLogger(ctx).Debug("Started kitchen sink workflow")
+	err := workflow.Sleep(ctx, 10*time.Second)
+	if err != nil {
+		return nil, err
+	}
```
dandavison, line 24: What's the deal with this?

loadgen/generic_executor.go
```diff
@@ -106,24 +127,30 @@ func (g *genericRun) Run(ctx context.Context) error {
...
+				g.executeTimer.Record(time.Since(startTime))
 			}
 		}()
 	}
-	// Wait for all to be done or an error to occur
-	for runErr == nil && ctx.Err() == nil && currentlyRunning > 0 {
-		waitOne()
+	g.logger.Debugf("No longer starting new iterations")
+	if currentlyRunning == 0 {
+		return fmt.Errorf("no runs were started. Iterations should be >0 or duration should be >0")
```
dandavison, line 144: > "no runs were started"

`currentlyRunning` could also be zero at this point due to an error setting `runErr`, right?

loadgen/generic_executor.go
```diff
@@ -106,24 +127,30 @@ func (g *genericRun) Run(ctx context.Context) error {
...
-					// Record/log here, not if it was cut short by context complete
-					g.executeTimer.Record(time.Since(startTime))
-				}
+			if err != nil {
+				err = fmt.Errorf("iteration %v failed: %w", run.Iteration, err)
+				g.logger.Error(err)
+			}
+			select {
+			case <-ctx.Done():
+			case doneCh <- err:
```
dandavison, line 136: I don't see where anything writes to `doneCh`.

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
-		ctx, cancel = context.WithTimeout(ctx, g.config.Duration)
+	if g.config.Timeout > 0 {
+		g.logger.Debugf("Will timeout after %v", g.config.Timeout)
+		ctx, cancel = context.WithTimeout(ctx, g.config.Timeout)
 	}
 	defer cancel()
 
+	durationCtx := ctx
+	if g.config.Duration > 0 {
+		durationCtx, cancel = context.WithTimeout(ctx, g.config.Duration)
```
cretz, line 82: Reassigning cancel here is technically fine because it's not used later besides the next line, but it's not advisable

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
+		defer cancel()
+	}
+
 	startTime := time.Now()
 	var runErr error
 	doneCh := make(chan error)
 	var currentlyRunning int
-	waitOne := func() {
+	waitOne := func(exitOnDurationDone bool) {
+		timeoutOrPending := make(<-chan struct{})
```
cretz, line 91: Just an FYI, this also works I believe (no need to change anything)

```suggestion
		var timeoutOrPending <-chan struct{}
```

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
-	// Run all until we've gotten an error or reached iteration limit
-	for i := 0; runErr == nil && ctx.Err() == nil &&
+	if g.config.Iterations > 0 {
+		g.logger.Debugf("Will run %v iterations", g.config.Iterations)
+	} else {
+		g.logger.Debugf("Will start iterations for %v", g.config.Duration)
+	}
+
+	// Run all until we've gotten an error or reached iteration limit or timeout
+	for i := 0; runErr == nil && durationCtx.Err() == nil &&
```
cretz, line 113: > or timeout

If the regular context completes (now backed by timeout), this loop will not exit as may be expected. Do you want to set `runErr` on `<-ctx.Done()` above or add `&& ctx.Err() == nil` here?

workers/go/kitchensink/kitchen_sink.go
Sushisource: Ah! I forgot to remove that. Thank you.

loadgen/generic_executor.go
Sushisource: Hmm, yeah, probably should. The inner stuff it's starting would time out right away but good to have it there too.

loadgen/generic_executor.go
Sushisource: Yeah, good point.

cretz (approved):

bergundy:
I'm sorry I missed this PR. I have some concerns about the approach but didn't have time to review.
I hope to get to it this evening or early tomorrow morning.

loadgen/generic_executor_test.go
```diff
@@ -128,3 +128,36 @@ func TestRunFailDuration(t *testing.T) {
...
+		DefaultConfiguration: RunConfiguration{
+			Duration: 100 * time.Millisecond,
+			Timeout:  10 * time.Millisecond},
+	})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "timed out")
+}
+
+func TestRunIterationsWithTimeout(t *testing.T) {
+	tracker := newIterationTracker()
```
bergundy, line 149: Doesn't look like this tracker is asserting anything here of in the test above. You can remove it.

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
+		defer cancel()
+	}
+
 	startTime := time.Now()
 	var runErr error
 	doneCh := make(chan error)
 	var currentlyRunning int
-	waitOne := func() {
+	waitOne := func(respectDuration bool) {
+		durationDone := make(<-chan struct{})
```
bergundy, line 91: nit: no need to create a channel for every iteration. You could pass the duration context here instead.

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
-	// Run all until we've gotten an error or reached iteration limit
-	for i := 0; runErr == nil && ctx.Err() == nil &&
+	if g.config.Iterations > 0 {
+		g.logger.Debugf("Will run %v iterations", g.config.Iterations)
+	} else {
+		g.logger.Debugf("Will start iterations for %v", g.config.Duration)
+	}
+
+	// Run all until we've gotten an error or reached iteration limit or timeout
+	for i := 0; runErr == nil && durationCtx.Err() == nil && ctx.Err() == nil &&
```
bergundy, line 113: durationCtx is either equal to ctx or a child of ctx, no need to check both.

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
 		(g.config.Iterations == 0 || i < g.config.Iterations); i++ {
 		// If there are already MaxConcurrent running, wait for one
 		if currentlyRunning >= g.config.MaxConcurrent {
-			waitOne()
+			waitOne(true)
 			// Exit loop if error
```
bergundy, line 118: Update the comment:

```
// Exit loop on first error or scenario exceeds given duration.
```

loadgen/generic_executor.go
```diff
@@ -106,24 +127,27 @@ func (g *genericRun) Run(ctx context.Context) error {
...
-				case doneCh <- err:
-					// Record/log here, not if it was cut short by context complete
-					g.executeTimer.Record(time.Since(startTime))
-				}
+			if err != nil {
+				err = fmt.Errorf("iteration %v failed: %w", run.Iteration, err)
+				g.logger.Error(err)
+			}
+			select {
+			case <-ctx.Done():
```
bergundy, line 135: Should check durationCtx instead or this can deadlock.

bergundy (commented):
Sorry for the late review.

Not sure I like the complexity this feature added to the generic executor and it's not the most intuitive to me (compared to other approaches I suggested in the liked issue) but it works and has value for both iteration and duration scenarios.

loadgen/generic_executor.go
```diff
@@ -68,34 +71,52 @@ func (g *GenericExecutor) newRun(info ScenarioInfo) (*genericRun, error) {
...
+			durationDone = durationCtx.Done()
+		}
 		select {
 		case err := <-doneCh:
 			currentlyRunning--
 			if err != nil {
 				runErr = err
 			}
 		case <-ctx.Done():
+		case <-durationDone:
```
bergundy, line 102: No need to select both, you can select either durationCtx.Done or ctx.Done.

I would just pass the context to `waitOne` instead.

loadgen/generic_executor.go
Sushisource: I don't think that's right - this needs to be like this. If we exited when durationCtx was done, always, then we would fail to properly wait on outstanding workflows at the end. I don't see how it could deadlock - it could take forever if the runs never complete (and no timeout is specified), but that's kinda the point.

loadgen/generic_executor.go
Sushisource: Ah, yeah, good point. I blindly applied Chad's suggestion.

PR #67, opened 2024-01-24
Add more TLS options for Go.

robholland:
## What was changed
Adds --tls-cert-path and --tls-skip-verify to Go runner/worker.

## Why?

This is required to test against servers that need custom CA certificates, such as Temporal Cloud.

Also adds the skip verify option for servers whose certificates cannot be validated (self signed or similar).

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/cmdoptions/client.go
+++ b/cmd/cmdoptions/client.go
@@ -3,6 +3,7 @@ package cmdoptions
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -36,6 +37,10 @@ type ClientOptions struct {
 	Namespace string
 	// Enable TLS
 	EnableTLS bool
+	// Skip host verification
+	SkipVerify bool
+	// TLS CA cert
+	CACertPath string
 	// TLS client cert
 	ClientCertPath string
 	// TLS client private key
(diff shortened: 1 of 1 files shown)


cmd/cmdoptions/client.go
```diff
@@ -36,6 +37,10 @@ type ClientOptions struct {
 	Namespace string
 	// Enable TLS
 	EnableTLS bool
+	// Skip host verification
+	SkipVerify bool
```
cretz, line 41: Just a bit of qualification to ensure it's not confused with general options:

```suggestion
	// Skip host verification
	TLSSkipVerify bool
```

But are you sure you need this? Or would the following make more sense:

```suggestion
	// SNI override
	TLSServerName string
```

Not that it matters much

cmd/cmdoptions/client.go
```diff
@@ -46,6 +51,10 @@ type ClientOptions struct {
 
 // loadTLSConfig inits a TLS config from the provided cert and key files.
 func (c *ClientOptions) loadTLSConfig() (*tls.Config, error) {
```
cretz, line 53: Before this function would return a non-nil config if `EnableTLS` was false but one of the other TLS options was present like `ClientCertPath`. But now this will return `nil` even when one of those values is set.

Sushisource:
We'll want to support these in the other languages too to have parity, but it doesn't have to be done in this PR. Let's open an item for that if they won't go in this PR.

robholland:
Closing this PR. After testing to see if ServerName would fix the issues instead of SkipVerify I discovered that I should not be setting the CA cert at all. This was the reason my verification was failing as it prevented the default CA bundle from being used to verify the server's certificate. The CA used to sign the client certificates is not related to the certificate/CA the server uses.

PR #75, opened 2024-01-31
Allow adding a Search Attribute to fail, but warn.

robholland:
## What was changed
Warn but don't fail if a Search Attribute creation request fails.

Make cleanup-scenario batch job IDs unique.

## Why?

This allows running Omes against Cloud which does not permit search attributes to be added via the WorkflowService API.

Unique batch job IDs for cleanup allows it to be used multiples times on the same namespace.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cmd/cleanup_scenario.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os/user"
 	"time"
 
+	"github.com/pborman/uuid"
```
bergundy, line 10: Use the more standard https://github.com/google/uuid please.

scenarios/throughput_stress.go
```diff
@@ -34,13 +34,18 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
 			SearchAttributes: attribMap,
 		})
-	var svcErr *serviceerror.AlreadyExists
-	if !errors.As(err, &svcErr) {
+	var deniedErr *serviceerror.PermissionDenied
+	var alreadyErr *serviceerror.AlreadyExists
+
+	if errors.As(err, &deniedErr) {
+		info.Logger.Warnf("Failed to add Search Attribute: %v", err)
+	} else if !errors.As(err, &alreadyErr) {
```
bergundy, line 48: Wanna log info here and on the successful case?

cretz (approved):

PR #87, opened 2024-04-26
Add more TLS options 

yux0:
## What was changed
Add more TLS options 

## Why?
Add the tls options to support Private Link testing

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

bergundy:
We need this in all SDKs not just Go. Doesn't have to be in this PR but should at least be tracked in an issue.

Issue #93, opened 2024-07-16
[Feature Request] Support env setup step


bergundy:
Each scenario should be able to define the environment it expects to run in (what namespaces, search attributes, etc..) it depends on.

The exact way to express this is TBD. This functionality should work on both self-hosted servers and cloud cells.

PR #94, opened 2024-07-16
Remove search attribute registration before starting every workflow

bergundy:
Omes is supposed to replicate user behavior to generate load on the server. Registering search attributes doesn't belong in the hot path, it should be done as a separate setup step.

See also: https://github.com/temporalio/omes/issues/93

### DIFF
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -10,8 +10,6 @@ import (
 	"time"
 
 	"github.com/temporalio/omes/loadgen/kitchensink"
-	"go.temporal.io/api/enums/v1"
-	"go.temporal.io/api/operatorservice/v1"
 	"go.temporal.io/sdk/client"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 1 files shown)


loadgen/scenario.go
```diff
@@ -199,33 +197,6 @@ func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSi
...
-			if strings.Contains(err.Error(), s) {
-				isAlreadyExistsErr = true
-				break
-			}
-		}
-		if !isAlreadyExistsErr {
-			return fmt.Errorf("failed to register search attributes: %w", err)
-		}
-	}
-
```
dandavison, line 228: I don't have the necessary context here. This was added by @Sushisource in https://github.com/temporalio/omes/pull/41. Is it not needed for anything?

dandavison: Discussed internally. The plan is for clients of omes such as cicd to do this.

dandavison (approved):

loadgen/scenario.go
Sushisource: We still need some way for this to happen automatically so that when developers come along and want to just run omes than they don't have to interpret why things are failing.

This should be automatically done _somewhere_ in this codebase still. It doesn't have to be here, but I don't want to merge this PR without doing it somewhere else, or, at the bare minimum, documenting the necessity

bergundy: That works for me.

Sushisource: wait, but, you've just merged it without those things :sweat_smile:

cretz: Yeah, I was hoping that at least the scenario requirements would be documented (though would prefer a bootstrapping command)

PR #96, opened 2024-07-23
Update SDKs & show path to running update tests

Sushisource:
Updated SDKs & added some stuff to readmes and the fuzzer scenario to better explain how to run it and how to do repeatable runs.

Also added some seeds I encountered issues on while testing.

It was very, very easy to encounter issues (often update related, sometimes not). So I don't know that you really need to use the saved seeds. This was using the CLI as of https://github.com/temporalio/cli/commit/29b1eb17d29cf968df0c9ce89c4627901d869ce3


dandavison (approved):
Great, LGTM

dandavison:
Actually, after approving I had a couple of thoughts / findings:

- When running with `--embedded-server` I'm getting
  ```
  $ go run ./cmd run-scenario-with-worker --scenario workflow_with_single_noop_activity --language go --embedded-server
  compile: version "go1.22.5" does not match go tool version "go1.21.2"
  ```
- Since users will often want to test a specific server version, perhaps remove `--embedded-server` from README examples, so that people know it expects to connect to a running server?
- Is it possible to collect information on the server version used in the yaml file?

dandavison:
Also

```
go run ./cmd run-scenario-with-worker --embedded-server --scenario fuzzer --iterations 1 --language cs --option seed=131962944538087455
```

```
UpdateWorkflowExecution operation is disabled on this namespace.
```

PR #99, opened 2024-07-31
Allow building worker images from local SDK paths

Sushisource:
Allows the image building workflow to use a reference to an SDK repo to build an image with it as of that commit

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -61,14 +61,6 @@ jobs:
           ref: ${{ inputs.sdk-repo-ref }}
           fetch-depth: 0
 
-      - name: Set SDK Version Variable
-        run: |
-          if [ -z "${{ inputs.sdk-version }}" ]; then
-              echo "SDK_VERSION=checked-out-sdk/" >> $GITHUB_ENV
-          else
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-images.yml
```diff
@@ -36,11 +41,33 @@ jobs:
...
+          ref: ${{ inputs.sdk-repo-ref }}
+          fetch-depth: 0
+
+      - name: Set SDK Version Variable
+        run: |
+          if [ -z "${{ inputs.sdk-version }}" ]; then
+              echo "SDK_VERSION=checked-out-sdk/" >> $GITHUB_ENV
+          else
+              echo "SDK_VERSION=${{ inputs.sdk-version }}" >> $GITHUB_ENV
+          fi
```
dandavison, line 70: Up to you but this might be less surprising to read if we did the equivalent of

```
--version ${{ env.SDK_VERSION || checked-out-sdk/ }}
```

below (or whatever the correct syntax is for doing that) and didn't set the env var to the strange value.

dandavison (approved):
One suggestion but LGTM

.github/workflows/docker-images.yml
Sushisource: Sure, done. Can't remember why I thought I needed this lol.

PR #111, opened 2024-08-15
Add Signal-with-Start

stephanos:
## What was changed

Add Signal-with-Start to fuzzer.

## Why?

First part of adding Signal-with-Start and soon Update-with-Start.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -65,6 +66,7 @@ message DoSignal {
     // Send an arbitrary signal
     HandlerInvocation custom = 2;
   }
+  bool with_start = 3;
```
stephanos, line 69: I also considered adding it to `ClientAction` instead. How are we dealing with breaking changes; ie could I move it later or is there a backwards-compatibility guarantee?

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -15,6 +15,7 @@ import "temporal/api/enums/v1/workflow.proto";
 message TestInput {
   WorkflowInput workflow_input = 1;
   ClientSequence client_sequence = 2;
+  ClientAction with_start_action = 3;
```
stephanos, line 18: I had considered using the client sequence's first item, but that seemed too surprising and messy (ie how do you represent "just start").

.github/workflows/ci.yml
```diff
@@ -183,7 +183,16 @@ jobs:
...
+          git config --global core.safecrlf false
+          git diff > generator.diff
+          git diff --exit-code
+      - name: Upload generator diff
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: generator-diff
+          path: generator.diff
+          if-no-files-found: ignore
```
stephanos, line 195: This was a suggestion I added to get the protoc results from the CI, download them and apply them locally. Managing to get the right version locally, at least on OSX, is a nightmare.

Sushisource: Could add it to the README somewhere

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: Breaking changes aren't a problem here

workers/proto/kitchen_sink/kitchen_sink.proto
Sushisource: I prefer it here as well

Sushisource, line 18: I think this should be a new type though. Query doesn't make sense as a with start action, for example.

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -302,10 +302,21 @@ impl<'a> Arbitrary<'a> for TestInput {
...
         let mut client_sequence: ClientSequence = u.arbitrary()?;
+
+        // Sometimes we want a with_start_action
+        let with_start_action = if u.ratio(80, 100)? {
+            None
+        } else {
+            let mut signal_action: DoSignal = u.arbitrary()?;
+            signal_action.with_start = true;
+            Some(ClientAction { variant: Some(client_action::Variant::DoSignal(signal_action)) })
+        };
```
Sushisource, line 313: If we apply my other suggestion of using a specific `WithStartClientAction` type, then this can become the `Arbitrary` implementation for that, which fits nicely.

Sushisource (commented):
Looking pretty reasonable to me - just suggest one change to break out a new type.

Sushisource (approved):
Noice.

Any cool outcomes from running it?

stephanos:
@Sushisource so far I've only run it for a few seconds to make sure my change was working

PR #115, opened 2024-09-12
Update .NET SDK to 1.3.1

cretz:
## What was changed

* Update .NET SDK 1.3.1
* Added https://github.com/temporalio/sdk-core/issues/803 to trophy case


.github/workflows/ci.yml
```diff
@@ -205,4 +205,4 @@ jobs:
       ts-version: 'v1.10.2'
       java-version: 'v1.24.2'
```
yuandrew, line 206: Are these versions supposed to be in sync with the versions earlier in the file?

yuandrew (approved):

.github/workflows/ci.yml
cretz: I am unsure, but they look in sync to me

PR #116, opened 2024-09-19
Use deterministic version of asyncio.wait

dandavison:
https://python.temporal.io/temporalio.workflow.html#wait

### DIFF
--- a/workers/python/kitchen_sink.py
+++ b/workers/python/kitchen_sink.py
@@ -98,11 +98,12 @@ class KitchenSinkWorkflow:
         should_return_task = asyncio.create_task(
             workflow.wait_condition(lambda: return_value is not None)
         )
+        # mypy cannot handle the heterogeneous arguments to `wait`
         done, _ = await workflow.wait(
             [gather_fut, should_return_task], return_when=asyncio.FIRST_COMPLETED
         )  # type: ignore
         for fut in done:
-            await fut
+            await fut  # type: ignore
         return return_value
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

workers/python/kitchen_sink.py
```diff
@@ -98,7 +98,7 @@ async def run_action(action: Action) -> None:
         should_return_task = asyncio.create_task(
             workflow.wait_condition(lambda: return_value is not None)
         )
-        done, _ = await asyncio.wait(
+        done, _ = await workflow.wait(
             [gather_fut, should_return_task], return_when=asyncio.FIRST_COMPLETED
         )  # type: ignore
         for fut in done:
```
dandavison, line 104: Annoyingly, `mypy` doesn't infer that `done` is awaitable here. Not sure why that's not causing lint errors in the SDK (e.g. tests).

dandavison:
```
kitchen_sink.py:105: error: Incompatible types in "await" (actual type "object", expected type "Awaitable[Any]")
```

dandavison: Investigated a bit. The problem is the heterogenous arguments (an `asyncio.Future` and an `asyncio.Task`). `mypy` cannot handle this, whether for `asyncio.wait` or for our version of `wait`, but the wrong inference it makes is slightly different between the two.

PR #117, opened 2024-10-16
Add build-id option to Go worker.

robholland:
## What was changed
Added an option to pass a build ID to the worker.

## Why?
This allows testing of build-id versioning using omes.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


workers/go/worker/worker.go
```diff
@@ -61,6 +61,8 @@ func runWorkers(client client.Client, taskQueues []string, options cmdoptions.Wo
 		taskQueue := taskQueue
 		go func() {
 			w := worker.New(client, taskQueue, worker.Options{
+				BuildID:                                options.BuildID,
+				UseBuildIDForVersioning:                options.BuildID != "",
```
robholland, line 65: This feels safe, but maybe this needs to be an explicit option?

robholland:
cc @Prajithp

cretz (approved):

PR #120, opened 2024-11-04
Update gradle to v8.10.2

Quinn-With-Two-Ns:
Update gradle to v8.10.2


Quinn-With-Two-Ns:
Updating to fix https://github.com/temporalio/sdk-java/actions/runs/11668450477/job/32488133845 this action

cretz (approved):

PR #126, opened 2025-01-17
Update SDKs

dandavison:


cretz:
Note there is a branch at `features` at https://github.com/temporalio/features/tree/python-version-upgrade that fixes the Python `--no-dev` thing. May be able to merge that now, quick glance at https://github.com/temporalio/features/compare/main...python-version-upgrade I don't shows any incompatibilities

dandavison:
> Note there is a branch at features at https://github.com/temporalio/features/tree/python-version-upgrade that fixes the Python `--no-dev` thing. 

The features Python `--no-dev` fix is also in the `Fix CI` features PR https://github.com/temporalio/features/pull/577. Unfortunately nothing in features can merge until we fix the cloud tests in the PHP and Go docker builds there, so looks like this is another thing blocked on that.

PR #132, opened 2025-03-27
RPS limit

stephanos:
## What was changed

Add RPS parameter for executor.

## Why?

To generate continuous but throttled load.

The new parameter allows to create a scenario where a large number of iterations are started one-by-one at a constant rate (e.g. 5/s).

This is useful for creating a backlog of workflows (by turning off the workers but keep creating new workflows).

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cmd/run_scenario.go
```diff
@@ -52,6 +52,7 @@ type scenarioRunConfig struct {
 	iterations                    int
 	duration                      time.Duration
 	maxConcurrent                 int
+	ratePerSecond                 float64
```
Sushisource, line 55: maxIterationsPerSecond might be a better name

Sushisource (approved):

PR #136, opened 2025-04-12
Activity Priorities

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


scenarios/throughput_stress.go
```diff
@@ -38,12 +41,33 @@ type tpsExecutor struct {
...
+	}
+	sleepActivityDurationDistribution, err := loadgen.ParseDurationDistribution(info.ScenarioOptions[SleepActivityDurationDistributionFlag])
+	if err != nil {
+		return fmt.Errorf("failed to parse %v: %w", SleepActivityDurationDistributionFlag, err)
+	}
+	visibilityVerificationTimeout, err := time.ParseDuration(cmp.Or(info.ScenarioOptions[VisibilityVerificationTimeout], "3m"))
+	if err != nil {
+		return fmt.Errorf("failed to parse %v: %w", VisibilityVerificationTimeout, err)
+	}
+	timeout := time.Duration(1*internalIterations) * internalIterTimeout
```
stephanos, line 63: Moved all of these to the top so (a) if there's an error, it aborts immediately and (b) so it's not re-parsed on every run.

PR #138, opened 2025-04-21
Fix python worker build

dandavison:
- Migrate to `uv` in order to work with latest SDK
- Specify SDK version when running scenario-with-worker in CI


cmd/prepare_worker.go
```diff
@@ -120,31 +120,13 @@ func main() {
...
-		for _, line := range strings.Split(string(b), "\n") {
-			line = strings.TrimSpace(line)
-			if strings.HasPrefix(line, "temporalio = ") {
-				version = line[strings.Index(line, `"`)+1 : strings.LastIndex(line, `"`)]
-				break
-			}
-		}
-		if version == "" {
-			return nil, fmt.Errorf("version not found in pyproject.toml")
-		}
```
dandavison, line 139: This was a hack and I would like it not to be necessary.

cretz: Don't all langs support `omes` and `features` runs without specifying an SDK version?

dandavison: OK, I think that to get the "not specifying an SDK version" functionality we should modify `sdkbuild` to allow the version to be missing, and delegate to the package resolvers.

However for now I've got rid of the hacky TOML parsing here and replaced it with a call to `uv tree`.

Sushisource (approved):

README.md
```diff
@@ -15,8 +15,7 @@ Omes (pronounced oh-mess) is the Hebrew word for "load" (עומס).
     available for download that you can use with `git apply`
 - [Java](https://openjdk.org/) 8+
 - TypeScript: [Node](https://nodejs.org) 16+
-- [Python](https://www.python.org/) 3.10+
-  - [Poetry](https://python-poetry.org/): `poetry install`
+- [uv](https://docs.astral.sh/uv/)
```
cretz, line 18: Do they no longer need Python installed? So `uv` installs it lazy? If so, can make it clear `uv` is only for the purposes of Python like TypeScript does with Node:

```suggestion
- Python: [uv](https://docs.astral.sh/uv/)
```

cmd/prepare_worker.go
cretz: Works for me, though I think `features` repo will need the same treatment at some point

cretz (approved):

README.md
dandavison: Right, `uv` installs pythons and manages which is active.

PR #143, opened 2025-04-29
Make throughputStress resumable

stephanos:
## What was changed

Added affordances to make the throughputStress scenario resumable.

## Why?

If the throughputStress scenario crashes (e.g. pod restarted), we want to be able to continue where it left off.

## Test

Started Omes with

```
--run-id omes \
--language go \
--iterations 2 \
--max-concurrent 1 \
--option internal-iteratios-timeout=1s \
--option visibility-count-timeout=5s
```

Once the first iteration succeeded, I stopped it and then ran (numbers taken from the logged status update):

```
--run-id omes \
--language go \
--iterations 2 \
--option internal-iteratios-timeout=1s \
--option visibility-count-timeout=5s \
--option internal-iterations-resume-from=1 \
--option workflow-count-resume-from=15
```

And it completed successfully.

### DIFF
--- a/loadgen/helpers.go
+++ b/loadgen/helpers.go
@@ -502,6 +502,6 @@ func VisibilityCountIsEventually(
 			return fmt.Errorf("expected %d workflows in visibility, got %d after waiting %v",
 				expectedCount, visibilityCount.Count, waitAtMost)
 		}
-		time.Sleep(5 * time.Second)
+		time.Sleep(1 * time.Second)
 	}
 }
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -95,6 +95,8 @@ type ScenarioInfo struct {
 	Namespace string
 	// Path to the root of the omes dir
 	RootPath string
+	// Optional callback to receive status updates from the scenario.
+	StatusCallback func(any)
 }
 
 func (s *ScenarioInfo) ScenarioOptionInt(name string, defaultValue int) int {
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -95,7 +95,8 @@ type ScenarioInfo struct {
 	Namespace string
 	// Path to the root of the omes dir
 	RootPath string
-	// Optional callback to receive status updates from the scenario.
+	// Optional callback to receive status updates from the scenario. If, when and with what
+	// data (type) it is invoked are specific to the scenario. See `throughputStress` for an example.
 	StatusCallback func(any)
 }
 
--- a/scenarios/throughput_stress.go
+++ b/scenarios/throughput_stress.go
@@ -22,16 +22,16 @@ import (
 const (
 	IterFlag    = "internal-iterations"
 	IterTimeout = "internal-iterations-timeout"
-	// IterResumeFrom is the iteration from which to resume a previous run from. If set, it will skip the
+	// IterResumeAt is the iteration from which to resume a previous run from. If set, it will skip the
 	// run initialization and start the next iteration starting from the given iteration.
-	IterResumeFrom    = "internal-iterations-resume-from"
+	IterResumeAt      = "internal-iterations-resume-at"
 	SkipSleepFlag     = "skip-sleep"
 	CANEventFlag      = "continue-as-new-after-event-count"
 	NexusEndpointFlag = "nexus-endpoint"
 	// WorkflowIDPrefix is the prefix for each run's workflow ID. Use it to ensure that the workflow IDs are unique.
 	WorkflowIDPrefix = "workflow-id-prefix"
-	// WorkflowCountResumeFrom is the number of completed workflows from a previous run that is being resumed.
-	WorkflowCountResumeFrom = "workflow-count-resume-from"
+	// WorkflowCountResumeAt is the number of completed workflows from a previous run that is being resumed.
+	WorkflowCountResumeAt = "workflow-count-resume-at"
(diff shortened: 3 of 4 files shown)


loadgen/helpers.go
```diff
@@ -502,6 +502,6 @@ func VisibilityCountIsEventually(
 			return fmt.Errorf("expected %d workflows in visibility, got %d after waiting %v",
 				expectedCount, visibilityCount.Count, waitAtMost)
 		}
-		time.Sleep(5 * time.Second)
+		time.Sleep(1 * time.Second)
```
stephanos, line 505: 1s seems like a good compromise between too slow and too fast.

workers/go/throughputstress/workflow.go
```diff
@@ -104,6 +104,7 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
 				// run to the child.
 				attrs := workflow.GetInfo(ctx).SearchAttributes.IndexedFields
 				childCtx := workflow.WithChildOptions(ctx, workflow.ChildWorkflowOptions{
+					WorkflowID: fmt.Sprintf("%s/child-%d", workflow.GetInfo(ctx).WorkflowExecution.ID, output.ChildrenSpawned),
```
stephanos, line 107: Helps correlate the child workflows with the parent.

scenarios/throughput_stress.go
```diff
@@ -69,58 +94,34 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
+	// Start the scenario run.
 	genericExec := &loadgen.GenericExecutor{
 		DefaultConfiguration: loadgen.RunConfiguration{
 			Iterations:    20,
 			MaxConcurrent: 5,
 		},
 		Execute: func(ctx context.Context, run *loadgen.Run) error {
-			wfID := fmt.Sprintf("%s-%s-%d", workflowIDPrefix, run.RunID, run.Iteration)
+			curIteration := internalIterResumeFrom + run.Iteration
+			wfID := fmt.Sprintf("%s/%s/iter-%d", workflowIDPrefix, run.RunID, curIteration)
```
stephanos, line 116: Made the workflow ID a bit more expressive.

scenarios/throughput_stress.go
```diff
@@ -134,8 +135,17 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 					NexusEndpoint:                nexusEndpoint,
 					SleepActivityPerPriority:     sleepActivityPerPriority,
 				})
-			// The 1 is for the final workflow run
-			t.workflowCount.Add(uint64(result.TimesContinued + result.ChildrenSpawned + 1))
```
stephanos, line 138: Noticed when testing that this is actually wrong as it ignores the error.

scenarios/throughput_stress.go
```diff
@@ -41,18 +45,42 @@ const (
...
 	internalIterTimeout := info.ScenarioOptionDuration(IterTimeout, time.Minute)
+	internalIterResumeFrom := info.ScenarioOptionInt(IterResumeFrom, 0)
+	_, resumingFromPreviousRun := info.ScenarioOptions[IterResumeFrom]
 	continueAsNewCount := info.ScenarioOptionInt(CANEventFlag, 120)
 	workflowIDPrefix := cmp.Or(info.ScenarioOptions[WorkflowIDPrefix], defaultWorkflowIDPrefix)
+	workflowCountResume := info.ScenarioOptionInt(WorkflowCountResumeFrom, 0)
 	nexusEndpoint := info.ScenarioOptions[NexusEndpointFlag] // disabled by default
 	skipSleep := info.ScenarioOptionBool(SkipSleepFlag, false)
+	if info.StatusCallback == nil {
+		info.StatusCallback = func(data any) {}
```
stephanos, line 82: I considered doing this further up the callstack, but didn't find a good place to set the default.

stephanos: I could move it into `RunConfiguration` (which has a "set defaults" step)?

scenarios/throughput_stress.go
```diff
@@ -134,8 +138,17 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
 				})
-			// The 1 is for the final workflow run
-			t.workflowCount.Add(uint64(result.TimesContinued + result.ChildrenSpawned + 1))
+
+			if err == nil {
+				// The 1 is for the final workflow run.
+				curTotal := t.workflowCount.Add(uint64(result.TimesContinued + result.ChildrenSpawned + 1))
+
+				info.StatusCallback(tpsStatusUpdate{
+					CompletedIteration: curIteration,
```
stephanos, line 147: The caller will need to make sure to store the "low water mark"; and not just the latest iteration.

loadgen/scenario.go
```diff
@@ -95,6 +95,8 @@ type ScenarioInfo struct {
 	Namespace string
 	// Path to the root of the omes dir
 	RootPath string
+	// Optional callback to receive status updates from the scenario.
+	StatusCallback func(any)
```
Sushisource, line 99: A little more detail in this comment would be good. Explain that the callback is called with info specific to the scenario, and the person defining the callback needs to know what that type is, etc

Sushisource (approved):
I think this makes good sense - I'd like to see how the callback gets used but I don't have any objections here.

PR #144, opened 2025-04-30
Make throughputStress resumable (part 2)

stephanos:
## What was changed

Some tweaks to the previous PR https://github.com/temporalio/omes/pull/143

I decided to make the scenario take care of tracking the state exactly after all, as it allows the caller to be oblivious to its inner details.

## Checklist

Using it like this

Screenshot 2025-04-30 at 6 33 05 PM

### DIFF
--- a/scenarios/throughput_stress.go
+++ b/scenarios/throughput_stress.go
@@ -6,7 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	"github.com/temporalio/omes/loadgen"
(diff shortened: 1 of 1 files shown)


scenarios/throughput_stress.go
```diff
@@ -22,16 +22,13 @@ import (
...
 	IterFlag    = "internal-iterations"
 	IterTimeout = "internal-iterations-timeout"
-	// IterResumeAt is the iteration from which to resume a previous run from. If set, it will skip the
-	// run initialization and start the next iteration starting from the given iteration.
-	IterResumeAt      = "internal-iterations-resume-at"
-	SkipSleepFlag     = "skip-sleep"
-	CANEventFlag      = "continue-as-new-after-event-count"
-	NexusEndpointFlag = "nexus-endpoint"
-	// WorkflowIDPrefix is the prefix for each run's workflow ID. Use it to ensure that the workflow IDs are unique.
-	WorkflowIDPrefix = "workflow-id-prefix"
```
stephanos, line 32: Removing this option as it was a stopgap solution to dealing with Omes failures.

scenarios/throughput_stress.go
```diff
@@ -140,13 +134,16 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
-				curTotal := t.workflowCount.Add(uint64(result.TimesContinued + result.ChildrenSpawned + 1))
-
-				info.StatusCallback(ThroughputStressScenarioStatusUpdate{
-					CompletedIteration: curIteration,
-					CompletedWorkflows: curTotal,
-				})
+				t.updateState(result, curIteration)
+
+				if info.StatusCallback != nil {
+					stateJson, err := json.Marshal(t.state)
```
stephanos, line 140: Since we need to unmarshall (scenario option values are strings), it seems reasonable to also marshal it here. That way the client can treat the state as an opaque string.

Sushisource (approved):
Cool! Yeah I like this more.

PR #153, opened 2025-06-17
Allow StartFromIteration with Duration

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


loadgen/generic_executor_test.go
```diff
@@ -52,6 +81,8 @@ func (i *iterationTracker) track(iteration int) {
 }
 
 func (i *iterationTracker) assertSeen(t *testing.T, iterations int) {
+	i.Lock()
+	defer i.Unlock()
```
stephanos, line 90: Noticed a data race here when I ran this locally.

loadgen/generic_executor_test.go
```diff
@@ -12,27 +12,56 @@ import (
...
+func TestScenarioConfigValidation(t *testing.T) {
+	tests := []struct {
+		name  string
+		info  ScenarioInfo
+		error string
+	}{
+		{
+			name:  "default",
+			info:  ScenarioInfo{Configuration: RunConfiguration{}},
+			error: "",
```
stephanos, line 24: Left the default value to signal to the reader that this is expected to be valid.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Build exe
         run: go build -o temporal-omes ./cmd
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
stephanos, line 40: Good practice.

loadgen/scenario.go
```diff
@@ -149,6 +146,10 @@ const DefaultMaxConcurrent = 10
 type RunConfiguration struct {
 	// Number of iterations to run of this scenario (mutually exclusive with Duration).
 	Iterations int
+	// StartFromIteration is the iteration to start from when resuming a run.
+	// This is used to skip iterations that have already been run.
+	// Default is zero. If Iterations is set, too, must be less than or equal to Iterations.
+	StartFromIteration int
```
stephanos, line 152: Moved StartFromIteration here since it seem it's really a configuration setting. (isn't it?)

PR #158, opened 2025-07-01
 Advanced sleep activity config

stephanos:
## What was changed

Redo of the sleep activity config.

## Why?

Allow deep customization of several parameters; some existing, some new.

We'll use these to test the Fairness feature.

## Checklist

Example from local run:
Screenshot 2025-07-02 at 4 10 12 PM

Using
```
{
  "count": {
    "type": "fixed",
    "value": 5
  },
  "groups": {
    "high": {
      "weight": 2,
      "sleepDuration": {
        "type": "uniform",
        "min": "2s",
        "max": "4s"
      },
      "priorityKeys": {
        "type": "uniform",
        "min": "3",
        "max": "4"
      },
      "fairnessKeys": {
        "type": "zipf",
        "s": 2.0,
        "v": 1.0,
        "n": 1000000
      },
      "fairnessWeight": {
        "type": "uniform",
        "min": 0.001,
        "max": 1.0
      }
    },
    "low": {
      "weight": 3,
      "sleepDuration": {
        "type": "discrete",
        "weights": {
          "5s": 3,
          "10s": 1
        }
      },
      "priorityKeys": {
        "type": "uniform",
        "min": "1",
        "max": "5"
      },
      "fairnessKeys": {
        "type": "zipf",
        "s": 2.0,
        "v": 1.0,
        "n": 1000000
      },
      "fairnessWeight": {
        "type": "uniform",
        "min": 0.001,
        "max": 1.0
      }
    }
  }
}
```

### DIFF
--- a/workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
+++ b/workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
@@ -365,7 +365,7 @@ public class KitchenSinkWorkflow
         {
             throw new ApplicationFailureException("FairnessKey is not supported yet");
         }
-        if (eaa.FairnessWeight != null)
+        if (eaa.FairnessWeight > 0)
         {
             throw new ApplicationFailureException("FairnessWeight is not supported yet");
         }


workers/go/throughputstress/workflow.go
```diff
@@ -138,15 +139,17 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
...
+				if actInput.SleepDuration == 0 {
 					return nil
 				}
 				opts := defaultActivityOpts(ctx)
-				opts.Priority = temporal.Priority{
-					PriorityKey: int(actInput.Priority),
+				opts.Priority = actInput.Priority
+				if actInput.FairnessKey != "" {
+					// TODO(fairness): hack until there is a fairness key in the SDK
+					opts.ActivityID = fmt.Sprintf("x-temporal-internal-fairness-key[%s]-%s", actInput.FairnessKey, uuid.New().String())
```
stephanos, line 152: This part might be contentious. I don't think anyone outside of the Fairness crew would even consider using this; so I think it's fine/safe.

It would make the Fairness crew's life easier if we could get this PR into main; but if there are strong objections here, we can find another way to run off the branch.

workers/go/throughputstress/workflow.go
```diff
@@ -138,19 +139,29 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
...
+				sleepInputs := params.SleepActivities.Sample()
+
+				var sleepFuncs []func(workflow.Context) error
+				for _, actInput := range sleepInputs {
+					sleepFuncs = append(sleepFuncs, func(ctx workflow.Context) error {
+						opts := defaultActivityOpts(ctx)
+						opts.Priority.PriorityKey = int(actInput.PriorityKey)
+						if actInput.FairnessKey != "" {
+							// TODO(fairness): hack until there is a fairness key in the SDK
+							opts.ActivityID = fmt.Sprintf("x-temporal-internal-fairness-key[%s:%f]-%s",
```
stephanos, line 164: This part might be contentious. I don't think anyone outside of the Fairness crew would even consider using this; so I think it's fine/safe.

README.md
```diff
@@ -158,25 +158,22 @@ for more.
...
 
-The configuration is done via a JSON file, which is passed to
-the scenario with the `--option sleep-activity-per-priority-json=@<file>` flag. Example:
+The configuration is done via a JSON file, which is passed to the scenario with the
+`--option sleep-activity-per-priority-json=@<file>` flag. Example:
 
 ```
-echo '{"patterns_dist": {"type":"discrete", "weights": {"1":1, "5":9}}, "pattern_durations_dist": {"1": {"type":"discrete", "weights": {"1s": 1}}, "5": {"type":"discrete", "weights": {"1s": 1, "5s": 4}}}}' > sleep.json
-go run ./cmd run-scenario-with-worker --scenario throughput_stress --language go --option sleep-activity-per-priority-json=@sleep.json --run-id default-run-id
+echo '{"count":{"type":"fixed","value":5},"groups":{"high":{"weight":2,"sleepDuration":{"type":"uniform","min":"2s","max":"4s"}},"low":{"weight":3,"sleepDuration":{"type":"discrete","weights":{"5s":3,"10s":1}}}}}' > sleep.json
```
dnr, line 169:
```suggestion
echo '{
  "count": {
    "type": "fixed",
    "value": 5
  },
  "groups": {
    "high": {
      "weight": 2,
      "sleepDuration": {
        "type": "uniform",
        "min": "2s",
        "max": "4s"
      }
    },
    "low": {
      "weight": 3,
      "sleepDuration": {
        "type": "discrete",
        "weights": {
          "5s": 3,
          "10s": 1
        }
      }
    }
  }
}' > sleep.json
```

loadgen/throughputstress/throughput_stress.go
```diff
@@ -24,21 +26,136 @@ type WorkflowParams struct {
...
+	Count *loadgen.DistributionField[int64] `json:"count"`
+
+	// Map of groups to their configuration. Required.
+	Groups map[string]SleepActivityGroupConfig `json:"groups"`
+}
+
+// SleepActivityGroupConfig defines a group configuration.
+type SleepActivityGroupConfig struct {
+	// Weight for this group, used for sampling. Defaults to 1.
+	Weight int `json:"weight"`
```
dnr, line 54: float?

loadgen/throughputstress/throughput_stress.go
```diff
@@ -24,21 +26,136 @@ type WorkflowParams struct {
...
+	indexWeights := make(map[int64]int, len(config.Groups))
+	for groupID := range config.Groups {
+		groupIDs[i] = groupID
+		indexWeights[i] = max(1, config.Groups[groupID].Weight)
+		i++
+	}
+	indexDist := loadgen.NewDiscreteDistribution(indexWeights)
+
+	activities := make([]SleepActivityInput, 0, count)
+	for i := int64(0); i < count; i++ {
```
dnr: nit:
```suggestion
	for range count {
```

loadgen/throughputstress/throughput_stress.go
```diff
@@ -24,21 +26,136 @@ type WorkflowParams struct {
...
+		}
+
+		activities = append(activities, instance)
+	}
+
+	return activities
+}
+
+// SleepActivityInput represents the input for a sleep activity.
+type SleepActivityInput struct {
```
dnr: put at the top of the file?

workers/go/throughputstress/workflow.go
```diff
@@ -138,19 +139,29 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
...
-				if actInput == nil {
+				// This activity simulates custom traffic patterns by generating activities that sleep
+				// for a specified duration, with optional priority and fairness keys.
+				if params.SleepActivities == nil {
 					return nil
 				}
-				opts := defaultActivityOpts(ctx)
-				opts.Priority = temporal.Priority{
-					PriorityKey: int(actInput.Priority),
+				sleepInputs := params.SleepActivities.Sample()
```
dnr: are there determinism concerns here? where does this get its random numbers from?

workers/go/throughputstress/workflow.go
```diff
@@ -138,19 +139,29 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
...
+						}
+						opts.StartToCloseTimeout += actInput.SleepDuration // make sure there's enough time for the sleep
+						actCtx := workflow.WithActivityOptions(ctx, opts)
+						return workflow.ExecuteActivity(actCtx, activityStub.Sleep, actInput).Get(ctx, nil)
+					})
 				}
-				opts.StartToCloseTimeout += actInput.SleepDuration // make sure there's enough time for the sleep
-				actCtx := workflow.WithActivityOptions(ctx, opts)
-				return workflow.ExecuteActivity(actCtx, activityStub.Sleep, actInput).Get(ctx, nil)
+				return workflowutils.RunConcurrently(ctx, sleepFuncs...)
```
dnr, line 171: so much concurrency!

dnr (approved):
I'm not familiar with omes at all so can't say much about the context, but the changes look like they do what we're looking for

README.md
stephanos: I think it's better to keep it in a single line because it's just an example to use/modify locally. I wouldn't want to bloat the README just for this snippet. It would claim more screen real-estate that it should IMO.

loadgen/throughputstress/throughput_stress.go
stephanos: Math and reasoning seems easier as int to me. With float you have to clarify whether it means percentage (plus validation) or not. int is just int. And you can do the same things with the int.

workers/go/throughputstress/workflow.go
stephanos: I wasn't trying to be deterministic. The sampling just uses the global `rand` package. Could make it deterministic, not sure it's worth it.

README.md
dnr: that's fine. it was just harder to see without formatting

loadgen/throughputstress/throughput_stress.go
dnr: aren't these values just relative to each other? it's neither a percentage nor a fraction of 1...

with an int if you want to do like 1½:1, you have to make it 3:2 instead. yeah, you can still do any ratio, it just may be harder to write. fine either way though.

workers/go/throughputstress/workflow.go
dnr: I mean, it's a workflow, so it "has to" be deterministic. the normal thing would be to put all this stuff in a SideEffect. if it's only activity inputs, then maybe it doesn't matter.... but the activity ids will be different on replay, so I'm worried the SDK will get confused

loadgen/throughputstress/throughput_stress.go
```diff
@@ -24,21 +26,136 @@ type WorkflowParams struct {
...
+	FairnessWeight *loadgen.DistributionField[float32] `json:"fairnessWeight"`
+}
+
+// SleepActivityInput represents the input for a sleep activity.
+type SleepActivityInput struct {
+	SleepDuration  time.Duration `json:"sleepDuration"`
+	PriorityKey    int64         `json:"priorityKey"`
+	FairnessKey    string        `json:"fairnessKey"`
+	FairnessWeight float32       `json:"fairnessWeight"`
+}
```
Sushisource: Consider defining this in a proto like I've done with the kitchen sink stuff -- it'll be really helpful when it comes time to do these in multiple languages.

(As well as the workflow input, etc). I'm not 100% sure if this particular thing will need to be in all langs, but, I want to at least have that option and the cost seems low to do inputs in proto. If more and more gets added, at least we're not digging a deeper hole.

Alternatively, I mentioned maybe implementing this scenario generically in terms of the kitchen sink workflow, and growing that capability would be even easier for porting later, but I understand that's a big ask.

stephanos: 👍 I can change `kitchensink.ExecuteActivityAction` to include the new fields; and use that as the input instead.

I can't do the other structs right now tbh since the `DistributionField` stuff etc. would take a bit longer. I'm worried that's a rabbit hole. Could try an AI agent for that in a follow-up PR.

As much as I'd like to port over the whole thing, I really don't have the time right now. 

In terms of adding the functionality to non-Go workers; is the best practice to silently ignore non-applicable fields or fail hard?

workers/go/throughputstress/workflow.go
stephanos: right; we're is inside a workflow ... I forgot about that. I've wrapped it inside a side effect now.

loadgen/throughputstress/throughput_stress.go
Sushisource: Greatly appreciate it. No worries about not having time to port. I would say just fail if they're not implemented.

PR #159, opened 2025-07-02
Float distribution value and fixed distribution type

stephanos:
## What was changed

(1) New distribution value type float32.
(2) New distribution type "fixed" that always returns the same value.
(3) Expanded existing validation to all types.

PS: This was all AI generated and reviewed/tweaked by me.

## Why?

(1) To support fairness weights.
(2) To allow specifying a constant value when no distribution is actually needed.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/loadgen/helpers.go
+++ b/loadgen/helpers.go
@@ -12,7 +12,9 @@ import (
 )
 
 // distValueType constrains the types that can be used as distribution values.
-type distValueType interface{ int64 | time.Duration }
+type distValueType interface {
+	int64 | time.Duration | float32
+}
 
 type distribution[T distValueType] interface {
 	// Sample returns a random value from the distribution.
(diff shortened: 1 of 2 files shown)


loadgen/helpers.go
```diff
@@ -322,29 +400,73 @@ func (d normalDistribution[T]) GetType() string {
...
+		}
+		if any(d.max).(time.Duration) < any(d.min).(time.Duration) {
+			return fmt.Errorf("max value (%v) cannot be less than min value (%v)", d.max, d.min)
+		}
 	}
-
 	return nil
 }
 
 func (d normalDistribution[T]) Sample() (T, bool) {
```
Sushisource, line 429: This function is the most disturbingly "Go" code I've seen in a while

stephanos: 😬

PR #160, opened 2025-07-03
Ebb and flow scenario

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,172 @@
...
+	defer e.runningActivities.Add(-int64(count))
+
+	actionSet := &kitchensink.ActionSet{
+		Actions:    []*kitchensink.Action{},
+		Concurrent: true,
+	}
+	delayInSec := int64(sleep.Seconds())
+	for i := 0; i < int(count); i++ {
+		actionSet.Actions = append(actionSet.Actions,
+			&kitchensink.Action{
```
Sushisource: 🙏 The kitchen sink! Thanks!!

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,172 @@
...
+	"go.temporal.io/api/common/v1"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+const maxConsecutiveErrors = 5
+
+func init() {
+	loadgen.MustRegisterScenario(loadgen.Scenario{
+		Description: "Spawns activities to upper bound, drains to lower bound, rinse and repeat. " +
+			"Options: min-activities (default 1), max-activities (default 10), " +
```
dnr: mismatch in defaults?

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,172 @@
...
+	iter := 1
+	var draining bool
+	var startWG sync.WaitGroup
+
+	e.Logger.Infof(
+		"Starting ebb and flow scenario: min=%d, max=%d, rate=%d/s, sleep=%v, duration=%v",
+		minActivities, maxActivities, spawnRatePerSec, activitySleep, e.Configuration.Duration,
+	)
+
+	for elapsed := time.Duration(0); elapsed < e.Configuration.Duration; elapsed = time.Since(time.Now()) {
```
dnr: pretty sure that update clause isn't going to work

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,172 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-errCh:
+			if err == nil {
+				consecutiveErrCount = 0
+			} else {
+				consecutiveErrCount++
+				if consecutiveErrCount >= maxConsecutiveErrors {
+					return fmt.Errorf("got %v consecutive errors, most recent: %w", maxConsecutiveErrors, err)
```
dnr, line 104: should we cancel/clean up on abort?

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,172 @@
...
+			"activity-sleep (default 1s). Duration must be set.",
+		Executor: loadgen.ExecutorFunc(func(ctx context.Context, runOptions loadgen.ScenarioInfo) error {
+			return (&ebbAndFlow{ScenarioInfo: runOptions}).run(ctx)
+		}),
+	})
+}
+
+type ebbAndFlow struct {
+	loadgen.ScenarioInfo
+	runningActivities atomic.Int64
```
dnr: The goal is to send load while keeping a target standing backlog size.. I'm not really seeing how this achieves that. What's preventing the worker from polling quickly and just keeping the backlog zero or very small here? It seems like we'd need to use a slotsupplier or something to get the worker to stop/slow down polling

scenarios/ebb_and_flow.go
stephanos: Most Omes scenarios don't spend any/too much effort on cleanup. I think it's reasonable to expect the namespaces to be ephemeral in the sense that an Omes user will abandon them after the test run.

loadgen/throughputstress/throughput_stress.go
```diff
@@ -33,133 +24,10 @@ type WorkflowParams struct {
...
+	SleepActivities *loadgen.SleepActivityConfig `json:"sleepActivities"`
 }
 
 type WorkflowOutput struct {
 	// The total number of children that were spawned across all continued runs of the workflow.
 	ChildrenSpawned int `json:"childrenSpawned"`
 }
-
-// SleepActivityConfig defines the configuration for sleep activities with flexible distribution support
-type SleepActivityConfig struct {
```
stephanos, line 45: Moved to own file for reusability.

loadgen/helpers.go
```diff
@@ -2,653 +2,12 @@ package loadgen
...
 	"fmt"
-	"math/rand"
-	"sort"
-	"strconv"
 	"time"
 
 	"go.temporal.io/api/workflowservice/v1"
 )
 
-// distValueType constrains the types that can be used as distribution values.
```
stephanos, line 15: Split all this out into its own file.

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if e.maxBacklog <= e.minBacklog {
+		return fmt.Errorf("max-backlog must be greater than min-backlog")
+	}
+	if e.frequency <= 0 {
+		return fmt.Errorf("frequency must be greater than 0")
+	}
+
+	// Activity config
+	var sleepActivityConfig *loadgen.SleepActivityConfig
```
stephanos, line 62: `SleepActivityConfig` isn't a perfect fit, but it gets us very far without any extra work.

loadgen/helper_dist.go
```diff
@@ -0,0 +1,654 @@
...
+	if len(d.weights) == 0 {
+		return zero, false
+	}
+	if len(d.weights) == 1 {
+		for v := range d.weights {
+			return v, true
+		}
+	}
+
+	keys := make([]T, 0, len(d.weights))
```
dnr: is this cacheable? maybe using a weak map or something like that?

loadgen/helper_dist.go
```diff
@@ -0,0 +1,654 @@
...
+	var zero T
+	rawMsg, ok := rawMap[fieldName]
+	if !ok {
+		return zero, fmt.Errorf("missing required field '%s'", fieldName)
+	}
+	if len(rawMsg) == 0 {
+		return zero, fmt.Errorf("field '%s' cannot be empty", fieldName)
+	}
+
+	switch any(zero).(type) {
```
dnr, line 572: we can't just unmarshal into a value of type T? I'm not sure I get why all the extra work

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,251 @@
...
+	// Override activity count to fixed rate.
+	fixedDist := loadgen.NewFixedDistribution(int64(rate))
+	config := loadgen.SleepActivityConfig{
+		Count:  &fixedDist,
+		Groups: template.Groups,
+	}
+
+	// Generate activities.
+	var activities []*kitchensink.ExecuteActivityAction
+	activities = append(activities, config.Sample(e.rng)...)
```
dnr: weird way to write this?
```suggestion
	activities := config.Sample(e.rng)
```
am I missing something?

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// Start one workflow per activity.
```
dnr, line 162: is this okay? is there any reason to try to batch multiple activities in a workflow?

scenarios/ebb_and_flow.go
```diff
@@ -0,0 +1,251 @@
...
+				if consecutiveErrCount >= maxConsecutiveErrors {
+					return fmt.Errorf("got %v consecutive errors, most recent: %w", maxConsecutiveErrors, err)
+				}
+			} else {
+				consecutiveErrCount = 0
+			}
+		case <-ticker.C:
+			generated := e.generatedCount.Load()
+			processed := e.processedCount.Load()
+			backlog := generated - processed
```
dnr, line 112: any concerns about the discrepancy between how this is computed and the "actual" backlog? I don't think so, and the actual number is too hard to get anyway, so nothing to do

dnr (approved):
the rest looks great. unlike my unit test, this actually checks that the backlog hits the targets!

scenarios/ebb_and_flow.go
stephanos: 👍 Ah, there was a for loop before and I didn't inline it all the way.

loadgen/helper_dist.go
stephanos: I wrote/generated this part weeks ago; I'll have to try that again.

scenarios/ebb_and_flow.go
stephanos: Initially, I was starting _all_ activities per tick in _one_ workflow. But I found it difficult to get the backlog target right since you would get a huge batch of activities finishing all at once. So I went to the most granular option. We could do _some_ batching, but now it becomes another parameter to configure and tweak.

Is your concern overloading the system with workflows and taking capacity away from the activities?

scenarios/ebb_and_flow.go
stephanos: Yeah, I figured this is the best we can do from here. We could add an extra buffer to the min/max to move it into the 'actual' backlog numbers inside of Temporal. I didn't want to hard-code that either.

loadgen/helper_dist.go
stephanos: I hadn't considered any performance optimizations as I thought it'll be "fast enough". I could throw caching in there for sure.

loadgen/helper_dist.go
stephanos: Okay, threw in `err := json.Unmarshal(rawMsg, &zero)` and most are failing. One reason, for example, is `time.Duration` which cannot be JSON deserialized out of the box 🙄

scenarios/ebb_and_flow.go
stephanos: I'll add a parameter so we can tweak it if needed after we merge.

loadgen/helper_dist.go
stephanos: Okay, I'm adding per-struct caching now instead. Not the same as a weak map-backed global cache, but I doubt there's going to be a lot of repeated distributions - this way is much easier since I don't need to figure out a good global caching key.

scenarios/ebb_and_flow.go
stephanos: Hm, actually, I think I'll revert that again. Running it it seems it's struggling more to meet the target now. I think the batching causes more instability.

PR #166, opened 2025-07-14
Fix SleepActivityConfig

stephanos:
## What was changed

Using a random seed instead of wrapping randomly generated output in `workflow.SideEffect`.

## Why?

`workflow.SideEffect` doesn't work with arrays of proto messages. The serialization breaks the result in this case, as one of the fields is not JSON exportable.

## Checklist

**Reviewer: Looks bigger than it really is. I marked the two actual changes.**

### DIFF
--- a/loadgen/helpers.go
+++ b/loadgen/helpers.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"sort"
 	"strconv"
 	"time"
 
@@ -17,8 +18,8 @@ type distValueType interface {
 }
 
 type distribution[T distValueType] interface {
-	// Sample returns a random value from the distribution.
-	Sample() (T, bool)
+	// Sample returns a random value from the distribution using the provided seed.
+	Sample(seed int64) (T, bool)
 	// GetType returns the distribution type identifier.
 	GetType() string
 	// Validate checks if the distribution is valid.
@@ -126,7 +127,7 @@ func (d fixedDistribution[T]) Validate() error {
 	return nil
 }
 
-func (d fixedDistribution[T]) Sample() (T, bool) {
+func (d fixedDistribution[T]) Sample(_ int64) (T, bool) {
 	return d.value, true
 }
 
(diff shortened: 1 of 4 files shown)


loadgen/helpers.go
```diff
@@ -17,8 +18,8 @@ type distValueType interface {
 }
 
 type distribution[T distValueType] interface {
-	// Sample returns a random value from the distribution.
-	Sample() (T, bool)
+	// Sample returns a random value from the distribution using the provided seed.
+	Sample(seed int64) (T, bool)
```
stephanos: `Sample` now takes a seed.

workers/go/throughputstress/workflow.go
```diff
@@ -146,14 +145,14 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
...
 
-				var sleepInputs []*kitchensink.ExecuteActivityAction
-				genSleepInputs := workflow.SideEffect(ctx, func(ctx workflow.Context) any {
-					return params.SleepActivities.Sample()
+				generateSeed := workflow.SideEffect(ctx, func(ctx workflow.Context) any {
+					return time.Now().UnixNano()
 				})
-				genSleepInputs.Get(&sleepInputs)
+				var random int64
+				generateSeed.Get(&random)
```
stephanos: Generating and using a seed here.

loadgen/helpers.go
dnr: I'm thinking instead of a seed, it should take a rand.Rand. That way you can sample multiple distributions in sequence

loadgen/throughputstress/throughput_stress.go
```diff
@@ -92,40 +93,43 @@ func ParseAndValidateSleepActivityConfig(jsonStr string) (*SleepActivityConfig,
...
 
 // Sample generates a list of SleepActivityInput instances based on the SleepActivityConfig.
-func (config *SleepActivityConfig) Sample() []*kitchensink.ExecuteActivityAction {
+func (config *SleepActivityConfig) Sample(seed int64) []*kitchensink.ExecuteActivityAction {
 	if config == nil {
 		return nil
 	}
 
-	count, ok := config.Count.Sample()
+	count, ok := config.Count.Sample(seed)
```
dnr: using the same `seed` multiple times here is a little worrying, the selections may be correlated. using the seed once and then passing around the rand.Rand seems better.

workers/go/throughputstress/workflow.go
dnr: you can use workflow.Now().UnixNano() instead of a sideeffect

PR #167, opened 2025-07-15
Fix throughput_stress NDE

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


workers/go/throughputstress/workflow.go
```diff
@@ -157,8 +156,8 @@ func ThroughputStressWorkflow(ctx workflow.Context, params *throughputstress.Wor
 						}
 						if actInput.FairnessKey != "" {
 							// TODO(fairness): hack until there is a fairness key in the SDK
-							opts.ActivityID = fmt.Sprintf("x-temporal-internal-fairness-key[%s:%f]-%s",
-								actInput.GetFairnessKey(), actInput.GetFairnessWeight(), uuid.New().String())
+							opts.ActivityID = fmt.Sprintf("x-temporal-internal-fairness-key[%s:%f]-%d",
+								actInput.GetFairnessKey(), actInput.GetFairnessWeight(), workflow.Now(ctx).UnixNano())
```
dnr: workflow.Now() returns the same value for the extent of the whole wft so these could collide.. why not use the rng, e.g. rng.Uint64()? or for more entropy you could use rng.Read into a buffer and hex-encode

stephanos: Ah, it helps to read the docs

> Now returns the time when the workflow task was first started

Which is not what I expected.

> why not use the rng, e.g. rng.Uint64()

Almost too easy.

PR #168, opened 2025-07-16
throughput_stress retries

stephanos:
## What was changed

Added retries for the Workflow Start.

## Why?

When we run the throughput_stress scenario with faults in our CICD pipeline, the workflow start can fail. Omes will abort the run immediately. CICD will restart Omes and resume.

All that creates unnecessary delays and noise. Internal retries mitigate that.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/scenarios/throughput_stress.go
+++ b/scenarios/throughput_stress.go
@@ -24,10 +24,10 @@ const (
 	ContinueAsNewAfterIterFlag = "continue-as-new-after-iterations"
 	SkipSleepFlag              = "skip-sleep"
 	NexusEndpointFlag          = "nexus-endpoint"
-	// MaxAttemptFlag is a flag to set the maximum number of attempts for starting a workflow.
-	MaxAttemptFlag = "max-attempts"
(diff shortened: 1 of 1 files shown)


scenarios/throughput_stress.go
```diff
@@ -24,6 +24,10 @@ const (
 	ContinueAsNewAfterIterFlag = "continue-as-new-after-iterations"
 	SkipSleepFlag              = "skip-sleep"
 	NexusEndpointFlag          = "nexus-endpoint"
+	// MaxAttemptFlag is a flag to set the maximum number of attempts for starting a workflow.
+	MaxAttemptFlag = "max-attempts"
+	// MaxAttemptBackoffFlag is a flag to set the maximum backoff time between attempts for starting a workflow.
+	MaxAttemptBackoffFlag = "max-attempt-backoff"
```
Sushisource: Maybe call these MaxStartAttempt

Sushisource (approved):

PR #173, opened 2025-07-18
Build proto in Docker

stephanos:
## What was changed

(1) Provides script to build protos locally.
(2) Moves all version identifier to file.

## Why?

(1) Allow contributors to build correct protos locally.
(2) Single source of truth.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


scripts/build_proto.sh
```diff
@@ -0,0 +1,29 @@
...
+  --build-arg PROTOC_GEN_GO_VERSION="${PROTOC_GEN_GO_VERSION}" \
+  --build-arg NODE_VERSION="${NODE_VERSION}" \
+  -t "${IMAGE_NAME}" "${PROJECT_ROOT}"
+
+echo "Running kitchen-sink-gen build..."
+docker run --rm -v "${PROJECT_ROOT}:/workspace" -w /workspace "${IMAGE_NAME}" bash -c "
+    cd /workspace/loadgen/kitchen-sink-gen
+    cargo build
+    cd /workspace/workers/typescript
+    npm install && npm run proto-gen
```
stephanos, line 26: Easy to forget ...

README.md
```diff
@@ -10,9 +10,6 @@ Omes (pronounced oh-mess) is the Hebrew word for "load" (עומס).
 ## Prerequisites
 
 - [Go](https://golang.org/) 1.24+
-  - `protoc` and `protoc-gen-go` must be installed
-  - tip: don't worry about the specific versions here; instead, the GitHub action will make a diff
-    available for download that you can use with `git apply`
```
Sushisource, line 15: Might be worth changing this to say you _either_ need proto or docker to get kitchen sink going locally

scripts/build_proto.sh
```diff
@@ -0,0 +1,29 @@
...
+docker build -f "${PROJECT_ROOT}/dockerfiles/proto.Dockerfile" \
+  --build-arg RUST_TOOLCHAIN="${RUST_TOOLCHAIN}" \
+  --build-arg GO_VERSION="${GO_VERSION}" \
+  --build-arg PROTOC_VERSION="${PROTOC_VERSION}" \
+  --build-arg PROTOC_GEN_GO_VERSION="${PROTOC_GEN_GO_VERSION}" \
+  --build-arg NODE_VERSION="${NODE_VERSION}" \
+  -t "${IMAGE_NAME}" "${PROJECT_ROOT}"
+
+echo "Running kitchen-sink-gen build..."
+docker run --rm -v "${PROJECT_ROOT}:/workspace" -w /workspace "${IMAGE_NAME}" bash -c "
```
Sushisource, line 22: Not a huge fan of mounting the whole project dir into the docker image. This can make for slow loads as it copies everything, and what it writes back out often has weird, incorrect permissions.

That said, this is still easier than before, so I wouldn't block on this, but, if possible might be nice to cleanly copy the result files out.

Sushisource (approved):

scripts/build_proto.sh
stephanos: 👍 Good idea!

stephanos: Hm, I spent almost an hour on this and it's harder than I thought: I don't want to copy all files into the container (node_modules is large) which requires some filtering, I don't want to specify every single output because that's brittle, but copying all files out of the container also defeats the purpose, doing a diff is harder than it seems, ...

I'll merge for now and we can discuss improvements.

PR #174, opened 2025-07-19
Verify fairness

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cmd/build_worker_image.go
```diff
@@ -105,6 +105,7 @@ func (b *workerImageBuilder) build(ctx context.Context) error {
 		// including semver build metadata.
 		langTagComponent := lang + "-" + strings.TrimPrefix(b.version, "v")
 		b.tags = append(b.tags, omesVersion+"-"+langTagComponent)
+		b.tags = append(b.tags, lang+"-"+omesVersion)
```
stephanos, line 108: Also publish under the tag like `go-979932c` so CICD can automatically find the worker version for its Omes dependency.
Screenshot 2025-07-18 at 10 28 15 PM

loadgen/ebbandflow/fairness_tracker_test.go
```diff
@@ -0,0 +1,376 @@
+package ebbandflow
```
stephanos, line 1: I had this generated to make Claude verify the algorithms work; I don't feel strongly about checking them in or not.

scenarios/throughput_stress.go
```diff
@@ -251,32 +249,6 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 	)
 }
 
-func (t *tpsExecutor) initSearchAttribute(ctx context.Context, info loadgen.ScenarioInfo) error {
```
stephanos, line 254: Moved to re-usable helper.

loadgen/ebbandflow/fairness_tracker_test.go
Sushisource: May as well keep it in. This is probably the most statistics code anyone's ever added at Temporal lol

workers/go/ebbandflow/workflow.go
```diff
@@ -0,0 +1,102 @@
...
+	// Check if all activities failed
+	if len(results) == 0 {
+		return nil, fmt.Errorf("failed to start any of the %d activities", len(activities))
+	}
+
+	return &ebbandflow.WorkflowOutput{Timings: results}, nil
+}
+
+// EbbAndFlowReportWorkflow receives fairness reports and handles logging and metrics
+func EbbAndFlowReportWorkflow(ctx workflow.Context, report ebbandflow.FairnessReport) error {
```
Sushisource, line 93: Hmm almost like we should build some kind of stand-alone activity... 😄

Sushisource (approved):
I did not look closely at the statistics functions, but the workflow/structure looks good to me 👍

loadgen/ebbandflow/fairness_tracker_test.go
stephanos: I can't claim much ownership over any of them as statistics is not my strong suit; we'll see if Claude did a good job there - it made sense to me :D

loadgen/ebbandflow/ebb_and_flow.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+type WorkflowOutput struct {
+	Timings []ActivityTiming `json:"timings"`
+}
+
+type ActivityTiming struct {
+	FairnessKey       string        `json:"fairnessKey"`
+	FairnessWeight    float32       `json:"fairnessWeight"`
+	ScheduleToStartMS time.Duration `json:"scheduleToStartMS"`
```
dnr, line 20: it's a duration?
```suggestion
	ScheduleToStart time.Duration `json:"scheduleToStart"`
```
also maybe make these json keys shorter, e.g. single char, since there's going to be a big array of them?

loadgen/ebbandflow/fairness_tracker.go
```diff
@@ -0,0 +1,454 @@
...
+
+	heapData := make(map[string][]float64, len(ft.latencies))
+	weights := make(map[string]float64, len(ft.weights))
+	for key, h := range ft.latencies {
+		if h.Len() == 0 {
+			continue
+		}
+		values := make([]float64, h.Len())
+		copy(values, *h)
+		heapData[key] = values
```
dnr, line 103: does `slices.Clone` work here? (with some casting)

loadgen/ebbandflow/fairness_tracker.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, fmt.Errorf("need at least 2 keys with data, got %d", len(p95Values))
+	}
+
+	// Extract P95 values for distribution analysis
+	p95Slice := make([]float64, len(p95Values))
+	i := 0
+	for _, p95 := range p95Values {
+		p95Slice[i] = p95
+		i++
+	}
```
dnr, line 133: maps.Values? (golang.org/x/exp/maps)

loadgen/ebbandflow/fairness_tracker.go
```diff
@@ -0,0 +1,454 @@
...
+	sort.Float64s(p95Slice)
+
+	// Calculate weight-adjusted P95 values for all metrics.
+	weightAdjustedP95s := make(map[string]float64, len(p95Values))
+	for key, p95 := range p95Values {
+		if weight := weights[key]; weight > 0 {
+			weightAdjustedP95s[key] = p95 / weight
```
dnr, line 140: are we mixing up the two usages of "weight" here?

there's "the FairnessWeight value that we give to the key to control the proportion of tasks dispatched" and "the number of tasks from this key that we created", which in other contexts we've been calling "heaviness".

suppose we give everyone the same FairnessWeight to everyone, and one key has 10x as many tasks as all the rest. the desired behavior of fairness is to "slow down" that one for the benefit of the others. so its p95 will be higher. but this weightAdjustedP95s will be adjusted by the default 1.0 FairnessWeight.

it seems like need to adjust by the number of tasks created? of course if we create more tasks we expect higher average (and p95) task latency, and that's true whether we're being fair or not.

----

in the other direction, suppose we have a bunch of keys and give one of them FairnessWeight=10, the others default to 1. each key has an equal number of tasks generated. the one with higher FairnessWeight should have a _lower_ latency since we dispatched its tasks first.

loadgen/ebbandflow/fairness_tracker.go
```diff
@@ -0,0 +1,454 @@
...
+		}
+	}
+
+	// Approach 1: Weight-Adjusted Proportional Fairness
+	// Normalizes each key's P95 by weight (P95/weight), then checks if the distribution
+	// of normalized values is flat. Higher weight keys should get lower latencies, so
+	// after normalization all keys should have similar values. We compare P95-of-normalized
```
dnr, line 147: yes, keys with higher FairnessWeight should have lower latencies, but we're dividing by FairnessWeight above, right? so a lower number divided by a higher number is doubly lower... we should multiply instead

PR #171, opened 2025-07-18
Port throughput_stress to kitchensink

stephanos:
## What was changed

Ported the throughput_stress scenario to use kitchensink.

## Why?

(1) Allows re-using it in other SDKs.
(2) Moves all the scenario code into one place.
(3) Makes scenario generally easier to test in CICD since a worker change might not be needed.
(4) Improves debugging since inputs are instructive and readable (at least most of them).

## Checklist

1. Closes 

2. How was this tested:

Ran a longer local test, too:

```
go run ./cmd run-scenario-with-worker \
  --scenario throughput_stress \
  --language go \
  --iterations 50 \
  --max-concurrent 20 \
  --option internal-iterations=5 \
  --option continue-as-new-after-iterations=2 \
  --option sleep-time=1ms
```

Some of the kitchensink tests seem to be flaky, though ... will need to address that.

3. Any docs updates needed?

PR #175, opened 2025-07-19
Nexus in kitchensink

stephanos:
## What was changed

Add executing a Nexus operation to kitchensink.

Out of scope for this PR: non-Go implementations.

## Why?

Remove blocker for [porting throughput_stress to kitchensink](https://github.com/temporalio/omes/pull/171).

## Checklist

Will test more thoroughly in https://github.com/temporalio/omes/pull/171


workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -433,3 +435,17 @@ enum ActivityCancellationType {
   ABANDON = 2;
 }
 
+// Execute a Nexus operation
+message ExecuteNexusOperation {
```
stephanos, line 439: The new kitchensink action.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -357,3 +401,29 @@ type ReturnOrErr struct {
 	retme *common.Payload
 	err   error
 }
+
+var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Context, s string, opts nexus.StartOperationOptions) (string, error) {
```
stephanos, line 392: All of these were moved 1:1 from throughput_stress.

workers/go/throughputstress/workflow.go
```diff
@@ -248,7 +249,7 @@ func nexusClient(ctx workflow.Context, params *throughputstress.WorkflowParams)
 		workflow.GetLogger(ctx).Debug("not running nexus tests, set nexusEndpoint in options to enable these tests")
 		return nil
 	}
-	return workflow.NewNexusClient(params.NexusEndpoint, ThroughputStressServiceName)
+	return workflow.NewNexusClient(params.NexusEndpoint, kitchensink.KitchenSinkServiceName)
```
stephanos, line 252: This whole workflow will disappear shortly; no need to maintain two Nexus services/names.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -314,6 +321,44 @@ func withAwaitableChoiceCustom[F workflow.Future](
...
 
+func handleNexusOperation(ctx workflow.Context, nexusOp *kitchensink.ExecuteNexusOperation, state *KSWorkflowState) error {
+	client := workflow.NewNexusClient(KitchenSinkServiceName, KitchenSinkServiceName)
+
+	nexusOptions := workflow.NexusOperationOptions{}
+
+	var input interface{}
+	if nexusOp.Input != nil {
+		// For now, assuming the input is a string for simplicity.
+		input = string(nexusOp.Input.Data)
```
stephanos: Not sure if I should make these just string instead of Payload?

Sushisource: I would. Easier to add more later if/when it matters, but, honestly feels like we can just use strings forever for the purposes of omes

workers/go/throughputstress/workflow.go
Sushisource: 🥳

Sushisource (approved):
Awesome

workers/go/kitchensink/kitchen_sink.go
stephanos: ~~Actually, let me put a pin in that; I'm worried I'm gonna chase the linters again for an hour. It might be best to change all `payload`s across the proto to `string` in one go.~~ Right, I only Go right now, so no big deal.

PR #177, opened 2025-07-22
Use Priority for fairness key and weight

stephanos:
## What was changed

Bumped up Go API.

Removed `fairness_key` and `fairness_weight`.


workers/java/build.gradle
```diff
@@ -21,7 +21,7 @@ spotless {
             include '**/*.java'
             exclude 'io/temporal/omes/KitchenSink.java'
         }
-        googleJavaFormat('1.16.0')
+        googleJavaFormat('1.28.0')
```
stephanos, line 24: Older version didn't work with OpenJDK for me.

PR #179, opened 2025-07-29
Client ActivityType [Go]

stephanos:
ℹ️ This PR includes the https://github.com/temporalio/omes/pull/189 PR; you can just look at the "Client ActivityType" commit for review.

## What was changed

Adds new "Client" ActivityType to execute client actions (e.g. Update, Signal) from a workflow activity.

## Why?

- Keeps all action execution in one place (ie the workflow).
- Makes client actions part of the event history (free resume-ability, easier debugging).
- Execution is (more) deterministic.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
+++ b/workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
@@ -140,15 +140,17 @@ public class KitchenSinkWorkflow
         else if (action.Timer is { } timer)
         {
             await HandleAwaitableChoiceAsync(
-                Workflow.DelayAsync((int)timer.Milliseconds, tokenSrc.Token)
-                    .ContinueWith(_ => true),
+                ToBool(Workflow.DelayAsync((int)timer.Milliseconds, tokenSrc.Token)),
                 tokenSrc,
                 timer.AwaitableChoice);
         }
         else if (action.ExecActivity is { } execActivity)
         {
-            await LaunchActivity(execActivity, tokenSrc);
-            await HandleAwaitableChoiceAsync(Task.FromResult<object?>(null), tokenSrc, execActivity.AwaitableChoice);
+            await HandleAwaitableChoiceAsync(
+                ToBool(LaunchActivity(execActivity, tokenSrc)),
+                tokenSrc,
+                execActivity.AwaitableChoice
+            );
         }
         else if (action.ExecChildWorkflow is { } execChild)
         {
@@ -362,6 +364,12 @@ public class KitchenSinkWorkflow
         }
     }
 
+    private static async Task<bool> ToBool(Task task)
+    {
+        await task;
+        return true;
+    }
+
     // Duped for now, if exposed by SDK use it from there.
     private static Temporalio.Common.RetryPolicy RetryPolicyFromProto(RetryPolicy proto)
     {
@@ -432,5 +440,4 @@ public class ClientActivitiesImpl
         var executor = new ClientActionsExecutor(_client, workflowId, taskQueue);
         await executor.ExecuteClientSequence(clientActivity.ClientSequence);
     }
-
 }


loadgen/kitchensink/client_executor.go
```diff
@@ -0,0 +1,190 @@
+package kitchensink
+
+import (
```
stephanos, line 3: Simply etxtracted into separate file.

scripts/clean.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Cleaning TypeScript..."
+cd "$ROOT_DIR/workers/typescript"
+npm run clean 2>/dev/null || true
+
+echo "Cleaning .NET..."
+cd "$ROOT_DIR/workers/dotnet"
+dotnet clean 2>/dev/null || true
+
+echo "✅ Clean completed"
```
stephanos, line 35: Added for (my) convenience. I found it helpful; but no strong feelings.

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
```diff
@@ -146,10 +147,8 @@ await HandleAwaitableChoiceAsync(
         }
         else if (action.ExecActivity is { } execActivity)
         {
-            await HandleAwaitableChoiceAsync(
-                LaunchActivity(execActivity, tokenSrc).ContinueWith(_ => true),
-                tokenSrc,
-                execActivity.AwaitableChoice);
+            await LaunchActivity(execActivity, tokenSrc);
+            await HandleAwaitableChoiceAsync(Task.FromResult<object?>(null), tokenSrc, execActivity.AwaitableChoice);
```
stephanos: Full disclosure; C# is the only language here I have absolutely zero experience in. My AI agent(s) suggested this change since it was swallowing the "unsupported" exception.

loadgen/kitchensink/client_action_executor.go
```diff
@@ -0,0 +1,190 @@
+package kitchensink
```
stephanos, line 1: Moved over here as-is to match all the other SDKs.

loadgen/scenario.go
```diff
@@ -326,8 +329,8 @@ func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSi
 	if executeErr != nil {
 		return fmt.Errorf("failed to execute kitchen sink workflow: %w", executeErr)
 	}
-	if clientActionsErr != nil {
```
stephanos, line 329: Data race here.

workers/run.go
```diff
@@ -128,13 +128,15 @@ func (r *Runner) Run(ctx context.Context, baseDir string) error {
...
 	args = append(args, r.WorkerOptions.ToFlags()...)
 
-	// Start the command. Do not use the context so we can send interrupt.
 	cmd, err := prog.NewCommand(context.Background(), args...)
 	if err != nil {
 		return fmt.Errorf("failed creating command: %w", err)
 	}
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true} // set process group ID for shutdown
+	cmd.Stdout = &logWriter{logger: r.Logger}
+	cmd.Stderr = &logWriter{logger: r.Logger}
```
stephanos, line 137: Otherwise the output isn't correctly associated with the test we're executing.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -159,75 +157,740 @@ func TestKitchensink(t *testing.T) {
...
+					InitialActions: ListActionSet(
+						&Action{Variant: nil}), // unsupported action
+				},
+			},
+			unsupportedSDKs:        sdks,
+			expectedUnsupportedErr: "unrecognized action",
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
stephanos, line 875: Parallel!

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -581,13 +582,28 @@ impl<'a> Arbitrary<'a> for ExecuteActivityAction {
...
+                    Duration::from_millis(delay)
+                        .try_into()
+                        .expect("proto duration works"),
+                )
+            }
+            86..=90 => {
+                execute_activity_action::ActivityType::Payload(u.arbitrary()?)
+            }
+            91..=100 => {
+                execute_activity_action::ActivityType::Client(u.arbitrary()?)
```
stephanos, line 600: Including the new client activity.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -84,6 +96,17 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 			})
 		}
 	})
+
+	// Handle initial set.
```
stephanos, line 100: Needs to be after signal handler is registered.

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -691,6 +707,36 @@ impl<'a> Arbitrary<'a> for UpsertSearchAttributesAction {
     }
 }
 
+impl<'a> Arbitrary<'a> for ClientActivity {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let client_sequence = ClientSequence {
+            action_sets: vec![ClientActionSet {
+                actions: vec![u.arbitrary()?],
+                concurrent: false,
```
Sushisource, line 715: Should this be allowed to be true? Seemingly the support is already there?

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
Sushisource: Hmmm... not sure that's right. Probably we need to just fix `HandleAwaitableChoice` to not eat the exception in this case (though not sure why it is in the first place).

Otherwise I would expect something like `var actAwait = LaunchActivity(execActivity, tokenSrc);` and then passing that to `HandleAwaitableChoiceAsync`

Sushisource (approved):
Looking good to me. The C# bit might need some verification.

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
stephanos: ~let me just roll this part back~ actually, it times out

loadgen/kitchen-sink-gen/src/main.rs
stephanos: The support for Go is there but the support for the other SDKs isn't there yet. https://github.com/temporalio/omes/pull/188 will only add support for non-current client activity there.

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
stephanos: Also tried
```
var actTask = LaunchActivity(execActivity, tokenSrc);
await HandleAwaitableChoiceAsync(
    actTask.ContinueWith(_ => true),
    tokenSrc,
    execActivity.AwaitableChoice
);
```
which also times out.

cretz: ^ That code is effectively the exact same as the original code from a compiler POV, so it wouldn't make sense that one times out and not another. So your issue may be with the original code and not your alteration.

PR #181, opened 2025-07-30
Scripts

stephanos:
## What was changed

Added 3 sets of scripts plus helpers:

- install scripts: using `mise`. optional, other scripts don't depend on `mise`. users are free to ignore these.
- lint scripts: run the formatting and linting/compiling steps
- build scripts: build the docker image and test them

Plus, changed the kitchensink script to use local tools instead of Docker.

To get the full experience locally, run:
```
scripts/install-all.sh
scripts/lint-all.sh
scripts/build-all.sh
```

The CI is using the same scripts, so we know they'll continue to work.
Plus, now CI will use the exact versions (before it sometimes just installed "latest").

## Why?

Make local development easier. 🙏 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: Continuous Integration
 
+env:
+  MISE_VERSION: v2025.8.1
+
 on: # rebuild any PRs and main branch changes
   pull_request:
   push:
@@ -29,7 +32,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Go
         run: ./scripts/install-go.sh
       - name: Lint Go
@@ -52,7 +55,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
           ./scripts/install-java.sh
@@ -77,7 +80,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
           ./scripts/install-python.sh
@@ -104,7 +107,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
           ./scripts/install-node.sh
@@ -131,7 +134,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
           ./scripts/install-dotnet.sh
@@ -156,7 +159,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@v2.4.4
         with:
-          version: v2025.8.1
+          version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
           ./scripts/install-rust.sh


versions.env
```diff
@@ -1,11 +1,14 @@
 # Tool versions
-GO_VERSION=1.21.0
+DOTNET_VERSION=8.0
+GO_VERSION=1.24.5
+GRADLE_VERSION=8.5
```
stephanos, line 4: Gradle version was not specified before.

workers/go/go.mod
```diff
@@ -1,11 +1,10 @@
 module github.com/temporalio/omes/workers/go
 
-go 1.24.2
+go 1.24.5
```
stephanos, line 3: Might as well.

workers/go/ebbandflow/activities.go
```diff
@@ -76,7 +76,7 @@ func (a Activities) ProcessFairnessReport(ctx context.Context, report ebbandflow
 	metricsHandler.Gauge("ebbandflow_fairness_atkinson_index").Update(report.AtkinsonIndex)
 	metricsHandler.Gauge("ebbandflow_fairness_weight_adjusted").Update(report.WeightAdjustedFairness)
 	metricsHandler.Gauge("ebbandflow_fairness_key_count").Update(float64(report.KeyCount))
-	
+
```
stephanos, line 79: From applying `go fmt` (which wasn't enforced by CI before).

.github/workflows/ci.yml
```diff
@@ -7,33 +7,36 @@ on: # rebuild any PRs and main branch changes
       - main
 
 jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Run shellcheck
+        uses: ludeeus/action-shellcheck@master
```
stephanos, line 19: Make sure scripts are all good.

.github/workflows/ci.yml
```diff
@@ -42,137 +45,86 @@ jobs:
...
-        run: docker run --rm --detach -i -p 10233:10233 omes:java-${{ env.JAVA_SDK_VERSION }} --scenario workflow_with_single_noop_activity --log-level debug --language java --run-id ${{ github.run_id }} --embedded-server-address 0.0.0.0:10233
-      - name: Run scenario against image
-        run: ./temporal-omes run-scenario --scenario workflow_with_single_noop_activity --log-level debug --server-address 127.0.0.1:10233 --run-id ${{ github.run_id }} --connect-timeout 1m --iterations 5
+          ./scripts/install-java.sh
+          ./scripts/install-go.sh
+      - name: Lint Java
+        run: SKIP_FORMAT=true ./scripts/lint-java-worker.sh
+      - name: Check Java formatting
+        run: |
+          git diff --exit-code || (echo "Java worker has uncommitted formatting changes after linting" && exit 1)
```
stephanos, line 68: Added these to make sure nothing slips through the cracks.

workers/dotnet/App.cs
```diff
@@ -82,6 +82,13 @@ public static class App
         description: "Prometheus handler path",
         getDefaultValue: () => "/metrics");
 
+    /// <summary>
+    /// Main entry point.
+    /// </summary>
+    /// <param name="args">CLI args.</param>
+    /// <returns>Task for completion.</returns>
+    public static Task Main(string[] args) => RunAsync(args);
```
stephanos, line 90: The linter kept flagging that there's no `Main`. The tests pass with this so it seems good?

workers/dotnet/App.cs
```diff
@@ -87,7 +87,7 @@ public static class App
     /// </summary>
     /// <param name="args">CLI args.</param>
     /// <returns>Task for completion.</returns>
-    public static Task RunAsync(string[] args) => CreateCommand().InvokeAsync(args);
+    public static Task<int> RunAsync(string[] args) => CreateCommand().InvokeAsync(args);
```
stephanos, line 90: Had to make this change to make the linter happy.

workers/dotnet/Temporalio.Omes.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <OutputType>Exe</OutputType>
+        <OutputType>Library</OutputType>
```
stephanos, line 4: I made the lint step compile the project to check it type checks etc and had to change this since there's no `Main`. I see that the build step is generating it's own csproj file, so this seems fine to me.

workers/dotnet/Temporalio.Omes.csproj
```diff
@@ -10,6 +10,10 @@
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     </PropertyGroup>
 
+    <!-- When testing/linting, treat this as a library so no Main method is needed. -->
+    <PropertyGroup Condition="'$(Configuration)' == 'Library'">
+        <OutputType>Library</OutputType>
+    </PropertyGroup>
```
stephanos, line 16: So this is needed because without it, the build will complain that there's no `Main`. But if you define a `Main`, the image build process will complain that there are two `Main`s ...

.github/workflows/ci.yml
```diff
@@ -42,137 +47,94 @@ jobs:
...
-      - name: Run local scenario with worker
-        run: ./temporal-omes run-scenario-with-worker --scenario workflow_with_single_noop_activity --log-level debug --language python --embedded-server --iterations 5 --version ${{ env.PYTHON_SDK_VERSION }}
-      - name: Build worker image
-        run: ./temporal-omes build-worker-image --language python --version ${{ env.PYTHON_SDK_VERSION }} --tag-as-latest
-      - name: Run worker image
-        run: docker run --rm --detach -i -p 10233:10233 omes:python-${{ env.PYTHON_SDK_VERSION }} --scenario workflow_with_single_noop_activity --log-level debug --language python --run-id ${{ github.run_id }} --embedded-server-address 0.0.0.0:10233
-      - name: Run scenario against image
-        run: ./temporal-omes run-scenario --scenario workflow_with_single_noop_activity --log-level debug --server-address 127.0.0.1:10233 --run-id ${{ github.run_id }} --connect-timeout 1m --iterations 5
+          ./scripts/install-python.sh
+          ./scripts/install-go.sh
```
stephanos: If this is leaning too much on mise, I can also revert this.

While it is an extra dependency, it is one that is very easy to remove again. And with this, we also test the install scripts, too.

scripts/lint-dotnet-worker.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "Error: .NET is not installed. Please install .NET first."
+    exit 1
+fi
+
+echo ".NET version: $(dotnet --version)"
+
+cd "$REPO_ROOT/workers/dotnet"
+
+echo "Cleaning .NET workspace..."
+dotnet clean
```
Sushisource: Not sure why all these clean first, seems unnecessary.

stephanos:
Yes, the scripts were all generated; but I also took significant effort to simplify and standardize as much as possible. That took a few hours actually. I also don't think they will need to be changed much.

scripts/lint-dotnet-worker.sh
stephanos: I had an issue that I thought was related to a partial build; but I'm happy to remove it again since that might have just been a fluke. I'd expect the respective tools to behave reasonably well with partial builds.

EDIT: removed!

PR #182, opened 2025-07-30
Build ARM images

stephanos:
## What was changed

Build Omes Docker images as amd64 _and_ arm64.

## Why?

To use Omes on arm64 machines.

## Checklist

1. Closes 

2. How was this tested:

Screenshot 2025-07-30 at 5 39 11 PM

```
docker run --rm --platform linux/arm64 temporaliotest/omes:go-6659928 --scenario=throughput_stress --run-id=test
2025-07-31T00:47:16.519Z        INFO    cmd/run_worker.go:184   Starting worker with command: [./program --task-queue throughput_stress:test --server-address localhost:7233 --namespace default --prom-handler-path /metrics --log-level info --log-encoding console]
2025-07-31T00:47:16.526Z        FATAL   cmdoptions/client.go:81 failed to dial: failed reaching server: last connection error: connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:7233: connect: connection refused"
```

3. Any docs updates needed?


cmd/build_worker_image.go
```diff
@@ -20,9 +20,26 @@ func buildWorkerImageCmd() *cobra.Command {
...
+			}
+		},
+	}
+	b.addCLIFlags(cmd.Flags())
+	cmd.MarkFlagRequired("language")
+	cmd.MarkFlagRequired("version")
+	return cmd
+}
+
+func buildPushWorkerImageCmd() *cobra.Command {
```
stephanos, line 36: Multi-platform builds need to be build and pushed together; hence the new command.

anmolsahoo25:
@kepe-temporal might help to get omes running in test cells?

.github/workflows/docker-images.yml
```diff
@@ -54,57 +49,87 @@ jobs:
...
+        with:
+          repository: ${{ inputs.sdk-repo-url }}
+          submodules: recursive
+          path: checked-out-sdk/
+          ref: ${{ inputs.sdk-repo-ref }}
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '^1.20'
```
bergundy: Can we bump this?

cmd/build_worker_image.go
```diff
@@ -53,16 +71,17 @@ func (b *workerImageBuilder) addCLIFlags(fs *pflag.FlagSet) {
 		"SDK version to build a worker image for - treated as path if slash present, but must be beneath this dir and at least one tag required")
 	fs.BoolVar(&b.tagAsLatest, "tag-as-latest", false,
 		"If set, tag the image as latest in addition to the omes commit sha tag")
-	fs.StringVar(&b.platform, "platform", "", "Platform for use in docker build --platform")
+	fs.StringVar(&b.platform, "platform", "amd64", "Platform for use in docker build --platform")
```
bergundy: Please make this a `StringSliceVar` instead.

.github/workflows/docker-images.yml
```diff
@@ -54,57 +49,87 @@ jobs:
...
 
-      - name: Push image to DockerHub
-        if: inputs.do-push && !github.event.pull_request.head.repo.fork
-        run: go run ./cmd push-images
+      - name: Build and push to Docker Hub
+        run: |
+          go run ./cmd build-push-worker-image --language ${{ inputs.lang }} \
+          --version ${{ inputs.sdk-version || 'checked-out-sdk/' }} \
+          --platform linux/amd64,linux/arm64 \
+          --repo-prefix temporaliotest \
```
bergundy, line 129: Should this be the default?

stephanos: So in ci.yml we also build the Docker images, for testing. Those do not have that prefix right now, ie they are just tagged with `omes:typescript-...`. So we could either set the prefix here explicitly (which I like since this is for actually pushing them); or tag them with this prefix by default over in ci.yml as well. I prefer the status quo.

stephanos: I'll merge for now; but please feel free to reply if you think we should make a change. Like I said, I prefer this approach, as the only place we're defining `temporaliotest` is right in this file (and it's the only place we push).

PR #183, opened 2025-07-31
Use buildx cache

stephanos:
## What was changed

Add GHA-based Docker layer caching.

## Why?

Without cache a multi-platform build takes about 9min.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stephanos:
Doesn't seem to help.

PR #198, opened 2025-09-10
Convert scripts to Go

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


loadgen/kitchen-sink-gen/build.rs
```diff
@@ -6,6 +6,7 @@ use std::process::Command;
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-changed=../../workers/proto");
+    println!("cargo:rerun-if-changed=../../versions.env");
```
stephanos, line 9: Ensure it re-builds when versions change.

go.mod
```diff
@@ -1,52 +1,53 @@
...
 	github.com/golang/protobuf v1.5.4
 	github.com/google/uuid v1.6.0
-	github.com/prometheus/client_golang v1.16.0
-	github.com/spf13/cobra v1.7.0
-	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.10.0
-	github.com/temporalio/features v0.0.0-20250808182149-bb2a99cdf200
-	go.temporal.io/api v1.50.0
-	go.temporal.io/sdk v1.35.0
+	github.com/prometheus/client_golang v1.23.2
```
stephanos: Drive-by version bump

cmd/dev/build.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+
+	fmt.Println("Building kitchen-sink-gen...")
+	kitchenSinkGenDir := filepath.Join(rootDir, "loadgen", "kitchen-sink-gen")
+
+	// Set PROTOC environment variable to use the correct protoc
+	protocPath, _ := getCommandPath("protoc") // already validated earlier
+	os.Setenv("PROTOC", protocPath)
+	fmt.Printf("Setting PROTOC=%s\n", protocPath)
```
stephanos: It kept using the wrong protoc without this for me.

workers/java/build.gradle
```diff
@@ -25,6 +25,8 @@ spotless {
     }
 }
 
+compileJava.dependsOn spotlessJava
```
stephanos, line 28: This was from this error (local and on CI):

```
Some problems were found with the configuration of task ':compileJava' (type 'JavaCompile').
  - Gradle detected a problem with the following location: '/Users/stephan/Workspace/default/omes/workers/java/build/classes/java/main'.
    
    Reason: Task ':spotlessJava' uses this output of task ':compileJava' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```
Not sure why it came up now, though

cmd/dev/build.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+
+	versions, err := loadVersions()
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Building %s worker(s)...\n", strings.Join(languages, ", "))
```
bergundy: nit: use `Println` everywhere.

cmd/dev/build.go
```diff
@@ -0,0 +1,175 @@
...
+	fmt.Printf("===========================================\n")
+
+	if err := validateLanguageTools(language); err != nil {
+		return err
+	}
+
+	sdkVersion := getSdkVersion(language, versions)
+	fmt.Printf("Building %s worker (SDK %s)...\n", language, sdkVersion)
+
+	if err := buildTemporalOmes(rootDir); err != nil {
```
bergundy: Doesn't seem to belong under `buildWorker`

cmd/dev/build.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+
+	if err := runWorkerScenario(rootDir, language, sdkVersion); err != nil {
```
bergundy: Neither does this...

cmd/dev/build.go
bergundy, line 1: I find this command confusing, it either generates protos, or delegates to the main command to build a docker image plus do things like run a scenario and build the omes CLI, which is not what I would expect.

cmd/dev/install.go
```diff
@@ -0,0 +1,202 @@
...
+func installUv(versions map[string]string) error {
+	version := versions["UV_VERSION"]
+	if version == "" {
+		return fmt.Errorf("version not found for uv")
+	}
+	return installViaMise("uv", version)
+}
+
+func installNpm(versions map[string]string) error {
+	// npm comes with node, so install node
```
bergundy: Then why do we need this alias?

cmd/dev/install.go
```diff
@@ -0,0 +1,202 @@
...
+
+	fmt.Printf("Installing %s...\n", strings.Join(tools, ", "))
+
+	for _, tool := range tools {
+		var err error
+		switch tool {
+		case "dotnet":
+			err = installDotnet(versions)
+		case "go":
+			err = installGo(versions)
```
bergundy: Do we need a separate Go version than what this tool is already running with?

cmd/dev/install.go
```diff
@@ -0,0 +1,202 @@
...
+			err = installPython(versions)
+		case "node":
+			err = installNode(versions)
+		case "npm":
+			err = installNpm(versions)
+		case "rust":
+			err = installRust(versions)
+		case "protoc":
+			err = installProtoc(versions)
+		case "uv":
```
bergundy: Should this just be part of installing Python?

cmd/dev/lint.go
```diff
@@ -0,0 +1,194 @@
...
+	case "typescript":
+		return lintTypescriptWorker(workerDir)
+	case "dotnet":
+		return lintDotnetWorker(workerDir)
+	default:
+		return fmt.Errorf("unsupported language: %s", language)
+	}
+}
+
+func lintGoWorker(workerDir string) error {
```
bergundy: This isn't just linting

cmd/dev/lint.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	fmt.Printf("uv version: ")
+	if err := runCommand("uv", "--version"); err != nil {
+		return err
+	}
+
+	fmt.Println("Checking Python dependencies...")
+	if err := runCommandInDir(workerDir, "uv", "sync"); err != nil {
```
bergundy: This isn't part of linting

cmd/dev/lint.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	fmt.Println("Applying Python format...")
+	if err := runCommandInDir(workerDir, "poe", "format"); err != nil {
```
bergundy: Is this just checking but also auto-formatting? There should be a distinction between the two.

cmd/dev/lint.go
```diff
@@ -0,0 +1,194 @@
...
+	return nil
+}
+
+func lintTypescriptWorker(workerDir string) error {
+	if err := validateLanguageTools("typescript"); err != nil {
+		return err
+	}
+
+	fmt.Println("Installing TypeScript dependencies...")
+	if err := runCommandInDir(workerDir, "npm", "ci"); err != nil {
```
bergundy: Not part of linting.

cmd/dev/lint.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	fmt.Println("Running .NET formatting check...")
+	if err := runCommandInDir(workerDir, "dotnet", "format", "--verify-no-changes"); err != nil {
+		return err
+	}
+
+	fmt.Println("Running .NET tests...")
+	if err := runCommandInDir(workerDir, "dotnet", "test"); err != nil {
```
bergundy: Not part of linting.

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+
+var supportedLanguages = []string{
+	"dotnet", "go", "java", "python", "typescript",
+}
+
+var supportedTools = []string{
+	"dotnet", "go", "java", "node", "npm", "protoc", "python", "rust",
+}
+
+var versionVarNames = map[string]string{
```
bergundy: Do we need these env vars? Why not just take command line options?

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+// checkCommand verifies that a command is available in PATH
+func checkCommand(cmd string) error {
+	if _, err := getCommandPath(cmd); err != nil {
+		return fmt.Errorf("%s is not installed. Please install %s first", cmd, cmd)
+	}
+	return nil
+}
+
+// getCommandPath returns the full path to a command
+func getCommandPath(cmd string) (string, error) {
```
bergundy: Redundant helper.

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+	return exec.LookPath(cmd)
+}
+
+// runCommand executes a command and pipes output to stdout/stderr
+func runCommand(name string, args ...string) error {
+	cmd := exec.Command(name, args...)
```
bergundy: nit: use CommandContext everywhere for better hygiene.

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return string(output), nil
+}
+
+// getRootDir returns the project root directory (current working directory when running go run ./cmd/dev)
+func getRootDir() (string, error) {
```
bergundy: Does this mean that this script only works if it runs from the repository root? It should probably verify that then somehow.

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+	return os.Getwd()
+}
+
+// loadVersions parses versions.env file and returns a map of version variables
+func loadVersions() (map[string]string, error) {
```
bergundy, line 114: Can you explain why we need `versions.env`?

cmd/dev/util.go
```diff
@@ -0,0 +1,225 @@
...
+	// Find versions.env file by walking up the directory tree
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	for dir != "/" {
```
bergundy: Does this work in windows?

loadgen/kitchen-sink-gen/build.rs
bergundy: Fairly certain that cargo build will rerun if the Rust version changes already.

stephanos: Right; but not if the protoc version changes. (which is defined in versions.env)

cmd/dev/lint.go
stephanos: Happy to discuss a different term; I don't think "linting" has to literally mean "check formatting". I've taken it here to mean as "is this code/project correct" which is different from "building" in that it's much faster (doesn't build Docker) and finds other kinds of issues.

cmd/dev/lint.go
stephanos: I got it the first time ...

cmd/dev/util.go
stephanos: To have a canonical source of all the versions across the entire project; ie source of truth. Granted, now that this is ported to Go, there's only one more use in CI: https://github.com/temporalio/omes/blob/main/.github/workflows/ci.yml#L233 - but still something I'd rather not duplicate.

stephanos: We can discuss an approach to move everything into Go; but I'm not quite sure how to re-use that in ci.yml.

At the same time, I like having the versions so clearly discoverable and not embedded/hidden in the Go script helper, too.

cmd/dev/lint.go
stephanos: "check"? I've renamed it to check now.

cmd/dev/util.go
stephanos: This map maps Go vars/identifiers to the entries in `versions.env`.

cmd/dev/install.go
stephanos: I suppose since the Go script shares the same go.mod it should use the same Go that Omes needs. However; the Omes Go worker has its own go.mod, so that _could_ differ. But I believe Go auto-installs the right toolchain these days. So it seems safe to remove.

cmd/dev/install.go
stephanos: It is actually already installed earlier, I'll make this a noop.

cmd/dev/build.go
stephanos: So the genesis of this are the CI steps `build-lint-test-go` etc. [[link](https://github.com/temporalio/omes/pull/181/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL10)] which performed exactly the steps that are now codified as scripts.

stephanos: A meaningful change was that I merged the kitchensink build into this command since it seemed odd to have a separate one, but I can revert that.

And that the "test" was dropped form the command name. I could add a separate "test" command.

cmd/dev/util.go
stephanos: I've considered a few options here and they all seem to come with tradeoffs:

1. `os.Getwd()` (requires to run from project root)
2. `runtime.Caller(0)` to find Go source file and thereby the root (only works via `go run`)
3. iterate through parents until a certain marker file is found

I don't want to over-engineer this; (2) seems like a reasonable tradeoff to me?

cmd/dev/util.go
stephanos: Replaced this now by relying on `getRootDir`

cmd/dev/util.go
stephanos: neat, didn't know they added that

cmd/dev/build.go
stephanos: Hm, I see now that the Go script tests the Go worker directly, not the image. That was what the original script (and CI) was doing.

cmd/dev/util.go
stephanos: I've actually removed it now and calculate the key programatically.

cmd/dev/build.go
stephanos: Added a separate test command now 👍

bergundy: Wondering if we even need a build command since the main command already does that for us. Maybe just add a build-proto dev tool instead?

cmd/dev/check.go
```diff
@@ -0,0 +1,195 @@
...
+	if err := runCommandInDir(ctx, workerDir, "go", "fmt", "./..."); err != nil {
+		return err
+	}
+
+	fmt.Println("Running Go tests...")
+	if err := runCommandInDir(ctx, workerDir, "go", "test", "-race", "./..."); err != nil {
+		return err
+	}
+
+	fmt.Println("Building Go worker...")
```
bergundy: Why is this here? Is build supposed to be one of the checks?

cmd/dev/check.go
```diff
@@ -0,0 +1,195 @@
...
+	if err := checkTool(ctx, "python"); err != nil {
+		return err
+	}
+
+	fmt.Print("uv version: ")
+	if err := runCommand(ctx, "uv", "--version"); err != nil {
+		return err
+	}
+
+	fmt.Println("Checking Python dependencies...")
```
bergundy: This is just installing, not checking...

cmd/dev/check.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+
+func checkGoWorker(ctx context.Context, workerDir string) error {
+	if err := checkTool(ctx, "go"); err != nil {
+		return err
+	}
+
+	fmt.Println("Checking Go modules...")
+	if err := runCommandInDir(ctx, workerDir, "go", "mod", "tidy"); err != nil {
```
bergundy: This isn't checking anything.

cmd/dev/lint_and_format.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	fmt.Println("Checking Go modules...")
+	if err := runCommandInDir(ctx, workerDir, "go", "mod", "tidy"); err != nil {
+		return err
+	}
+
+	fmt.Println("Applying Go format...")
+	if err := runCommandInDir(ctx, workerDir, "go", "fmt", "./..."); err != nil {
```
bergundy, line 86: I haven't seen `go fmt` run in CI typically, this isn't really a check command though unless we check that no files were changed but that seems redundant.

cmd/dev/lint_and_format.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	fmt.Println("Checking Python dependencies...")
+	if err := runCommandInDir(ctx, workerDir, "uv", "sync"); err != nil {
+		return err
+	}
+
+	fmt.Println("Applying Python format...")
+	if err := runCommandInDir(ctx, workerDir, "poe", "format"); err != nil {
```
bergundy, line 123: Same here, this already does the formatting.

cmd/dev/check.go
```diff
@@ -0,0 +1,195 @@
...
+	fmt.Println("✅ Python check completed successfully!")
+	return nil
+}
+
+func checkTypescriptWorker(ctx context.Context, workerDir string) error {
+	if err := checkTool(ctx, "typescript"); err != nil {
+		return err
+	}
+
+	fmt.Println("Installing TypeScript dependencies...")
```
bergundy: Same here, this does the installation. Can we separate installation from checks?

cmd/dev/check.go
```diff
@@ -0,0 +1,195 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func checkCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "check [language...]",
+		Short: "Check worker for specified language(s)",
+		Long: fmt.Sprintf(`Check worker for specified language(s)
```
bergundy: Maybe explain exactly what the definition of check is?
Should we split this into format (only to be run by dev), check-format to ensure things are submitted formatted in CI, lint, and test?

Also installing the dependencies should be in a separate command.

cmd/dev/clean.go
```diff
@@ -0,0 +1,122 @@
...
+			return runClean(cmd.Context(), languages, cleanAll)
+		},
+	}
+
+	return cmd
+}
+
+func cleanLanguage(ctx context.Context, language, rootDir string) error {
+	workerDir := filepath.Join(rootDir, "workers", language)
+	if _, err := os.Stat(workerDir); os.IsNotExist(err) {
```
bergundy: We should error out on any other error.

cmd/dev/util.go
bergundy: I would be okay with 1 to allow this to work with a built binary but maybe just have some sanity check that the script is run in the proper directory? E.g. it has a `go.mod` containing `module github.com/temporalio/omes` on the first line?

bergundy: I would consider this a nice to have.

cmd/dev/util.go
bergundy: Sounds good to me.

cmd/dev/check.go
stephanos: I've moved the installation steps into `install` now.

cmd/dev/lint_and_format.go
stephanos: Yes, the CI has this step:
```
git diff --exit-code || (echo "${{ matrix.sdk }} worker has uncommitted formatting changes" && exit 1)
```

Why would that be redundant?

stephanos: Also, I was using the "check" (formerly "lint") scripts to quickly fix any formatting when developing locally. It just seems more complicated to introduce yet another command or flag for that.

cmd/dev/check.go
stephanos: I've moved the dependency installation step into `install` now.

cmd/dev/check.go
stephanos: It _checks_ that the Go source code is valid; ie has no compile errors. Seems reasonable?

cmd/dev/build.go
stephanos: It's prob true that most of the time building the image locally isn't necessary for dev purposes; I can move that step fully to the CI only.

cmd/dev/check.go
stephanos: I'd like to keep the list of commands small. As a user of these scripts, it's very valuable to just be able to run a single command to quickly verify that everything is working and automatically adjust minor things like formatting. It's very convenient and I don't think splitting it further adds any meaningful benefits.

It's also great for AI-driven development since you can just say "run go run ./cmd/dev check to verify your work" and it'll do the right thing.

cmd/dev/build.go
stephanos: Okay, I've removed the image building and testing from the dev scripts and made them CI-only now.

cmd/dev/util.go
stephanos: So apparently `features` does the same as (2): https://github.com/temporalio/features/blob/main/cmd/run.go#L561C1-L564C2

So might as well be consistent?

.github/workflows/ci.yml
```diff
@@ -46,92 +34,97 @@ jobs:
...
-        with:
-          version: ${{ env.MISE_VERSION }}
-      - name: Install Tools
-        run: ${{ matrix.install }}
-      - name: Build ${{ matrix.sdk }} worker
-        run: ${{ matrix.build }}
+          submodules: true
+      - name: Load versions
+        run: |
+          set -a && source versions.env && set +a
```
bergundy: Can we just load this file in the Go scripts? Don't think this needs to be in env var form. Just yaml or even JSON would do.

.github/workflows/ci.yml
```diff
@@ -46,92 +34,97 @@ jobs:
...
+      - name: Set SDK version variable
+        run: |
+          SDK_VAR_NAME="${{ matrix.sdk }}_SDK_VERSION"
+          SDK_VAR_NAME=$(echo "$SDK_VAR_NAME" | tr '[:lower:]' '[:upper:]')
+          echo "SDK_VERSION=${!SDK_VAR_NAME}" >> $GITHUB_ENV
+      - name: Build ${{ matrix.sdk }} worker image
+        run: |
+          go run ./cmd build-worker-image \
+            --language ${{ matrix.sdk }} \
+            --version v${{ env.SDK_VERSION }} \
```
bergundy: For example, if `--version` isn't given, we can default to loading the version from the file, or we can take `--versions-file` as an argument.

.github/workflows/ci.yml
bergundy: Not blocking the PR, I just think it's a little cleaner to contain everything in the Go scripts and do less bash scripting in CI. It's easier for running everything locally.

.github/workflows/ci.yml
```diff
@@ -195,13 +184,13 @@ jobs:
           version: ${{ env.MISE_VERSION }}
       - name: Install Tools
         run: |
-          ./scripts/install-rust.sh
-          ./scripts/install-go.sh
-          ./scripts/install-protoc.sh
-          ./scripts/install-node.sh
+          go run ./cmd/dev install rust go protoc node
```
bergundy, line 177: Wondering if we actually need node here.

cmd/dev/build_proto.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+
+	fmt.Println("Generating TypeScript protos...")
+	typescriptWorkerDir := filepath.Join(rootDir, "workers", "typescript")
+	if err := runCommandInDir(ctx, typescriptWorkerDir, "npm", "install"); err != nil {
+		return err
+	}
+
+	if err := runCommandInDir(ctx, typescriptWorkerDir, "npm", "run", "proto-gen"); err != nil {
```
bergundy, line 57: Wouldn't this be part of the TS setup?

cmd/dev/lint_and_format.go
bergundy: Let's call this `lint-and-format` or something and make sure that we also format the TS code?

cmd/dev/clean.go
bergundy: Just pinging to make sure this is going to be fixed before merging.

cmd/dev/util.go
bergundy: Leaving it up to you.

bergundy (approved):
Left some more comments but this is already much better than before. Feel free to merge when you feel comfortable doing so.

.github/workflows/ci.yml
stephanos: We do, unfortunately, as we are running `npm run proto-gen` as part of the build.

cmd/dev/build_proto.go
stephanos: Well, I'd expect a user to run `build-proto` when they changed the proto files to generate the new kitchen-sink-gen _and_ generate the new TS proto files at the same time (that's what the CI was doing before, too). So I don't think we want to run it as part of `install`, that seems surprising. That would only leave `check`, and that might not be ideal either. So I think it has to be here.

cmd/dev/lint_and_format.go
stephanos: 👍 added the TS formatting

stephanos: IDK about `lint-and-format` or anything else that's a compound phrase since it's compiling as well ... might be even run worker tests in the future, if we have any.

(1) How about `check --fix` to signal that it would address any issues (like formatting) where applicable?

(2) Or we could have `check` (no formatting) and `fix` (both formatting and check)?

(3) Or just `tidy`? (can be read as a verb and adjective) 

Having said that, I don't see the use case for just `check` _without formatting_. A developer would always want both (why manually fix things?) and CI is also running both.

.github/workflows/ci.yml
stephanos: Hm, I was going back to doing it this way since you had objected to the `build` command as it was just a thin wrapper around `build-worker-image` with only the version resolution being added.

> Wondering if we even need a build command since the main command already does that for us. Maybe just add a build-proto dev tool instead?

I can revert it back to that. I also liked that more.

stephanos: I see your next comment suggests essentially merging the previous `build` command into the existing Omes `build-worker-image`. That sounds okay to me if we move it into the `dev` command since it doesn't really belong to the Omes CLI imo. Same for `prepare-worker`, actually. Then those two would just be dev commands.

PR #205, opened 2025-09-17
signals(dedup): deduplicate signals and ensure all signals received

### LINKED ISSUE #204: [Feature Request] Count and de-duplicate signals in kitchensink
### Is your feature request related to a problem? Please describe.

1. kitchensink workflows (all SDKs) need to deduplicate signals
2. kitchensink workflows (all SDKs) need to verify that they have received the expected number of _unique signals_ before they return a result

> Note: this could be done starting with just Go

**why**
1. to ensure (2) is accurate and that no actions from signals are executed more than once
2. to verify that all expected signals have been processed and no signal was lost

### Describe the solution you'd like

* Pass the expected number of signals as a workflow input (ie. 3)
* Create a set of signal IDs (ie. [1, 2, 3])
* assign a number to each signal (example: 2) and remove the signal's ID from the set of expected signal IDs (example: [1, 3]).
* Only execute the signal's actions if the ID was present in the set
* Verify the set is empty before returning OR return an error with missing signal IDs

### Additional context
NA

spkane31:
## What was changed
Add # of signals as a workflow input

## Why?
To ensure expected number of unique signals is accurate and no action from signals that are executed more than once.

## Checklist

1. Closes #204 

2. How was this tested: added unit tests

3. Any docs updates needed?

### DIFF
--- a/loadgen/kitchen_sink_executor_test.go
+++ b/loadgen/kitchen_sink_executor_test.go
@@ -648,6 +648,9 @@ func TestKitchenSink(t *testing.T) {
 				},
 			},
 			expectedWorkflowError: "error after signals sent",
+			expectedUnsupportedErrs: map[clioptions.Language]string{
+				clioptions.LangJava: "client actions activity is not supported",
+			},
 			historyMatcher: PartialHistoryMatcher(`
 				WorkflowExecutionStarted
 				WorkflowTaskScheduled


workers/go/kitchensink/kitchen_sink.go
```diff
@@ -39,13 +39,28 @@ func (ca *ClientActivities) ExecuteClientActivity(ctx context.Context, clientAct
...
 
 func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (*common.Payload, error) {
 	workflow.GetLogger(ctx).Debug("Started kitchen sink workflow")
 
 	state := KSWorkflowState{
-		workflowState: &kitchensink.WorkflowState{},
+		workflowState:       &kitchensink.WorkflowState{},
+		expectedSignalCount: 0,
+		expectedSignalIDs:   make(map[int32]interface{}),
+		receivedSignalIDs:   make(map[int32]interface{}),
```
stephanos: ~Would it be reasonable/possible to only use one map?~ Ah, I see the check for "not expected" now.

spkane31: Done

workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
```diff
@@ -28,8 +28,28 @@ public class KitchenSinkWorkflowImpl implements KitchenSinkWorkflow {
   KitchenSink.WorkflowState state = KitchenSink.WorkflowState.getDefaultInstance();
   WorkflowQueue<KitchenSink.ActionSet> signalActionQueue = Workflow.newWorkflowQueue(1_000);
 
+  // signal de-duplication fields
+  int expectedSignalCount = 0;
+  Map<Integer, Object> expectedSignalIds = new HashMap<>();
+  Map<Integer, Object> receivedSignalIds = new HashMap<>();
```
bergundy: Java has sets that can be used for this.

workers/python/kitchen_sink.py
```diff
@@ -34,12 +34,66 @@ class KitchenSinkWorkflow:
     action_set_queue: asyncio.Queue[ActionSet] = asyncio.Queue()
     workflow_state = WorkflowState()
 
+    # signal de-duplication fields
+    expected_signal_count: int = 0
+    expected_signal_ids: set[int] = set()
```
bergundy: Why do you need both the expected signal IDs and the count?

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -206,15 +211,89 @@ export async function kitchenSink(input: WorkflowInput | undefined): Promise<IPa
...
 
-  setHandler(reportStateQuery, (_) => workflowState);
-  setHandler(actionsSignal, async (actions) => {
-    if (actions.doActionsInMain) {
-      actionsQueue.unshift(actions.doActionsInMain);
-    } else if (actions.doActions) {
-      await handleActionSet(actions.doActions);
+  async function handleSignal(actions: DoSignalActions): Promise<void> {
+    const receivedId = actions.signalId;
+    if (receivedId !== 0) {
```
bergundy: Prefer early return, and also can probably just assume that the receivedId would be valid.

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -206,15 +211,89 @@ export async function kitchenSink(input: WorkflowInput | undefined): Promise<IPa
...
+      // Handle signal with ID for deduplication
+      if (!expectedSignalIds.has(receivedId)) {
```
bergundy: If the expected signals are a continuous block from `1..N` can we use a range check instead?

workers/typescript/src/workflows/kitchen_sink.ts
```diff
@@ -50,6 +50,11 @@ export async function kitchenSink(input: WorkflowInput | undefined): Promise<IPa
   let workflowState: IWorkflowState = WorkflowState.create();
   const actionsQueue = new Array<IActionSet>();
 
+  // signal de-duplication fields
+  let expectedSignalCount = 0;
+  const expectedSignalIds = new Set<number>();
+  const receivedSignalIds = new Set<number>();
```
bergundy: How are you transferring these on continue-as-new?

workers/typescript/src/workflows/kitchen_sink.ts
spkane31: I did not do a continuous block, the signals (`1..N`) are randomly shuffled.

workers/typescript/src/workflows/kitchen_sink.ts
stephanos: Thinking out out loud: what if we always fed the `workflowState` into the new workflow? The CAN action has "arguments" that are propagated right now.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -237,6 +269,28 @@ func (ws *KSWorkflowState) handleAction(
...
+	return err
+}
+
+func (ws *KSWorkflowState) validateSignalCompletion(ctx workflow.Context) error {
+	if len(ws.expectedSignalIDs) != int(ws.expectedSignalCount) {
+		missing := []int32{}
+		for id := range ws.expectedSignalIDs {
+			missing = append(missing, id)
+		}
+		return fmt.Errorf("expected %d signals, got %d, missing %v", ws.expectedSignalCount, len(ws.expectedSignalIDs), missing)
```
stephanos: Why return an error if it's never printed? Could just use a bool. That would also simplify the `if` above.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,12 +101,31 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
+				if err != nil {
+					workflow.GetLogger(ctx).Error("error handling signal", "error", err)
+					retOrErrChan.Send(ctx, ReturnOrErr{nil, err})
+					return
 				}
-			})
+
+				// Check if all expected signals have been received
+				if state.expectedSignalCount > 0 {
+					if err := state.validateSignalCompletion(ctx); err != nil {
```
stephanos: Shouldn't this be `err == nil`?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,12 +101,31 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
+				err := state.handleSignal(ctx, receivedID, actionSet)
+				if err != nil {
+					workflow.GetLogger(ctx).Error("error handling signal", "error", err)
+					retOrErrChan.Send(ctx, ReturnOrErr{nil, err})
+					return
 				}
-			})
+
+				// Check if all expected signals have been received
+				if state.expectedSignalCount > 0 {
```
stephanos: Could move this `if` check into `validateSignalCompletion` to simplify.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,12 +101,31 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
 				actionSet = sigActions.GetDoActions()
 			}
-			workflow.Go(ctx, func(ctx workflow.Context) {
-				ret, err := state.handleActionSet(ctx, actionSet)
-				if ret != nil || err != nil {
-					retOrErrChan.Send(ctx, ReturnOrErr{ret, err})
+
+			receivedID := sigActions.GetSignalId()
+			if receivedID != 0 {
+				err := state.handleSignal(ctx, receivedID, actionSet)
```
stephanos: It looks like the `else` block is handling the signal inside a `Go`routine. I don't see one on this path. That might have to be addressed.

Generally, is there a world where we don't have branching? I'm thinking the `workflow.Go(` should be at the bottom of a successful list of `if`s that check for duplication etc.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +511,56 @@ func TestKitchenSink(t *testing.T) {
 			},
 			historyMatcher: PartialHistoryMatcher(`WorkflowExecutionSignaled`),
 		},
+		{
+			name: "ClientSequence/Signal/Deduplication",
+			testInput: &TestInput{
+				WorkflowInput: &WorkflowInput{
+					ExpectedSignalCount: 10,
+					InitialActions: ListActionSet(
+						NewAwaitWorkflowStateAction("signals_complete", "true"),
```
stephanos: This might be too brittle since `NewSetWorkflowStateAction` can override that entire state if I'm not mistaken. It would be better to safe-guard that special key somehow.

stephanos: Related to that, I've been wondering if `SetWorkflowStateAction` overriding the entire map is helpful. If it only updated a single key, that might be more useful.

Or we add a new `MergeWorkflowStateAction` that ... well merges selectively. But that would still be vulnerable to overrides from `SetWorkflowStateAction`.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -127,6 +130,13 @@ message WorkflowState {
 
 message WorkflowInput {
   repeated ActionSet initial_actions = 1;
+
+  // Number of signals the client will send to the workflow
+  int32 expected_signal_count = 2;
+
+  // Signal de-duplication state (used when continuing as new)
+  repeated int32 expected_signal_ids = 3;
```
bergundy, line 138: We should just use a continuous block from 1..N and remove this complexity IMHO.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -39,13 +39,26 @@ func (ca *ClientActivities) ExecuteClientActivity(ctx context.Context, clientAct
 
 type KSWorkflowState struct {
 	workflowState *kitchensink.WorkflowState
+
+	// signal de-duplication fields
+	expectedSignalCount int32
+	expectedSignalIDs   map[int32]interface{}
```
bergundy: Map to empty struct is the idiomatic way to represent sets in Go.

```suggestion
	expectedSignalIDs   map[int32]struct{}
```

loadgen/kitchensink/helpers.go
```diff
@@ -186,6 +187,35 @@ func NewAwaitWorkflowStateAction(key, value string) *Action {
...
+								),
+							},
+						},
+					},
+				},
+			},
+		}
+	}
+
+	// randomize the order of the actions
```
bergundy: Why does this matter?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,10 +101,31 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
 			workflow.Go(ctx, func(ctx workflow.Context) {
 				ret, err := state.handleActionSet(ctx, actionSet)
 				if ret != nil || err != nil {
 					retOrErrChan.Send(ctx, ReturnOrErr{ret, err})
+					return
+				}
+
+				// Check if all expected signals have been received (only for signals with IDs)
+				if receivedID != 0 && state.validateSignalCompletion() {
+					state.workflowState.Kvs["signals_complete"] = "true"
```
bergundy: Do you even need this in the workflow state kvs?
The workflow should just fail if it did not receive all of the expected signals.

workers/go/kitchensink/kitchen_sink.go
bergundy, line 1: Still don't understand how the list of expected signals would be carried over the continue as new boundary.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +511,62 @@ func TestKitchenSink(t *testing.T) {
...
+					),
+				},
+				ClientSequence: &ClientSequence{
+					ActionSets: []*ClientActionSet{
+						{
+							Actions: NewSignalActionsWithIDs(1, 2, 3, 4, 5, 6, 7, 8, 9, 10),
```
bergundy, line 527: Probably best to just give a range here instead of having to provide each individual signal ID.

loadgen/kitchensink/helpers.go
spkane31: Removing

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -436,11 +444,17 @@ impl<'a> Arbitrary<'a> for DoSignal {
             // Half of that in the handler half in main
             if u.ratio(50, 100)? {
                 do_signal::Variant::DoSignalActions(
-                    Some(do_signal_actions::Variant::DoActions(u.arbitrary()?)).into(),
+                    DoSignalActions {
+                        signal_id: u.arbitrary()?,
```
stephanos: So some signals use `u.arbitrary()` and some use `0` for the `signal_id` (for client actions). I can't quite tell why there are two strategies. Could we elaborate in a comment why one or the other is chosen; or use a single one? Since `expected_signal_ids` is empty, maybe this should also be `0`.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +511,68 @@ func TestKitchenSink(t *testing.T) {
 			},
 			historyMatcher: PartialHistoryMatcher(`WorkflowExecutionSignaled`),
 		},
+		{
+			name: "ClientSequence/Signal/Deduplication",
+			testInput: &TestInput{
+				WorkflowInput: &WorkflowInput{
+					ExpectedSignalCount: 10,
```
stephanos: nit: could just use 3 or 5 to keep it simpler. I was counting the signals even in the history to check it was really 10.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +511,68 @@ func TestKitchenSink(t *testing.T) {
...
+				WorkflowInput: &WorkflowInput{
+					ExpectedSignalCount: 3,
+					InitialActions: ListActionSet(
+						NewTimerAction(1000),
+					),
+				},
+				ClientSequence: &ClientSequence{
+					ActionSets: []*ClientActionSet{
+						{
+							Actions: NewSignalActionsWithIDs(2), // Only send 2 signals, expect 3 and WF should fail
```
stephanos: Maybe I did it wrong; but when I ran this test, I didn't see any indication of a failure (looked for logs). It'd be better if the test actually checked the expected error. The test harness doesn't support that yet; but it's hopefully an easy fix.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,10 +118,29 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
+					retOrErrChan.Send(ctx, ReturnOrErr{nil, err})
+					return
+				}
+			}
+
 			workflow.Go(ctx, func(ctx workflow.Context) {
 				ret, err := state.handleActionSet(ctx, actionSet)
 				if ret != nil || err != nil {
 					retOrErrChan.Send(ctx, ReturnOrErr{ret, err})
+					return
```
stephanos: nit: is this needed?

loadgen/kitchen_sink_executor_test.go
stephanos: On closer inspection, I actually don't see a check for `expectedSignalCount == actualSignalCount` anywhere.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -158,8 +207,9 @@ func (ws *KSWorkflowState) handleAction(
 	} else if re := action.GetReturnError(); re != nil {
 		return nil, temporal.NewApplicationError(re.Failure.Message, "")
 	} else if can := action.GetContinueAsNew(); can != nil {
-		// Use string arg to avoid the SDK trying to convert payload to input type
-		return nil, workflow.NewContinueAsNewError(ctx, "kitchenSink", can.GetArguments()[0])
+		// Create new workflow input preserving signal deduplication state
+		newInput := ws.createContinueAsNewInput(can.GetArguments()[0])
+		return nil, workflow.NewContinueAsNewError(ctx, "kitchenSink", newInput)
```
stephanos: Okay; I've thought about this for a while and I think it'd be easier and better for debugging if each workflow run checks its own signal expectations and that doesn't move across run boundaries. Let me explain:

- workflow input is just `expectedSignalCount`
- `expectedSignalIDs` is always generated from that count
- signal handler behaves like now; doing the book keeping and erroring on unexpected IDs
- _before it can return/CAN_, it needs to wait for _all_ signals to arrive (need to think about timeout condition here)
- once that happened, it can then return/CAN

This removes a lot of the complexity, I believe, and makes debugging easier since the missing signal will be in the run that's stuck/timed out. It makes action generation easier for the client, too, since it only has to look at single runs. WDYT?

loadgen/kitchen-sink-gen/src/main.rs
stephanos: (also, aren't the signal IDs meant to be consecutive numbers?)

spkane31: Fixed!

loadgen/kitchen_sink_executor_test.go
spkane31: Reducing to 5 and adding the check

workers/go/kitchensink/kitchen_sink.go
spkane31: Removed in two places

workers/go/kitchensink/kitchen_sink.go
spkane31: I'm not sure how much easier this is to debug and check signal count. I can make these changes but do we want to hold up the CAN boundary on waiting for all signals?

spkane31: Discussed on 10/21 and want to pursue the path where "before workflow completes w/out an error, check and wait for signals to be received".

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -237,6 +343,63 @@ func (ws *KSWorkflowState) handleAction(
...
+}
+
+// isSignalAlreadyReceived checks if a signal has already been received (for deduplication)
+func (ws *KSWorkflowState) isSignalAlreadyReceived(signalID int32) bool {
+	return slices.Contains(ws.receivedSignalIDs, signalID)
+}
+
+// createContinueAsNewInput creates a new WorkflowInput for continue-as-new that preserves
+// signal deduplication state by tracking received and expected signal IDs
+func (ws *KSWorkflowState) createContinueAsNewInput(originalArg *common.Payload) *common.Payload {
```
stephanos: This can be deleted now, right?

(same for some of the `(for continue-as-new deduplication)` comments)

spkane31: Removed and cleaned up functions

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -38,14 +38,30 @@ func (ca *ClientActivities) ExecuteClientActivity(ctx context.Context, clientAct
...
 
 type KSWorkflowState struct {
-	workflowState *kitchensink.WorkflowState
+	workflowState     *kitchensink.WorkflowState
+	expectedSignalIDs map[int32]struct{}
 }
 
 func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (*common.Payload, error) {
-	workflow.GetLogger(ctx).Debug("Started kitchen sink workflow")
+	workflow.GetLogger(ctx).Info("Started kitchen sink workflow")
```
stephanos: Could/should we revert this to debug again? Could be fairly noisy at scale.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -38,14 +38,30 @@ func (ca *ClientActivities) ExecuteClientActivity(ctx context.Context, clientAct
...
+		workflowState:     &kitchensink.WorkflowState{},
+		expectedSignalIDs: make(map[int32]struct{}),
+	}
+
+	if params != nil {
+		if len(params.ExpectedSignalIds) > 0 {
+			// Use the explicit expected signal IDs
+			for _, id := range params.ExpectedSignalIds {
+				state.expectedSignalIDs[id] = struct{}{}
+			}
```
stephanos: Since we're not passing them across CAN boundary, can we delete this now?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -54,17 +70,18 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 			return state.workflowState, nil
 		})
 	if queryErr != nil {
+		workflow.GetLogger(ctx).Info("Error setting query handler", queryErr)
```
stephanos: Do we need this? Wouldn't we get the queryErr as a workflow result?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -74,6 +91,7 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 			},
 		})
 	if updateErr != nil {
+		workflow.GetLogger(ctx).Info("Error setting update handler", updateErr)
```
stephanos: Same here as for queryErr.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,6 +106,18 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 			if actionSet == nil {
 				actionSet = sigActions.GetDoActions()
 			}
+
+			receivedID := sigActions.GetSignalId()
+
+			// Handle signal deduplication if signal ID is provided
+			if receivedID != 0 {
```
stephanos:
```suggestion
			// Handle signal deduplication if signal ID is provided
			if receivedID := sigActions.GetSignalId(); receivedID != 0 {
```

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -88,6 +106,18 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
+				if state.isSignalAlreadyReceived(receivedID) {
+					workflow.GetLogger(ctx).Debug("Signal already received, skipping", "signalID", receivedID)
+					continue
+				}
+				state.handleSignalDeduplication(receivedID)
```
stephanos: Let's inline that?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -98,21 +128,63 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
 
 	// Handle initial set.
+	// Store return value/error from initial actions instead of returning immediately
+	var initialRetOrErr *ReturnOrErr
 	if params != nil && params.InitialActions != nil {
 		for _, actionSet := range params.InitialActions {
 			if ret, err := state.handleActionSet(ctx, actionSet); ret != nil || err != nil {
-				workflow.GetLogger(ctx).Debug("Finishing early", "ret", ret, "err", err)
-				return ret, err
+				workflow.GetLogger(ctx).Info("Got return/error from initial actions", "ret", ret, "err", err, "actionSet", actionSet)
```
stephanos: Could we make that `Debug` level as well?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -98,21 +128,63 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
...
+				if err != nil {
+					return ret, err
+				}
+				// Otherwise, store the return value to use later
+				initialRetOrErr = &ReturnOrErr{ret, err}
+			}
+		}
+	}
+
+	workflow.GetLogger(ctx).Info("Entering main wait loop")
```
stephanos: Delete this?

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -98,21 +128,63 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 	})
 
 	// Handle initial set.
+	// Store return value/error from initial actions instead of returning immediately
```
stephanos: Can we delete this? It's a bit out of place; and not 100% accurate (we don't store the error).

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -237,6 +323,19 @@ func (ws *KSWorkflowState) handleAction(
 	return nil, nil
 }
 
+func (ws *KSWorkflowState) handleSignalDeduplication(signalID int32) {
+	// Simply remove the signal ID from the expected set
```
stephanos: The punctuation here and in other places is not quite consistent with the rest of the codebase. Usually if it starts with a capital letter it also needs a period at the end. Especially when there are two sentences.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -149,6 +221,21 @@ func (ws *KSWorkflowState) handleActionSet(
...
+func (ws *KSWorkflowState) validateAllSignalsReceived(ctx workflow.Context) []int32 {
+	if len(ws.expectedSignalIDs) > 0 {
+		missingSignals := make([]int32, 0, len(ws.expectedSignalIDs))
+		for signalID := range ws.expectedSignalIDs {
+			missingSignals = append(missingSignals, signalID)
+		}
+		workflow.GetLogger(ctx).Info("Missing expected signals", "missingSignalIDs", missingSignals)
+		return missingSignals
+	}
+	workflow.GetLogger(ctx).Info("All expected signals received")
```
stephanos: I'd rather remove this log line as AFAICT it would be logged even if no signal expectations were ever defined.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -149,6 +221,21 @@ func (ws *KSWorkflowState) handleActionSet(
 	return
 }
 
+// validateAllSignalsReceived checks if all expected signals have been received
+// and returns the list of missing signal IDs if any are missing
+func (ws *KSWorkflowState) validateAllSignalsReceived(ctx workflow.Context) []int32 {
```
stephanos: I would probably inline this function, too, honestly. Especially if you use `maps.Keys(m)` (from `golang.org/x/exp/maps`).

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -237,6 +323,19 @@ func (ws *KSWorkflowState) handleAction(
...
 }
 
+func (ws *KSWorkflowState) handleSignalDeduplication(signalID int32) {
+	// Simply remove the signal ID from the expected set
+	// If it's not there, that's fine - it was already processed or wasn't expected
+	delete(ws.expectedSignalIDs, signalID)
+}
+
+// isSignalAlreadyReceived checks if a signal has already been received (for deduplication)
+func (ws *KSWorkflowState) isSignalAlreadyReceived(signalID int32) bool {
```
stephanos: I'd inline this as well since it's a simple membership test.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -842,7 +907,26 @@ func testSupportedFeature(
...
+		for _, event := range historyEvents {
+			if event.EventType == enums.EVENT_TYPE_WORKFLOW_EXECUTION_FAILED {
+				hasWorkflowFailed = true
+				break
+			}
+		}
+		require.Truef(t, hasWorkflowFailed, "SDK %s workflow should have WorkflowExecutionFailed event in history", sdk)
+	} else {
+		require.NoError(t, execErr, "executor failed")
+	}
```
stephanos, line 1022: Neat!

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +512,70 @@ func TestKitchenSink(t *testing.T) {
 			},
 			historyMatcher: PartialHistoryMatcher(`WorkflowExecutionSignaled`),
 		},
+		{
+			name: "ClientSequence/Signal/Deduplication",
```
stephanos, line 519: Is this testing deduplication actually? It seems it's actually testing that the workflow waits for all signals to be received. I'd recommend emphasizing that more by removing `NewTimerAction(2000)` (so it would return immediately otherwise). Could also reduce the number of signals to 3 or sth as it's the same as 10 in terms of behavior.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -511,6 +512,70 @@ func TestKitchenSink(t *testing.T) {
...
+				WorkflowExecutionSignaled
+				WorkflowExecutionSignaled`),
+			expectedWorkflowError: "missing signal",
+			expectedUnsupportedErrs: map[clioptions.Language]string{
+				clioptions.LangJava:       "signal deduplication not implemented",
+				clioptions.LangPython:     "signal deduplication not implemented",
+				clioptions.LangTypeScript: "signal deduplication not implemented",
+				clioptions.LangDotNet:     "signal deduplication not implemented",
+			},
+		},
```
stephanos, line 578: It seems to me we want 5 test cases:

1. workflow waits for all expected signals (✔️)
2. missing signal fails workflow (✔️)
3. sending a signal with the same ID twice is fine
4. returning an error from main loop skips signal waiting
5. returning an error from signal skips signal waiting

workers/go/kitchensink/kitchen_sink.go
spkane31: If we remove this and use the `params.ExpectedSignalIds`, then I think we should keep not inline the handle and validate signals functions because they are not oneline anymore. I'm going to go with this because it's unnecessary to keep this in the workflow state.

workers/go/kitchensink/kitchen_sink.go
spkane31: Removing the field on the workflow so keeping this inlined.

loadgen/kitchen_sink_executor_test.go
spkane31: Added the last three

loadgen/kitchen_sink_executor_test.go
spkane31: Removed

PR #214, opened 2025-09-22
Reusable CLI

stephanos:
## What was changed

Moved CLI commands to exported package:

- new package `cli`
- renamed `cmdoptions` to `clioptions` to match

## Why?

This allows writing a thin Omes CLI wrapper application that defines its own scenario; but still provide all the Omes CLI options as if it was an "internal" scenario.

Of course this only works if the scenario uses the kitchensink executor.

```go
package main
 
import (
    "github.com/temporalio/omes/cmd/cli"
     _ "examples/custom-scenario/scenarios" // register our custom scenarios
)

func main() {
    cli.Main()
}
```

Using Omes programmatically as a library was already possible; but the CLI use case was not.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,19 @@ For example, here is [scenarios/workflow_with_single_noop_activity.go](scenarios
 ```go
 func init() {
 	loadgen.MustRegisterScenario(loadgen.Scenario{
-		Description: "Each iteration executes a single workflow with a noop activity.",
-		Executor: loadgen.KitchenSinkExecutor{
-			WorkflowParams: kitchensink.NewWorkflowParams(kitchensink.NopActionExecuteActivity),
-		},
-	})
+        Description: "Each iteration executes a single workflow with a noop activity.",
+        ExecutorFn: func() loadgen.Executor {
+            return loadgen.KitchenSinkExecutor{
+                TestInput: &kitchensink.TestInput{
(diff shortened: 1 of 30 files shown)


cmd/cli/main.go
```diff
@@ -0,0 +1,27 @@
...
+	rootCmd.AddCommand(prepareWorkerCmd())
+	rootCmd.AddCommand(runScenarioCmd())
+	rootCmd.AddCommand(runScenarioWithWorkerCmd())
+	rootCmd.AddCommand(runWorkerCmd())
+
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
```
stephanos, line 28: This is the main change.

README.md
```diff
@@ -65,11 +65,19 @@ For example, here is [scenarios/workflow_with_single_noop_activity.go](scenarios
...
+                    WorkflowInput: &kitchensink.WorkflowInput{
+                        InitialActions: []*kitchensink.ActionSet{
+                            kitchensink.NoOpSingleActivityActionSet(),
+                        },
+                    },
+                },
+            }
+        },
+    })
 }
```
stephanos, line 81: Updated out-of-date README.

PR #226, opened 2025-10-08
Remove GHA var interpolation

stephanos:
WISOTT

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -2,10 +2,6 @@ name: Build docker images
 on:
   workflow_call:
     inputs:
-      build-command:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-images.yml
```diff
@@ -2,10 +2,6 @@ name: Build docker images
 on:
   workflow_call:
     inputs:
-      build-command:
-        description: Build command to run (e.g., 'build-worker-image' or 'build-cli-image')
-        type: string
-        default: "build-worker-image"
```
stephanos, line 8: Actually, this isn't used since this workflow only runs if `do-push` is `true`.

PR #227, opened 2025-10-09
Add option to ignore already started error

stephanos:
## What was changed

WISOTT

## Why?

When running a scenario it can happen that transient failures lead to an incomplete scenario run and on retry the scenario will then fail because workflows were already created. It's not always possible here to use a new/clean task queue; in those cases, this new flag allows to skip over these errors.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/cli/run_scenario.go
+++ b/cmd/cli/run_scenario.go
@@ -159,6 +159,7 @@ func (r *scenarioRunner) run(ctx context.Context) error {
 			MaxIterationAttempts:          r.maxIterationAttempts,
 			Timeout:                       r.timeout,
 			DoNotRegisterSearchAttributes: r.doNotRegisterSearchAttributes,
+			IgnoreAlreadyStarted:          r.ignoreAlreadyStarted,
 		},
 		ScenarioOptions: scenarioOptions,
 		Namespace:       r.clientOptions.Namespace,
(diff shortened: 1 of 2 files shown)


scenarios/throughput_stress.go
```diff
@@ -201,19 +200,6 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 		info.Logger.Debugf("Completed iteration %d", run.Iteration)
 	}
 
-	// When resuming, it can happen that the workflow for the current iteration already exists since the snapshot
-	// was not up-to-date. In that case, we just skip this iteration and move on.
-	info.Configuration.HandleExecuteError = func(ctx context.Context, run *loadgen.Run, err error) error {
```
stephanos, line 206: This was the only use case for `HandleExecuteError`; we could consider removing it entirely until there's a use case again?

cmd/cli/run_scenario.go
```diff
@@ -56,6 +56,7 @@ type scenarioRunConfig struct {
 	scenarioOptions               []string
 	timeout                       time.Duration
 	doNotRegisterSearchAttributes bool
+	ignoreAlreadyStarted          bool
```
Sushisource, line 59: Where does this actually get read?

stephanos: Yup, that part somehow got lost. Added!

Sushisource (approved):

PR #229, opened 2025-10-15
Added `include-retry-scenarios` option to `throughput_stress` scenario

THardy98:
## What was changed
Added `include-retry-scenarios` option to `throughput_stress` scenario.

Setting `--option include-retry-scenarios=true` adds activities that fail and retry, as actions to `throughput_stress`.

**NOTE**: enabling this option increases runtime of a single iteration from ~23s to ~44s!

New activity actions were added to for this fail-and-retry coverage:
- retryable activity
- timeout activity
- heartbeat timeout activity

The added proto definitions are the same. I've separated them because they cover different activity failure cases, which I think would be neater to handle separately in the workers rather than a single failure activity. It also allows them to be expanded separately and more specifically.

I've added these actions to the existing throughput stress scenario, as well as a cancellable activity action.

Added to corresponding logic to each worker to handle these new activity actions.

## Why?
Provide additional activity coverage, particularly in cases where the activity does not succeed on the first attempt.

2. How was this tested:
Ran the existing `throughput_stress_test.go` test.
Ran:
```
    --language go \
    --iterations 10 \
    --option internal-iterations=10 \
    --option include-retry-scenarios=true
```

Having trouble running throughput stress on other language workers (even before this change). Will address small fixes in subsequent PRs.

3. Any docs updates needed?
No


THardy98:
Was concerned about the runtime, so decided to put this behind an option `--include-retry-scenarios` for `throughput_stress`.

scenarios/throughput_stress.go
```diff
@@ -385,6 +401,20 @@ func (t *tpsExecutor) createActionsChunk(
...
+		if t.config.IncludeRetryScenarios {
+			asyncActions = append(asyncActions,
+				// Add activity cancellation scenario
+				DelayActivityWithCancellation(1*time.Second, 10*time.Second),
+				// Test activity retry: fails, succeeds on retry
+				RetryableErrorActivity(1, RemoteActivityWithRetry(1*time.Second, 2, 500*time.Millisecond, 1.0)),
+				// Test activity timeout with retry: times out on 1st attempt, succeeds on 2nd
+				TimeoutActivity(1, RemoteActivityWithRetry(1*time.Second, 2, 500*time.Millisecond, 1.0)),
+				// Test heartbeat timeout: skips heartbeats on 1st attempt, sends them on 2nd
+				HeartbeatActivity(1, RemoteActivityWithHeartbeat(10*time.Second, 1*time.Second, 2, 500*time.Millisecond, 1.0)),
```
Sushisource: Stylistic: If these helpers are only used here I'd probably just inline it. Having helpers for the DSL feels a little over-verbose unless the helper is doing a lot of heavy lifting or re-used a lot (which I believe is the case with the ones already present). That said I can see how it makes this nice too, so, 🤷 . Curious if @stephanos has an opinion.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -211,6 +211,30 @@ message ExecuteActivityAction {
...
+  // Tests activity retry behavior.
+  message RetryableErrorActivity {
+    // How many attempts should fail before succeeding (1-indexed)
+    int32 fail_attempts = 1;
+  }
+
+  // Activity that runs too long for N attempts (causing timeout), then completes.
+  // Tests StartToCloseTimeout behavior with retries.
+  // On failing attempts, the activity runs until context cancellation (timeout).
+  // On success, it runs for duration < StartToCloseTimeout (runs for 0.5x StartToClose timeout).
```
Sushisource:
```suggestion
  // On success, it runs for 0.5x StartToClose timeout.
```
If it always does that, just say that.

That said - I would probably just parameterize the runtime too.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -211,6 +211,30 @@ message ExecuteActivityAction {
...
+  message TimeoutActivity {
+    // How many attempts should timeout before succeeding (1-indexed)
+    int32 fail_attempts = 1;
+  }
+
+  // Activity that skips heartbeats for N attempts (causing heartbeat timeout), then sends them.
+  // Tests HeartbeatTimeout behavior with retries.
+  // Duration is derived from activity info's HeartbeatTimeout (runs for 2x heartbeat timeout).
```
Sushisource:
```suggestion
  // On success, runs for 2x heartbeat timeout.
```
To be consistent with the other one - but, again I would also probably just parameterize here.

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
```diff
@@ -419,6 +434,63 @@ public static byte[] Payload(byte[] inputData, int bytesToReturn)
...
+    public static async Task Heartbeat(ExecuteActivityAction.Types.HeartbeatTimeoutActivity config)
+    {
+        var info = ActivityExecutionContext.Current.Info;
+        var shouldSendHeartbeats = info.Attempt > config.FailAttempts;
+
+        // Run activity for 2x the heartbeat timeout
+        // Ensures we miss enough heartbeat intervals (if not sending heartbeats).
+        var duration = TimeSpan.FromTicks(info.HeartbeatTimeout!.Value.Ticks * 2);
+
+        // Unified loop with conditional heartbeat sending
```
Sushisource: This comment is very AI

workers/java/io/temporal/omes/ActivitiesImpl.java
```diff
@@ -36,4 +36,56 @@ public void client(
...
+      io.temporal.omes.KitchenSink.ExecuteActivityAction.HeartbeatTimeoutActivity config)
+      throws InterruptedException {
+    var activityInfo = Activity.getExecutionContext().getInfo();
+    boolean shouldSendHeartbeats = activityInfo.getAttempt() > config.getFailAttempts();
+
+    // Run activity for 2x the heartbeat timeout
+    // Ensures we miss enough heartbeat intervals (if not sending heartbeats).
+    long durationMs = activityInfo.getHeartbeatTimeout().toMillis() * 2;
+
+    // Unified loop with conditional heartbeat sending
```
Sushisource: Same very AI comment 😆

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
THardy98: lol, removed 😂 
what can i say, it's good at replicating changes between the workers

workers/java/io/temporal/omes/ActivitiesImpl.java
THardy98: ditto 😅

workers/proto/kitchen_sink/kitchen_sink.proto
THardy98: Added fields for success/failure durations

workers/proto/kitchen_sink/kitchen_sink.proto
THardy98: ditto

PR #232, opened 2025-10-20
Add `min-throughput-per-hour` option for `throughput_stress`

THardy98:
## What was changed
Added a `min-throughput-per-hour` option for `throughput_stress` This can be used as criteria to determine whether a scenario succeeds or not.

For example, expecting 'x' number of workflows to completed within a certain amount of time.

## Why?
Provides the ability to run `throughput_stress` with work-based criteria.

### DIFF
--- a/cmd/cli/run_scenario.go
+++ b/cmd/cli/run_scenario.go
@@ -57,7 +57,6 @@ type scenarioRunConfig struct {
 	timeout                       time.Duration
 	doNotRegisterSearchAttributes bool
 	ignoreAlreadyStarted          bool
-	minThroughputPerHour          float64
 }
 
 func (r *scenarioRunner) addCLIFlags(fs *pflag.FlagSet) {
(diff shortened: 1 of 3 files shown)


cmd/cli/run_scenario.go
```diff
@@ -84,6 +85,9 @@ func (r *scenarioRunConfig) addCLIFlags(fs *pflag.FlagSet) {
 			"If the search attributes are not registed by the scenario they must be registered through some other method")
 	fs.BoolVar(&r.ignoreAlreadyStarted, "ignore-already-started", false,
 		"Ignore if a workflow with the same ID already exists. A Scenario may choose to override this behavior.")
+	fs.Float64Var(&r.minThroughputPerHour, "min-throughput-per-hour", 0,
+		"Minimum workflow throughput required (workflows/hour, default: 0, disabled). "+
+			"Only scenarios that implement this check will enforce it (currently: throughput_stress)")
```
Sushisource: This should probably be added as a scenario specific flag (at least for now) -- see the flag consts at the top of throughputstress

THardy98: Sure - figured it might be useful to generally have and noop on other scenarios, don't mind either way

Sushisource: Yeah, I agree if/when we add it to others.

stephanos (approved):
LGTM.

I was deliberating whether a test case in throughput_stress_test.go would be needed, but it seems pretty straight-forward.

PR #236, opened 2025-10-24
Overhaul workflow verification

stephanos:
## What was changed

(1) Added new flag `--continue-on-error` to run scenario. 
(2) Extracted workflow completion check into reusable component.
(3) Moved state for completion check into executor.
(4) Consolidated Search Attribute for completion checking.
(5) Added `WorkflowExecutionAlreadyStarted` handling for completion checking.
(6) Consolidated post-scenario verification. 

## Why?

(1) It is very useful to move past any non-transient errors during a scenario run and keep going; reporting the errors at the end. This helps with runs where a single workflow fails to complete and would block further execution.

(2-6) Checking workflow completion is a crucial verification step; this change make it easier to re-use across scenarios.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -342,7 +342,6 @@ func (e *ebbAndFlowExecutor) spawnWorkflowWithActivities(
 		e.Logger.Errorf("kitchensink workflow failed for iteration %d: %v", iteration, err)
 	}
 	e.completedActivities.Add(activities)
-	e.incrementTotalCompletedWorkflow()
 
 	// Record completion in executor state for verification
 	e.stateLock.Lock()


scenarios/throughput_stress.go
```diff
@@ -183,43 +180,30 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
-	}
-
 	t.lock.Lock()
 	isResuming := t.isResuming
 	currentState := *t.state
 	t.lock.Unlock()
 
 	if isResuming {
 		info.Logger.Info(fmt.Sprintf("Resuming scenario from state: %#v", currentState))
-		info.Configuration.StartFromIteration = int(currentState.CompletedIterations) + 1
```
stephanos, line 202: This was technically a good idea, but easily causes one-off issues. For example, when the all iterations were created but the run crashes and resumes, it would error since the start iteration cannot be larger than the target number of iterations. And now with the `WorkflowExecutionAlreadyStarted` handling, this is safe.

scenarios/throughput_stress.go
```diff
@@ -183,43 +180,30 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 	}
 	t.runID = info.RunID
 
-	// Track start time of current run
-	currentRunStartTime := time.Now()
```
stephanos, line 187: Moved into ExecutorState.

scenarios/throughput_stress.go
```diff
@@ -183,43 +180,30 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
...
-
-	// Add search attribute, if it doesn't exist yet, to query for workflows by run ID.
-	// Running this on resume, too, in case a previous Omes run crashed before it could add the search attribute.
-	if err := loadgen.InitSearchAttribute(ctx, info, ThroughputStressScenarioIdSearchAttribute); err != nil {
-		return err
-	}
```
stephanos, line 193: Moved into WorkflowCompletionChecker

loadgen/kitchen_sink_executor.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 type KitchenSinkExecutor struct {
+	GenericExecutor
```
stephanos, line 11: So it inherits the Resumable and Verifyable behavior.

loadgen/scenario.go
```diff
@@ -380,10 +425,12 @@ func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSi
...
 	executeErr := executor.Handle.Get(cancelCtx, nil)
 	if executeErr != nil {
-		return fmt.Errorf("failed to execute kitchen sink workflow: %w", executeErr)
+		return fmt.Errorf("failed to execute kitchen sink workflow (workflowID: %s, runID: %s): %w",
+			executor.Handle.GetID(), executor.Handle.GetRunID(), executeErr)
 	}
 	if clientActionsErr := clientActionsErrPtr.Load(); clientActionsErr != nil {
-		return fmt.Errorf("kitchen sink client actions failed: %w", *clientActionsErr)
+		return fmt.Errorf("kitchen sink client actions failed (workflowID: %s, runID: %s): %w",
+			executor.Handle.GetID(), executor.Handle.GetRunID(), *clientActionsErr)
```
stephanos, line 430: Provide more details on failure.

PR #239, opened 2025-10-24
Remove fairness validation

stephanos:
## What was changed

Remove fairness validation from ebb and flow scenario.

## Why?

Suspected memory leak. And it's not really used anyway.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


workers/go/ebbandflow/workflow.go
```diff
@@ -40,15 +40,6 @@ func EbbAndFlowTrackWorkflow(ctx workflow.Context, params *ebbandflow.WorkflowPa
 				opts.Priority.PriorityKey = int(activity.Priority.PriorityKey)
 			}
 
-			// Set fairness, if specified
-			fairnessKey := activity.GetFairnessKey()
-			fairnessWeight := activity.GetFairnessWeight()
-			if fairnessKey != "" {
```
dnr, line 46: we need fairness keys though.. put it in the real field now that it's released

stephanos: Oops; reverted!

stephanos: I'd need to update the SDK to use the real field now.

dnr (approved):

PR #241, opened 2025-10-28
Fixes for Java worker kitchen sink workflow

THardy98:
## What was changed
Fixes for Java worker kitchen sink workflow
- add search attributes to child workflow for visibility check
- add client action executor
- explicitly propagate search attributes for CAN workflows (they aren't propagated from CAN -> CAN workflows) for visibility check

## Why?
Blocker for enabling throughput stress testing on java workers

2. How was this tested:
Can now run:
```
    --language java \
    --iterations 10 \
    --option internal-iterations=10
```

3. Any docs updates needed?
No

### DIFF
--- a/workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
+++ b/workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
@@ -56,7 +56,7 @@ public class KitchenSinkWorkflowImpl implements KitchenSinkWorkflow {
     if (signalActions.hasDoActionsInMain()) {
       signalActionQueue.put(signalActions.getDoActionsInMain());
     } else if (signalActions.hasDoActions()) {
-      signalActionQueue.put(signalActions.getDoActions());
+      handleActionSet(signalActions.getDoActions());
     } else {
       throw ApplicationFailure.newNonRetryableFailure(
           "Signal actions must have a recognizable variant", "");


workers/java/io/temporal/omes/KitchenSinkWorkflowImpl.java
```diff
@@ -139,7 +140,22 @@ private Payload handleAction(KitchenSink.Action action) {
...
+      if (currentSearchAttributes != null && currentSearchAttributes.size() > 0) {
+        // Explicitly preserve search attributes
+        ContinueAsNewOptions options =
+            ContinueAsNewOptions.newBuilder()
+                .setTypedSearchAttributes(currentSearchAttributes)
+                .build();
+        Workflow.continueAsNew(options, continueAsNew.getArgumentsList().get(0));
+      } else {
+        // No search attributes to preserve
+        Workflow.continueAsNew(continueAsNew.getArgumentsList().get(0));
```
Sushisource, line 157: Nit: I think in principle you could just always write this like the first  branch and it'd have the same effect either way.

AI comments inside the branches not super useful 😄

THardy98: lol these comments were mine, rip 🙃 
I have grown indistinguishable from AI

PR #242, opened 2025-10-28
Various fixes for .NET kitchen sink workflow

THardy98:
## What was changed
Various fixes for .NET kitchen sink workflow
- fix workflow type
- fix concurrent action handling in .NET KitchenSink workflow (Replace async void with async Task in concurrent action processing. The previous implementation caused tasks to complete synchronously, leading to parent workflows finishing while handlers were still running)
- added search attributes to child workflows to pass visibility check

## Why?
Allow throughput stress to be run with .NET worker

2. How was this tested:
Can now run:
```
    --language dotnet \
    --iterations 10 \
    --option internal-iterations=10
```

3. Any docs updates needed?
No


workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
```diff
@@ -154,20 +158,25 @@ await HandleAwaitableChoiceAsync(
         }
         else if (action.ExecChildWorkflow is { } execChild)
         {
-            var childType = execChild.WorkflowType ?? "kitchenSink";
+            var childType = string.IsNullOrEmpty(execChild.WorkflowType) ? "kitchenSink" : execChild.WorkflowType;
```
cretz, line 161: Curious, does this not come over as null? I think it should from the kitchen sink caller side, right?

Sushisource (approved):

workers/dotnet/Temporalio.Omes/KitchenSinkWorkflow.cs
THardy98: It came as empty string

cretz: It should not come as empty string. We should fix the caller side IMO.

PR #249, opened 2025-11-01
Pass all non-default flags

stephanos:
## What was changed

Fixed a bug where explicitly set flags were not passed to the worker.

Before
```
% go run ./cmd run-worker --scenario workflow_on_many_task_queues --language go --run-id test --namespace test
2025-10-31T18:17:23.442-0700	INFO	workers/run.go:146	Starting worker with command: [./program --task-queue omes-test]
```

After
```
go run ./cmd run-worker --scenario workflow_on_many_task_queues --language go --run-id test --namespace test
2025-10-31T18:17:42.389-0700	INFO	workers/run.go:146	Starting worker with command: [./program --task-queue omes-test --namespace=test]
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/workers/run.go
+++ b/workers/run.go
@@ -187,7 +187,7 @@ func (r *Runner) Run(ctx context.Context, baseDir string) error {
 
 func passthrough(fs *pflag.FlagSet, prefix string) (flags []string) {
 	fs.VisitAll(func(f *pflag.Flag) {
-		if f.DefValue == f.Value.String() {
+		if !f.Changed {
 			return
 		}
 		flags = append(flags, fmt.Sprintf("--%s=%s",


workers/run.go
```diff
@@ -186,7 +186,10 @@ func (r *Runner) Run(ctx context.Context, baseDir string) error {
 }
 
 func passthrough(fs *pflag.FlagSet, prefix string) (flags []string) {
-	fs.Visit(func(f *pflag.Flag) {
+	fs.VisitAll(func(f *pflag.Flag) {
+		if f.DefValue == f.Value.String() {
```
dnr: ok, I see the problem, but I think using `f.Changed` would be better. does that work?

stephanos: Yup! works, too: `Starting worker with command: [./program --task-queue omes-test --namespace=test]`

PR #251, opened 2025-11-04
Ebb and flow scenario changes

dnr:
## What was changed
- Fix backlog count in ebbandflow (count activities instead of workflows)
- Do just time-based oscillation
- Rename phase to "period"

## Why?
More understandable and correct ebbandflow test


stephanos (approved):

stephanos:
I'll merge already because so can easily use this in my branch.

Issue #223, opened 2025-10-03
[Feature Request] Allow to ignore `concurrent client actions are not supported` failures


stephanos:
### Is your feature request related to a problem? Please describe.

When running the fuzzer, we have no control over what features it might select.

### Describe the solution you'd like

In that case, it would be better to silently ignore any unimplemented features.

PR #252, opened 2025-11-06
Add `worker-err-on-unimplemented` flag

THardy98:
## What was changed
Add `worker-err-on-unimplemented` option to explicitly fail on unimplemented actions for kitchen sink workflows. This is currently scoped to handling concurrent client actions which are not implemented in all workers except Go.

Behavior has changed to, by default, skip/no-op unimplemented concurrent client actions instead of failing. Setting this option to `true` enables the previous behavior (failing on unimplemented concurrent client actions).

Opted for a worker option because it's the simplest to implement. Other options included:
- creating an option for kitchen sink workflow, which would control a new flag in the `WorkflowInput` proto, used to tell the worker whether they are to ignore concurrent client actions
- creating an option in the fuzzer scenario which would control whether the fuzzer could generate `ClientActivity` with concurrent action sets

but they're both more complex. We can revisit this if necessary, but this is functional and a small diff.

Additional small fix for TS worker when spawning child workflows (found when running the fuzzer)

## Why?
The fuzzer scenario will always trigger this error otherwise, which makes it difficult to find actual bugs with the fuzzer

Resolves: https://github.com/temporalio/omes/issues/223

2. How was this tested:
Ran
```
      --scenario fuzzer \
      --language {language} \
      --iterations 1
```
with and without the `--worker-err-on-unimplemented` flag.

### DIFF
--- a/loadgen/helper_sleepactivity.go
+++ b/loadgen/helper_sleepactivity.go
@@ -39,7 +39,7 @@ type SleepActivityGroupConfig struct {
 	FairnessWeight *DistributionField[float32] `json:"fairnessWeight"`
 }
 
-func ParseAndValidateSleepActivityConfig(jsonStr string) (*SleepActivityConfig, error) {
+func ParseAndValidateSleepActivityConfig(jsonStr string, requireCount bool) (*SleepActivityConfig, error) {
 	if jsonStr == "" {
 		return nil, nil
 	}
@@ -47,7 +47,7 @@ func ParseAndValidateSleepActivityConfig(jsonStr string) (*SleepActivityConfig, 
 	if err := json.Unmarshal([]byte(jsonStr), config); err != nil {
 		return nil, fmt.Errorf("failed to parse SleepActivityConfig JSON: %w", err)
 	}
-	if config.Count == nil {
+	if requireCount && config.Count == nil {
 		return nil, fmt.Errorf("SleepActivityConfig: Count field is required")
 	}
 	if config.Groups == nil || len(config.Groups) == 0 {
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -21,12 +21,16 @@ const (
 
 const (
 	// MinBacklogFlag defines the minimum backlog to target.
+	// Note that it controls total backlog size, not counting partitions or priority levels.
 	MinBacklogFlag = "min-backlog"
 	// MaxBacklogFlag defines the maximum backlog to target.
+	// Note that it controls total backlog size, not counting partitions or priority levels.
 	MaxBacklogFlag = "max-backlog"
-	// PhaseTimeFlag defines the duration of each growing and draining phase.
-	PhaseTimeFlag = "phase-time"
-	// SleepDurationFlag defines the duration an activity sleeps for.
+	// PeriodFlag defines the period of oscilation of backlog size.
+	PeriodFlag = "period"
+	// SleepDurationFlag defines the duration an activity sleeps for (default 1ms).
+	// This can be used to slow down activity processing, but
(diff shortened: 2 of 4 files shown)


cmd/clioptions/worker.go
```diff
@@ -35,5 +36,6 @@ func (m *WorkerOptions) FlagSet(prefix string) *pflag.FlagSet {
 	m.fs.IntVar(&m.MaxConcurrentWorkflowTasks, prefix+"max-concurrent-workflow-tasks", 0, "Max concurrent workflow tasks")
 	m.fs.IntVar(&m.ActivityPollerAutoscaleMax, prefix+"activity-poller-autoscale-max", 0, "Max for activity poller autoscaling (overrides max-concurrent-activity-pollers")
 	m.fs.IntVar(&m.WorkflowPollerAutoscaleMax, prefix+"workflow-poller-autoscale-max", 0, "Max for workflow poller autoscaling (overrides max-concurrent-workflow-pollers")
+	m.fs.BoolVar(&m.ErrOnUnimplemented, prefix+"err-on-unimplemented", false, "Fail on unimplemented actions (currently this only applies to concurrenct client actions)")
```
Sushisource:
```suggestion
	m.fs.BoolVar(&m.ErrOnUnimplemented, prefix+"err-on-unimplemented", false, "Fail on unimplemented actions (currently this only applies to concurrent client actions)")
```

stephanos:
> I'm not sure this is the way we'd want to do this. Better to specify to the fuzzer that it just shouldn't generate these kinds of actions, I think. Otherwise you get this big mismatch between what you think would happen and what actually happens.

The big problem with that approach, though, is that the kind of actions that work/don't work are worker implementation specific (and the worker support was already fragmented before and silently ignoring actions - it returns an error now to make it explicit). That means the fuzzer needs to _know_ which actions are supported across _all_ implementations and would only be allowed to emit those. IMO that creates more work than skipping them. We could consider emitting a warning log so it's not completely silent.

The alternative is to only allow the fuzzer to produce actions that are valid in all workers. That would force us to ensure the implementations are all aligned, but also delay testing these until they are available everywhere. I'm worried it turns into a game of whack-a-mole.

THardy98:
> Better to specify to the fuzzer that it just shouldn't generate these kinds of actions, I think. Otherwise you get this big mismatch between what you think would happen and what actually happens.

Yeah, I considered this option in the PR desc.

I think if you want to avoid the big mismatch the easiest thing to do is flip the behavior, so that the default doesn't change and you'd have to explicitly provide the flag to skip unimplemented actions (and log if we prefer it to not be silent).

To your point, restricting the fuzzer to create specific actions per worker I think could be generally useful outside of just concurrent client actions (i.e. if we want the fuzzer to test/focus on a subset of behavior etc). Would we be ok with filing a separate issue for that (documenting that we could deprecate/remove this in the process) and moving forward with this in the meantime? I think we want to move quickly given our current testing efforts across SDK/server and this currently makes the fuzzer unusable for non-Go workers (and is much less work, at least FWICT)

Sushisource:
> Yeah, I considered this option in the PR desc.

My bad

>That means the fuzzer needs to know which actions are supported across all implementations

So I don't know that that part is necessarily true. The CI/runner/whatever kicking off the scenarios can be responsible for that and just pass the right flags to the runner for the given language.


>  I think we want to move quickly given our current testing efforts across SDK/server 

That's totally reasonable. I'm fine with this then. Maybe let's add a warning per Stephan's suggestion

THardy98:
@Sushisource good to go?

stephanos:
This is still there: https://github.com/temporalio/omes/blob/52e6dd9671dc561e8eaaf33f6c7dc3cfffc9826b/workers/python/client_action_executor.py#L25

PR #257, opened 2025-11-18
Wait for ExecuteChildWorkflowAction to complete by default

stephanos:
## What was changed

Make .NET, Python and TS kitchensink workflows wait for child workflow execution completion.

Go and Java already do the right thing.

## Why?

The default wait `AwaitableChoice` policy is `waitFinish`, but these impls only wait for the start, not completion.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/loadgen/kitchen_sink_executor_test.go
+++ b/loadgen/kitchen_sink_executor_test.go
@@ -144,20 +144,21 @@ func TestKitchenSink(t *testing.T) {
 									WorkflowType: "kitchenSink",
 									Input: []*common.Payload{
 										ConvertToPayload(&WorkflowInput{
-											InitialActions: ListActionSet(NewTimerAction(1 * time.Millisecond)),
+											InitialActions: ListActionSet(NewEmptyReturnResultAction()),
 										})},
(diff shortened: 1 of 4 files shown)


loadgen/kitchen_sink_executor_test.go
```diff
@@ -144,20 +144,21 @@ func TestKitchenSink(t *testing.T) {
 									WorkflowType: "kitchenSink",
 									Input: []*common.Payload{
 										ConvertToPayload(&WorkflowInput{
-											InitialActions: ListActionSet(NewTimerAction(1 * time.Millisecond)),
+											InitialActions: ListActionSet(NewEmptyReturnResultAction()),
```
THardy98, line 147: What's this change for?

stephanos: Without a `ReturnResult` action the workflow never returns. That's why it had the abandon policy on it before.

```
Error:      	Received unexpected error:
                	            	timed out while waiting for runs to complete: context deadline exceeded
                	Test:       	TestKitchenSink/ExecActivity/ExecChildWorkflow/go
```

PR #263, opened 2025-11-22
add wait command

chaptersix:
## What was changed
Add a global flag called `--wait` that will wait indefinitely after any command is executed

## Why?
When running omes in a Kubernetes environment, it executes a load test on container startup and then exits immediately. Upon exit, Kubernetes restarts the pod, causing the load test to run again. This creates an endless cycle of load tests without a clean way to stop load test execution.

The `--wait` flag addresses this by allowing omes to wait indefinitely after completing its work (or doing nothing at all), keeping the container alive. This enables:
- Setting `--wait`  in addition other argument to running one-time load tests
- Setting `--wait` as a default arg in the deployment, allowing the pod to start idle and wait for an engineer to update the deployment with the desired load test configuration.


chaptersix:
We could also allow the embedded server to stay alive. This would allow us to query it after the execution of a load test during development.

chaptersix:
solve the issue in k8

PR #266, opened 2025-11-26
Spin up local Prometheus instance

THardy98:
## What was changed
Add ability to spin up local Prometheus instance to collect metrics.
Prometheus instance can be snapshotted.
Metrics can be exported as a JSON file.

## Why?
Collect & dump metrics at the end of a run.
JSON metrics can be uploaded to data platform S3 which then can be queried via data viz tools like Omni

2. How was this tested:
```
    --run-id test-$(date +%s) \
    --language {language} \
    --duration 5m \
--prom-instance-addr http://127.0.0.1:9090 \
--prom-instance-config \
--prom-snapshot \
--prom-export-worker-metrics {your-filename}.jsonl \
--prom-export-metrics-step 3s
```
for all worker languages. 

Example of metrics output:
[temp.txt](https://github.com/user-attachments/files/23810079/temp.txt)

Currently, process-level metrics are limited to the Go worker

### DIFF
--- a/cmd/clioptions/metrics.go
+++ b/cmd/clioptions/metrics.go
@@ -142,15 +142,19 @@ func (m metricsTimer) Record(duration time.Duration) {
 // to collect CPU and memory metrics. The standard ProcessCollector from Prometheus
 // only works on Linux (requires /proc).
 type processCollector struct {
-	pid            int32
+	process        *process.Process
 	cpuDesc        *prometheus.Desc
 	memDesc        *prometheus.Desc
 	memPercentDesc *prometheus.Desc
 }
 
-func newProcessCollector() *processCollector {
+func newProcessCollector() (*processCollector, error) {
+	p, err := process.NewProcess(int32(os.Getpid()))
+	if err != nil {
+		return nil, err
+	}
 	return &processCollector{
-		pid: int32(os.Getpid()),
+		process: p,
 		cpuDesc: prometheus.NewDesc(
 			"process_cpu_percent",
 			"CPU usage as a percentage (100 = 1 core).",
@@ -166,7 +170,7 @@ func newProcessCollector() *processCollector {
 			"Memory usage as a percentage of total system memory.",
 			nil, nil,
 		),
-	}
+	}, nil
 }
 
 func (c *processCollector) Describe(ch chan<- *prometheus.Desc) {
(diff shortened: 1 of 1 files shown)


cmd/clioptions/metrics.go
```diff
@@ -137,40 +138,94 @@ func (m metricsTimer) Record(duration time.Duration) {
...
+
+func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
+	p, err := process.NewProcess(c.pid)
+	if err != nil {
+		return
+	}
+
+	// CPU times (user + system)
+	if times, err := p.Times(); err == nil {
+		ch <- prometheus.MustNewConstMetric(c.cpuDesc, prometheus.CounterValue, times.User+times.System)
```
Sushisource: You don't want to use these times, you'll want to use https://pkg.go.dev/github.com/shirou/gopsutil/v4@v4.25.10/cpu#Percent which will be much more interpretable

cmd/clioptions/prometheus.go
```diff
@@ -0,0 +1,258 @@
...
+	// Path to Prometheus config file for starting a local Prometheus instance.
+	// If empty, no local Prometheus will be started.
+	ConfigPath string
+	// If true, create a TSDB snapshot on shutdown.
+	Snapshot bool
+	// Path to export worker metrics (CPU/memory) as JSON on shutdown.
+	// If empty, no export will be performed.
+	ExportWorkerMetricsPath string
+	// Worker job to export.
+	ExportWorkerMetricsJob string
```
Sushisource, line 30: Do we need this one? Can it just always use the same job name (at least to start with, unless we have a need to change it)?

cmd/clioptions/prometheus.go
```diff
@@ -0,0 +1,258 @@
...
+	// Step interval to sample timeseries metrics.
+	// If not provided a default interval of 15s will be used.
+	// (only used if ExportWorkerMetricsPath is provided)
```
Sushisource:
```suggestion
	// Step interval when sampling timeseries metrics for the JSON output format.
	// If not provided a default interval of 15s will be used.
	// (only used if ExportWorkerMetricsPath is provided)
```

THardy98: fixed

cmd/clioptions/prometheus.go
THardy98: It defaults to `omes-worker` as is, so it effectively has the behavior you describe.

cmd/clioptions/metrics.go
THardy98: nice - `p.Percent(0)`, measuring CPU % since last call
ty

THardy98:
Added a metric to capture memory as a percentage of total system memory (complements and easier to read than # of bytes)

cmd/clioptions/prometheus.go
Sushisource: Sure, so, why don't we just get rid of the extra flag if we don't need it? It's just more noise right now

PR #269, opened 2025-12-04
Output metrics as parquet

THardy98:
## What was changed
Output metrics as parquet

## Why?
Smaller, better format for querying (as we will be doing with S3 + Apache Hive + Amazon Athena) for Omni.

2. How was this tested:
```
    --run-id test-$(date +%s) \
    --language {language} \
    --duration 5m \
--prom-instance-addr 127.0.0.1:9090 \
--prom-instance-config \
--prom-snapshot \
--prom-export-worker-metrics {your-filename}.parquet \
--prom-export-metrics-step 3s
```


Sushisource (commented):
Should we just get rid of the other JSON format and only support Parquet if that's what we're using for Omni? Seems like no real reason to keep around the other one

Sushisource (approved):

PR #271, opened 2025-12-09
add the go workflows for scheduler stress

chaptersix:
## What was changed
* add the go sdk workflows for scheduler stress
* modify scenario to work with the workers. Primarily using task queues the workers are actually listening on

## Why?
So workflows created by the scheduler are actioned on.

### DIFF
--- a/cmd/clioptions/prometheus.go
+++ b/cmd/clioptions/prometheus.go
@@ -2,12 +2,14 @@ package clioptions
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 	"time"
 
+	"github.com/parquet-go/parquet-go"
+	"github.com/parquet-go/parquet-go/compress/zstd"
 	"github.com/prometheus/client_golang/api"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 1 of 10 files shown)


scenarios/scheduler_stress.go
```diff
@@ -96,16 +97,16 @@ const (
...
 	DefaultScheduledWorkflowType        = NoopScheduledWorkflowType
 	DefaultOverlapPolicy                = "all"
 	DefaultEnableChasmScheduler         = true
 )
 
 // Default duration constants
 var (
-	DefaultSchedulerDurationPerIteration = time.Minute
-	DefaultWaitTimeBeforeCleanup         = 5 * time.Second
+	DefaultSchedulerDurationPerIteration = 30 * time.Second
```
lina-temporal, line 108: I would think we'd want this higher than lower, so that we can accumulate a large bulk of schedules as the stress test runs. That said, it looks like this is configurable via the `"scheduler-duration-per-iteration"` flag; is the plan to tweak the parameter at runtime while running the load test?

chaptersix: yes, that's correct. The defaults are slightly nicer for running it against the embedded server. I don't have a real sense of what they should be right now.

PR #273, opened 2025-12-12
Extend metrics schema

THardy98:
## What was changed
Minor cleanup/code movement and adding additional fields to `MetricLine`.

Added fields still need to be plumbed through (will be empty for now):
- `Environment`: environment that the scenario was run from 
- `BuildID`: build ID of the worker
- `Scenario`: scenario name
- `RunConfigProfile`: profile name of the run configuration
- `WorkerConfigProfile`: profile name of the worker configuration

where a `profile` maps to a configuration.

Want some feedback to establish the metrics schema. The goal is to have a schema we anticipate to not have to change.

Plumbing to come in a subsequent PR (in progress, will link).

## Why?
Allows us to filter queries on more parameters that will be useful.
Establish the schema for the data team to create omni topics for.

2. How was this tested:
```
    --run-id test-$(date +%s) \
    --language go \
    --duration 10m \
--prom-instance-addr 127.0.0.1:9090 \
--prom-instance-config \
--prom-snapshot \
--prom-export-worker-metrics go-10m.parquet \
--prom-export-metrics-step 3
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ permissions:
 
 env:
   MISE_VERSION: v2025.12.1
+  MISE_NODE_VERIFY: no
 
 on: # rebuild any PRs and main branch changes
   pull_request:


THardy98:
> @THardy98
> 
> > , I want to add an /info endpoint to the worker's prom metrics server, which we can query to get the worker's run config at metrics export time.
> 
> Is this just for reporting it with the tests? Shouldn't we know already because we set it? (I'm not necessarily opposed, just asking).
> 
> The separation of types makes sense to me

Yes - it'd be for reporting, but I can see the client having worker config/info being useful generally.

If we spawn the worker and client in separate processes, their configurations are not provided to each other. `run-scenario-with-worker` is an exception because we spawn both from a single process, so the client has complete information

PR #274, opened 2025-12-18
Worker Performance Testing on ECS

THardy98:
## What was changed
Add `run-scenario-ecs.yml` GHA workflow.

Can be used to manually dispatch worker performance tests that run on cloud (AWS Fargate).

Some characteristics:
- Worker runs in a separate Fargate task for isolation (separate firecracker vm)
- Server instance runs in cloud
- Can provide a GH commit to run against a specific version of worker
- Can specify `is-experiment` to distinguish from regular benchmark testing (useful for ad-hoc testing/experimenting, i.e. to test local perf-sensitive changes without disrupting signal from nightlies)
----

Some improvements to be made in the future:
- the workflow is quite long (though ~80 lines are inputs/env vars), could translate to code/scripts, and make it trigger-able from the CLI
- formalize the ECS setup with IaC
- handle worker crashes mid-way through the scenario (orphans the client)

-----
**NOTE**:
Leaving draft, to revert some changes before merging:
- `DURATION` should default to `5h` (shortened for testing)
- `TIMEOUT` should default to `5h30m` (shortened for testing)
- remove hook that triggers workflow on push
- change the `Checkout Omes` branch (currently `per-version-taskdefs`, will need to be `main`)

## Why?
Useful to run worker performance tests

2. How was this tested:
Numerous trial runs.

### DIFF
--- a/cmd/clioptions/metrics.go
+++ b/cmd/clioptions/metrics.go
@@ -105,8 +105,7 @@ func (p *PrometheusInstanceFlags) FlagSet(prefix string) *pflag.FlagSet {
 	}
 	p.fs = pflag.NewFlagSet(prefix+"prom_instance_options", pflag.ExitOnError)
 	p.fs.StringVar(&p.Address, prefix+"prom-instance-addr", "", "Prometheus instance address")
-	p.fs.StringVar(&p.ConfigPath, prefix+"prom-instance-config", "", "Start a local Prometheus instance with the specified config file (default: prom-config.yml)")
-	p.fs.Lookup(prefix + "prom-instance-config").NoOptDefVal = "prom-config.yml"
+	p.fs.StringVar(&p.ConfigPath, prefix+"prom-instance-config", "prom-config.yml", "Start a local Prometheus instance with the specified config file (default: prom-config.yml)")
 	p.fs.BoolVar(&p.Snapshot, "prom-snapshot", false, "Create a TSDB snapshot on shutdown")
 	p.fs.StringVar(&p.ExportWorkerMetricsPath, prefix+"prom-export-worker-metrics", "", "Export worker process metrics to the specified file on shutdown")
 	p.fs.StringVar(&p.ExportWorkerMetricsJob, prefix+"prom-export-worker-job", "omes-worker", "Name of the worker job to export metrics for")


.github/workflows/run-scenario-ecs.yml
```diff
@@ -0,0 +1,266 @@
...
+            --build-arg SDK_VERSION=./repo \
+            --build-arg BUILD_CORE_RELEASE=true \
+            --build-arg PREBUILD_WORKER=true \
+            -t $ECR_REPO:$IMAGE_TAG \
+            .
+
+          docker push $ECR_REPO:$IMAGE_TAG
+          echo "WORKER_IMAGE=$ECR_REPO:$IMAGE_TAG" >> $GITHUB_ENV
+
+      - name: Register worker task definition
```
Sushisource, line 157: Are we going to leave a bajillion "task definitions" lying around? 

I'm no AWS expert but I do know one thing that's always a problem is cleaning stuff up.

.github/workflows/run-scenario-ecs.yml
```diff
@@ -0,0 +1,266 @@
...
+          docker build \
+            -f dockerfiles/${LANGUAGE}.Dockerfile \
+            --build-arg SDK_DIR=./sdk \
+            --build-arg SDK_VERSION=./repo \
+            --build-arg BUILD_CORE_RELEASE=true \
+            --build-arg PREBUILD_WORKER=true \
+            -t $ECR_REPO:$IMAGE_TAG \
+            .
+
+          docker push $ECR_REPO:$IMAGE_TAG
```
Sushisource, line 154: We probably also need to clean up old images at some point. We get charged for those I'd assume?

.github/workflows/run-scenario-ecs.yml
```diff
@@ -0,0 +1,266 @@
...
+        run: |
+          set -e
+          IMAGE_TAG="omes-client"
+
+          docker build \
+            -f dockerfiles/cli-prometheus.Dockerfile \
+            -t $ECR_REPO:$IMAGE_TAG \
+            .
+
+          docker push $ECR_REPO:$IMAGE_TAG
```
Sushisource, line 188: Is the fact that there's only one client image we re-write potentially going to cause issues? Should we tag this with omes' git hash?

dockerfiles/cli-prometheus-entrypoint.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -e
+
+# Cleanup worker on any exit (success, failure, or signal)
+cleanup() {
+  if [ -n "$WORKER_TASK_ARN" ] && [ -n "$ECS_CLUSTER" ]; then
+    echo "Stopping worker task: $WORKER_TASK_ARN"
+    aws ecs stop-task --cluster "$ECS_CLUSTER" --task "$WORKER_TASK_ARN" --region "${AWS_REGION:-us-west-2}" --reason "Client exited" || true
+  fi
+}
```
Sushisource, line 10: This file name is about prometheus, but it's also just generally running omes and cleaning up the worker. Probably want a more generic file name

dockerfiles/cli-prometheus.Dockerfile
```diff
@@ -0,0 +1,81 @@
+# CLI image with Prometheus for scenarios that need worker metric scraping
```
Sushisource, line 1: This is mostly duped with the other docker file. Can this one just use that one as a layer, or otherwise dedupe?

.github/workflows/run-scenario-ecs.yml
THardy98: We have lifecycle policies in our ECR instance, they are:
- Delete untagged after 1 day (failed/partial pushes)
- Keep only the 2 most recent python-worker-* images
- Keep only the 2 most recent java-worker-* images
- Keep only the 2 most recent go-worker-* images
- Keep only the 2 most recent typescript-worker-* images
- Keep only the 2 most recent dotnet-worker-* images
- Keep only the 2 most recent omes-client* images

In principle, we could be really aggressive and delete all images after a day, since we're pushing them on each run anyway (i'm open to changing it if we want). The cost is pretty cheap regardless (100GB of storage is $10)

.github/workflows/run-scenario-ecs.yml
THardy98: No, the existing image will still exist, but only the latest omes-client image will be tagged

.github/workflows/run-scenario-ecs.yml
THardy98: Task definitions belong to a "task definition family". When you register new task definitions under a family, they are called "task definition revisions". A family is like a namespace for a bunch of different versions ("revisions") of a task definition. For example:
```
Task definition family: go-omes-worker

Task definition revisions:
go-omes-worker:1 (initial)
go-omes-worker:2 (registered a new task definition under the "go-omes-worker" family)
go-omes-worker:3 (...)
```
As I understand it, there's no monetary cost to having many task definition revisions (they are free). AWS imposes a hard limit of 1M per task definition family. So there's no immediate concern here.

We could likely add a hook to clean up the task definitions in the same trap we use to stop the worker ECS task.

dockerfiles/cli-prometheus.Dockerfile
THardy98: Yes likely

.github/workflows/run-scenario-ecs.yml
Sushisource: Dope. That works.

.github/workflows/run-scenario-ecs.yml
Sushisource: Works for me.

PR #276, opened 2025-12-18
Add Options to scheduler stress

chaptersix:
Rework the scheduler_stress scenario for flexible ad-hoc load testing.

Changes:
- Add `skip-deletion`, `skip-creation`, and `worker-count` options for controlling load patterns without rebuilding
- Refactor execution into worker pool with channel-based schedule dispatch (configurable concurrency via `worker-count`)
- Add `executeWithExistingSchedules` path to re-run operations against previously created schedules (`skip-creation=true`)
- Add `retryWithBackoff` with exponential backoff and jitter for all schedule operations (create, describe, update, delete)
- Per-operation 30s timeouts to avoid hanging on unresponsive calls
- Set `TriggerImmediately` on schedule creation
- Debug logging for schedule deletion lifecycle


scenarios/scheduler_stress.go
```diff
@@ -164,53 +182,202 @@ func (s *SchedulerExecutor) Execute(ctx context.Context, run *loadgen.Run) error
...
-				}
-				<-ticker.C // Wait between update operations
+	// Start workers that will create and process schedules as configs arrive
+	var wg sync.WaitGroup
+
+	for range s.config.WorkerCount {
+		wg.Go(func() {
+			// Each worker keeps consuming from the channel until it's closed
+			for config := range scheduleConfigChan {
+				func() {
```
lina-temporal: I don't think this needs to be in a lambda, given that it's already in the `wg.Go(func() { ... })` scope

scenarios/scheduler_stress.go
```diff
@@ -164,53 +182,202 @@ func (s *SchedulerExecutor) Execute(ctx context.Context, run *loadgen.Run) error
...
+									logger.Errorw("Failed to delete schedule", "scheduleID", id, "error", err)
+								}
+								logger.Debugw("deleted", "scheduleID", id)
+							case <-ctx.Done():
+								logger.Infow("Context canceled")
+							}
+						}(sc.ScheduleID, start)
+					}
+
+					<-ticker.C
```
lina-temporal: think you could replace all of the ticker business with `<-time.After(s.config.OperationInterval)` here (or, probably, select over that and `ctx.Done`)

scenarios/scheduler_stress.go
```diff
@@ -164,53 +182,202 @@ func (s *SchedulerExecutor) Execute(ctx context.Context, run *loadgen.Run) error
...
+
+func (s *SchedulerExecutor) executeWithExistingSchedules(ctx context.Context, client client.Client, logger *zap.SugaredLogger) error {
+	// Create a channel to receive schedule IDs
+	scheduleIDChan := make(chan string, 100)
+	listErrChan := make(chan error, 1)
+
+	// Start a goroutine to list schedules and send them to the channel
+	go func() {
+		defer close(scheduleIDChan)
+		listErrChan <- s.listSchedules(ctx, client, scheduleIDChan, logger)
```
lina-temporal: IMO, the iterator should be responsible for closing the channel, instead of the caller (and if the caller cancels the channel, the iterator should stop producing).

scenarios/scheduler_stress.go
```diff
@@ -164,53 +182,202 @@ func (s *SchedulerExecutor) Execute(ctx context.Context, run *loadgen.Run) error
...
+							select {
+							case <-time.After(dur):
+								if err := s.deleteSchedule(ctx, client, id, logger); err != nil {
+									logger.Errorw("Failed to delete schedule", "scheduleID", id, "error", err)
+								}
+							case <-ctx.Done():
+								logger.Infow("Context canceled")
+							}
+						}(scheduleID, start)
+					}
```
lina-temporal: I think this can be factored out to a helper.

scenarios/scheduler_stress.go
```diff
@@ -237,41 +404,145 @@ func (s *SchedulerExecutor) createSchedule(ctx context.Context, c client.Client,
...
 
-func (s *SchedulerExecutor) describeSchedule(ctx context.Context, c client.Client, scheduleID string, logger *zap.SugaredLogger) error {
-	handle := c.ScheduleClient().GetHandle(ctx, scheduleID)
-	_, err := handle.Describe(ctx)
-	if err != nil {
-		var notFoundErr *serviceerror.NotFound
-		if errors.As(err, &notFoundErr) {
-			// Return nil if schedule is not found (already deleted or never existed)
-			logger.Debug("Schedule not found during describe operation", "scheduleID", scheduleID)
+// retryConfig defines retry behavior for schedule operations
```
lina-temporal: stale comment? I don't see a `retryConfig`

scenarios/scheduler_stress.go
```diff
@@ -237,41 +404,145 @@ func (s *SchedulerExecutor) createSchedule(ctx context.Context, c client.Client,
...
+const (
+	maxRetries      = 5
+	baseBackoff     = 100 * time.Millisecond
+	maxBackoff      = 10 * time.Second
+	backoffMultiple = 2.0
+)
+
+// retryWithBackoff executes an operation with exponential backoff and jitter on context deadline errors
+func retryWithBackoff(ctx context.Context, operation string, logger *zap.SugaredLogger, fn func() error) error {
```
lina-temporal, line 445: Why not use the SDK's built in retry mechanism for this?

scenarios/scheduler_stress.go
chaptersix: I added the lambda so I could use defer and maybe some variable scoping issues. I'll see if I can remove it

scenarios/scheduler_stress.go
chaptersix: I didn't want the duration of he API call to impact the wait time between operations. there's a sizable latency different between v1/v2. though this should have a select.

scenarios/scheduler_stress.go
chaptersix: I was getting a bunch of context cancellations and I don't have control over the sdk config. don't remember if the SDK was actually retrying or not.

scenarios/scheduler_stress.go
chaptersix: makes sense

PR #277, opened 2026-01-08
Add process metrics sidecar with /info endpoint

THardy98:
## What was changed
Add a separate HTTP server for process metrics and worker metadata, started when spawning SDK workers. This works for all language workers (Go/Python/Java/TS/.NET).

The sidecar serves:
- /metrics: Process metrics (CPU/memory) for the worker PID
- /info: Worker metadata (sdk_version, build_id)

The /info endpoint only returns fields that run-scenario doesn't already know. During metrics export, /info is fetched to populate additional fields in the parquet output.

## Why?
We opt for a Prom server sidecar for process metrics because core-based SDKs have their existing Prom server started by core, making it difficult to:
- expand its registry to include process metrics
- add a handler to surface additional metadata (i.e. worker metadata from /info`)
Using a sidecar server to capture process metrics is simple.

We add an `/info` endpoint to this sidecar server so that the scenario runner can fetch additional worker metadata for metrics export. This is useful in scenarios where the scenario runner and the worker are running on different machines (as is the case in cloud worker testing).

2. How was this tested:
Added some basic integration tests

### DIFF
--- a/cmd/clioptions/metrics.go
+++ b/cmd/clioptions/metrics.go
@@ -33,7 +33,6 @@ func StartProcessMetricsSidecar(
 	buildID string,
 	language string,
 ) *http.Server {
-	// Create registry with process collector for worker PID
 	registry := prometheus.NewRegistry()
 	procCollector, err := metrics.NewProcessCollector(workerPID)
 	if err != nil {
@@ -41,7 +40,6 @@ func StartProcessMetricsSidecar(
 	}
 	registry.MustRegister(procCollector)
 
-	// HTTP handler for /metrics and /info
 	handler := http.NewServeMux()
 	handler.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
 	handler.HandleFunc("/info", func(w http.ResponseWriter, r *http.Request) {
@@ -53,7 +51,6 @@ func StartProcessMetricsSidecar(
 		})
 	})
 
-	// Start server
 	server := &http.Server{Addr: address, Handler: handler}
 	listener, err := net.Listen("tcp", address)
 	if err != nil {
--- a/cmd/clioptions/metrics_test.go
+++ b/cmd/clioptions/metrics_test.go
@@ -3,6 +3,7 @@ package clioptions
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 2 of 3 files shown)


cmd/clioptions/metrics.go
```diff
@@ -13,15 +14,78 @@ import (
...
+	handler.HandleFunc("/info", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(InfoResponse{
+			SDKVersion: sdkVersion,
+			BuildID:    buildID,
+			Language:   language,
+		})
+	})
+
+	// Start server
```
Sushisource: Boring comment (so are the other ones inside the func, really)

cmd/clioptions/metrics.go
```diff
@@ -13,15 +14,78 @@ import (
...
-	PrometheusHandlerPath     string
-	prometheusInstanceOptions PrometheusInstanceFlags
+	PrometheusHandlerPath string
+	// Address for separate process metrics server (CPU/memory only).
+	// If empty, process metrics will not be served separately.
+	WorkerProcessMetricsAddress string
+	// MetricsVersionTag is the SDK version/ref to report in metrics.
+	// This is used by the sidecar's /info endpoint and is NOT passed to the worker.
+	// If empty, falls back to the --version flag value.
+	MetricsVersionTag           string
```
Sushisource, line 84: Why would we want this to be different from version?

cmd/clioptions/metrics_test.go
```diff
@@ -0,0 +1,184 @@
...
+		fetchMetrics(t, "http://localhost:19091/metrics")
+
+		// Shutdown
+		err := metrics.Shutdown(ctx, logger, "", "")
+		if err != nil {
+			t.Errorf("Shutdown returned error: %v", err)
+		}
+
+		// Give time for shutdown
+		time.Sleep(100 * time.Millisecond)
```
Sushisource: These should instead just use an `eventually` kind of thing on the subsequent assertion (also more boring comments)

cmd/clioptions/metrics_test.go
```diff
@@ -0,0 +1,184 @@
...
+		fetchMetrics(t, "http://localhost:19094/metrics")
+
+		// Shutdown
+		err := sidecar.Shutdown(ctx)
+		if err != nil {
+			t.Errorf("Shutdown returned error: %v", err)
+		}
+
+		// Give time for shutdown
+		time.Sleep(100 * time.Millisecond)
```
Sushisource: Same as other comment

Sushisource (commented):
This mostly makes sense to me - an update to the README describing this architecture would be a good idea

cmd/clioptions/metrics.go
THardy98: `--version` can be a path to a local repo (which is what we use in our cloud worker testing), we'd prefer a git commit hash or something that actually identifies the version

cmd/clioptions/metrics.go
THardy98: removed

cmd/clioptions/metrics_test.go
THardy98: replaced with a util called `eventually` (assertion with a timeout)

THardy98:
updated README to include some info on the metrics sidecar

cmd/clioptions/metrics.go
Sushisource: Ah. I see.

PR #290, opened 2026-02-18
Update TS SDK to 1.15.0

mjameswh:
## What changed

- Update TS SDK to 1.15.0
- Update Node, TypeScript, eslint, and more, to match now-preferred TS SDK's setup


omes.sln
```diff
@@ -0,0 +1,29 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
Sushisource: Don't think this should be checked in.

Using visual studio??? Hehe

PR #295, opened 2026-03-10
Nexus support for Fuzzer

stephanos:
## What was changed

1. Added support for Nexus operations to the fuzzer.
2. Turned off local activities by default.

## Why?

1. Increase test coverage.
2. Make fuzzer results more actionable.

## Checklist

```
go run ./cmd run-scenario-with-worker --scenario fuzzer --language go --embedded-server --iterations 1  --run-id nexus
```

```
go run ./cmd run-scenario-with-worker --scenario fuzzer --language python --embedded-server --iterations 1  --run-id nexus
```

### DIFF
--- a/dockerfiles/java.Dockerfile
+++ b/dockerfiles/java.Dockerfile
@@ -1,6 +1,6 @@
 # Build in a full featured container
 ARG TARGETARCH
-FROM --platform=linux/$TARGETARCH eclipse-temurin:11-jammy AS build
+FROM --platform=linux/$TARGETARCH eclipse-temurin:21-jammy AS build
 
 # Install protobuf compiler and dependencies
 RUN apt-get update \
@@ -45,7 +45,7 @@ RUN CGO_ENABLED=0 ./temporal-omes prepare-worker --language java --dir-name prep
 
 # Copy the CLI and prepared feature to a "run" container. Distroless isn't used here since we run
 # through Gradle and it's more annoying than it's worth to get its deps to line up
-FROM --platform=linux/$TARGETARCH eclipse-temurin:11
+FROM --platform=linux/$TARGETARCH eclipse-temurin:21
 ENV GRADLE_USER_HOME="/gradle"
 
 COPY --from=build /app/temporal-omes /app/temporal-omes
--- a/dockerfiles/typescript.Dockerfile
+++ b/dockerfiles/typescript.Dockerfile
@@ -37,6 +37,12 @@ ARG SDK_VERSION
 ARG SDK_DIR=.gitignore
 COPY ${SDK_DIR} ./repo
 
+# Read BUILD_CORE_RELEASE env var. This builds TS worker with core bridge in release mode.
+# This is only relevant if building from source (if using a published version, the worker is already
+# built in release mode).
+ARG BUILD_CORE_RELEASE=false
+ENV BUILD_CORE_RELEASE=${BUILD_CORE_RELEASE}
+
 # Copy the worker files
 COPY workers/proto ./workers/proto
 COPY workers/typescript ./workers/typescript
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.11.1
-	github.com/temporalio/features v0.0.0-20251218214540-e36ee9b474e2
+	github.com/temporalio/features v0.0.0-20260324215619-e5868d9ba03f
 	go.temporal.io/api v1.62.1
 	go.temporal.io/sdk v1.40.0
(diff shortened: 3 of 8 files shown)


loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -89,6 +95,7 @@ impl Default for GeneratorConfig {
             max_client_action_set_wait: Duration::from_secs(1),
             action_chances: Default::default(),
             max_initial_actions: 10,
+            local_activities: false,
```
stephanos, line 101: Correct me if I'm wrong, but it seems findings in local activities don't provide any value. But they do distract from real findings that are actionable. So disabling it by default seems to be the only reasonable option.

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -692,6 +708,57 @@ impl<'a> Arbitrary<'a> for ExecuteChildWorkflowAction {
...
+
+impl<'a> Arbitrary<'a> for ExecuteNexusOperation {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let endpoint = ARB_CONTEXT.with_borrow(|c| c.nexus_endpoint.clone());
+        let &operation = u.choose(&NEXUS_OPERATIONS)?;
+        let val = format!("nexus-test-{}", u.int_in_range(1..=1000)?);
+        let (input, expected_output) = (val.clone(), val);
+
+        // Randomly generate before_actions for echo-async operations
+        let before_actions = if operation == "echo-async" && u.ratio(1, 3)? {
```
stephanos: Not sure if/how to make `u.ratio(1, 3)` configurable.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -140,7 +140,7 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 	} else {
 		// Otherwise wait for return/error from signals
 		for {
-			retOrErrChan.Receive(ctx, retOrErr)
+			retOrErrChan.Receive(ctx, &retOrErr)
```
stephanos, line 143: I kept getting "Corrupted signal received" without this from the fuzzer

loadgen/kitchen-sink-gen/src/main.rs
THardy98: I think disabling by default is reasonable

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -601,6 +614,12 @@ impl<'a> Arbitrary<'a> for Action {
             action::Variant::UpsertSearchAttributes(u.arbitrary()?)
         } else if chances.nested_action_set(action_kind) {
             action::Variant::NestedActionSet(u.arbitrary()?)
+        } else if chances.nexus_operation(action_kind) {
+            if ARB_CONTEXT.with_borrow(|c| c.action_set_nest_level >= 1) {
+                action::Variant::Timer(u.arbitrary()?)
+            } else {
+                action::Variant::NexusOperation(u.arbitrary()?)
+            }
```
THardy98, line 626: Why is producing a nexus operation dependent on if the context has action set nest level >= 1 ?
They don't seem correlated to me

loadgen/kitchen-sink-gen/src/main.rs
THardy98: I haven't thought much about it. How useful/necessary would this be to you currently?

THardy98 (approved):
generally looks good to me, left a couple questions/comments

loadgen/kitchen-sink-gen/src/main.rs
stephanos: Right now it's too early to tell; but if I wanted to target this particular use case I'd be nice to make it more likely. There's no simple way to add this right now as `ActionChances` has to add up to 100% but this is for a _nested_ action inside Nexus. So we'd need a nesting config, too. Didn't want to touch that problem just yet.

loadgen/kitchen-sink-gen/src/main.rs
stephanos: So what this means is that if a Nexus operation is supposed to be generated _but_ we're already nested, create a timer instead. IDK if there's a better way to do this.

The reason we don't want another Nexus op is that's too much nesting. I've had no limit on it before and runs would timeout much more often.

But we also can't "turn off" Nexus ops since AFAICT then the expectation that one action is picked is not guaranteed anymore.

stephanos: Actually; let me instead just add a _sync_ Nexus Operation.

THardy98: This resolves the timeout issue?

PR #298, opened 2026-03-11
Simplify and Test Nexus in kitchensink

stephanos:
## What was changed

Streamlined Nexus kitchensink and add tests.

## Why?

Makes it more flexible; also precursor for fuzzing support to various workflow actions inside the operation handler.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/workers/go/kitchensink/kitchen_sink.go
+++ b/workers/go/kitchensink/kitchen_sink.go
@@ -604,6 +604,9 @@ func NexusHandlerWorkflow(ctx workflow.Context, input *kitchensink.NexusHandlerI
 
 // EchoSyncOperation returns the input synchronously without starting a workflow.
 var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Context, input *kitchensink.NexusHandlerInput, opts nexus.StartOperationOptions) (string, error) {
+	if len(input.BeforeActions) > 0 {
+		return "", nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "before_actions not supported in echo-sync")
+	}
 	return input.Input, nil
 })
 


loadgen/kitchen_sink_executor_test.go
```diff
@@ -902,20 +1019,38 @@ func testForSDK(
...
+	require.NoError(t, err, "Failed to create Nexus endpoint")
+
 	executor := &KitchenSinkExecutor{
 		TestInput: tc.testInput,
+		PrepareTestInput: func(_ context.Context, _ ScenarioInfo, input *TestInput) error {
+			if input.WorkflowInput != nil {
+				for _, actionSet := range input.WorkflowInput.InitialActions {
+					for _, action := range actionSet.Actions {
+						if nexusOp := action.GetNexusOperation(); nexusOp != nil && nexusOp.Endpoint == "" {
+							nexusOp.Endpoint = nexusEndpoint
```
stephanos, line 1040: Each test has its own worker, task queue and Nexus endpoint for isolation.

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -586,27 +590,26 @@ type ReturnOrErr struct {
...
-var EchoAsyncOperation = temporalnexus.NewWorkflowRunOperation("echo-async", EchoWorkflow, func(ctx context.Context, s string, opts nexus.StartOperationOptions) (client.StartWorkflowOptions, error) {
-	return client.StartWorkflowOptions{
-		ID: opts.RequestID,
-	}, nil
+// EchoSyncOperation returns the input synchronously without starting a workflow.
+var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Context, input *kitchensink.HandlerWorkflowInput, opts nexus.StartOperationOptions) (string, error) {
+	return input.Input, nil
 })
 
-var WaitForCancelOperation = temporalnexus.NewWorkflowRunOperation("wait-for-cancel", WaitForCancelWorkflow, func(ctx context.Context, _ nexus.NoValue, opts nexus.StartOperationOptions) (client.StartWorkflowOptions, error) {
```
stephanos, line 609: main change: removed this

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -586,27 +590,26 @@ type ReturnOrErr struct {
...
-	})
+func NexusHandlerWorkflow(ctx workflow.Context, input *kitchensink.HandlerWorkflowInput) (string, error) {
+	state := KSWorkflowState{
+		workflowState: &kitchensink.WorkflowState{},
+	}
+	for _, actionSet := range input.BeforeActions {
+		if _, err := state.handleActionSet(ctx, actionSet); err != nil {
+			return "", err
+		}
+	}
```
stephanos, line 601: main change: now handles workflow actions

loadgen/kitchen_sink_executor_test.go
```diff
@@ -849,6 +856,115 @@ func TestKitchenSink(t *testing.T) {
...
+								},
+							},
+						}),
+				},
+			},
+			historyMatcher: PartialHistoryMatcher(`
+				NexusOperationScheduled {"operation":"echo-sync"}
+				...
+				WorkflowExecutionCompleted`),
+			expectedUnsupportedErrs: nexusUnsupportedSDKs,
```
stephanos, line 966: ^ a couple of basic tests to ensure Nexus support works in kitchensink

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -479,7 +479,11 @@ func handleNexusOperation(ctx workflow.Context, nexusOp *kitchensink.ExecuteNexu
 	return withAwaitableChoiceCustom(ctx, func(ctx workflow.Context) workflow.NexusOperationFuture {
 		client := workflow.NewNexusClient(nexusOp.Endpoint, KitchenSinkServiceName)
 		nexusOptions := workflow.NexusOperationOptions{}
-		return client.ExecuteOperation(ctx, nexusOp.Operation, nexusOp.Input, nexusOptions)
+		input := &kitchensink.HandlerWorkflowInput{
+			Input:         nexusOp.Input,
+			BeforeActions: nexusOp.BeforeActions,
```
Quinn-With-Two-Ns, line 484: What if the handler isn't starting a workflow? In the near future we will have handlers that can do things like send updates or signals.

stephanos: @Quinn-With-Two-Ns Good point; the `HandlerWorkflowInput` name is too restrictive; I renamed it to `NexusHandlerInput` now.

Then there's the fact that it's typed as `ActionSet`. That's fairly open; e.g. it already includes a `SendSignal` variant (which requires a `WorkflowId`). I think we'll need to make some adjustments to the kitchensink model once we have that capability. I'm hesitant to make those now before we have/know more. I've added an error there now to make sure it doesn't fail silently until then.

WDYT?

Quinn-With-Two-Ns: Yeah that makes sense. I think we will need to iterate on this when we add more operation types like async update, but we can cross that bridge when we get to it.

Quinn-With-Two-Ns (approved):

PR #301, opened 2026-03-18
Add project executor framework for workflow load testing

THardy98:
## Motivation

Omes today tests Temporal SDKs through **scenarios** — predefined load patterns that drive a generic KitchenSink worker using protobuf action sequences. This works well for cross-SDK conformance testing, but makes it difficult to test **real workflows** written in native Go (or any other language). The protobuf-action model is indirect: you describe *what* the workflow should do in protobuf, and a generic worker interprets it. You never actually test the workflow code that a team would write in practice.

**Project tests** are a new paradigm for omes. Instead of describing workflows indirectly, each project is a self-contained Go module with its own workflows, activities, worker, and execution logic. The framework coordinates these projects via gRPC, making them buildable, runnable, and testable independently — while still plugging into omes' load generation and metrics infrastructure.

## Architecture

A project test is a Go binary that exposes two subcommands:

- `<project> worker` — starts a Temporal worker with the project's real workflows and activities
- `<project> project-server` — starts a gRPC server that accepts `Init` and `Execute` RPCs

The **harness** (`projecttests/go/harness/`) provides this framework. Projects register three callbacks — `RegisterWorker`, `OnInit`, `OnExecute` — and call `Run()`. The harness handles gRPC server lifecycle, worker management, client pooling, Prometheus metrics, and TLS configuration.

The test runner (CLI or integration test) spawns both processes, sends an `Init` RPC with connection details and config, then drives iterations via `Execute` RPCs. This keeps projects decoupled from the omes CLI while allowing omes to orchestrate load patterns (steady-rate, ebb-and-flow, saturation) against any project.

## What's included

**Two example projects** demonstrate the framework:
- `helloworld` — minimal (~100 LOC), shows the basic pattern
- `throughputstress` — more involved, it's effectively a port of the existing throughput stress scenario. It is similar in that it parses a config file to conditionally execute a workflow (local/remote activities, child workflows, self-queries, self-signals, workflow updates, retry scenarios, heartbeats, and continue-as-new, etc.).

**Note**: The throughput stress project acts as an example/demonstration of porting an existing scenario to a project. It's also the most commonly used scenario for our load testing, so is a valuable scenario to transition. You are not limited to using this project for throughput testing. Any project using the `steady-state` executor (i.e. the generic executor) can run throughput testing.

**CLI integration** adds `omes project` and `omes exec` commands for running project-based load tests with configurable executors, metrics collection, and post-run verification.

**Note**: the post-run verification is quite simply at the moment, basically a small collection/library of existing verification functions we use already

**Docker support** uses a two-layer image pattern: a base image with the Go toolchain and harness, plus thin per-project overlay images. A Docker Compose stack provides Temporal server + Prometheus for local testing with metrics.

**CI** added a `projecttests.yml` workflow with three jobs for simple regression testing (integration tests - builds and runs helloworld + throughputstress against a dev server, docker image build verification, and proto generation/lint consistency checks)

This PR also refactors the existing ebb-and-flow scenario to use shared internal utilities and adds Prometheus export/query support (`metrics/prom_export.go`, `metrics/prom_query.go`)


.github/workflows/projecttests.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+      - name: Run Go project tests
+        run: |
+          go test -v -race -timeout 10m ./projecttests/... 2>&1 | \
+            go run github.com/jstemmer/go-junit-report/v2@latest \
+            -set-exit-code -iocopy -out junit-projecttests-go.xml
```
Sushisource, line 27: This should probably just be two separate steps

.github/workflows/projecttests.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          docker build \
+            -f projecttests/dockerfiles/projecttest-go.Dockerfile \
+            --build-arg BASE_IMAGE=omes-projecttest-go-base:ci \
+            --build-arg TEST_PROJECT=projecttests/go/tests/helloworld \
+            -t omes-projecttest-helloworld:ci \
+            .
+      - name: Smoke test overlay image
+        run: |
+          docker run --rm omes-projecttest-helloworld:ci --help
```
Sushisource, line 62: Not sure I'd call that a smoke test?

cmd/cli/exec.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, cancel := withCancelOnInterrupt(cmd.Context())
+			defer cancel()
+			return r.run(ctx, args)
+		},
+	}
+
+	r.sdkOpts.AddCLIFlags(cmd.Flags())
+	cmd.Flags().Lookup("language").Usage = "Language to use for workflow tests (go only)"
```
Sushisource, line 46: It won't be Go only except currently though?

Also, seemingly the language flag is redundant, since the projecttests path includes the language

cmd/cli/exec.go
```diff
@@ -0,0 +1,129 @@
...
+	r.programOpts.AddFlags(cmd.Flags())
+	cmd.Flags().AddFlagSet(r.loggingOpts.FlagSet())
+	cmd.Flags().StringVar(&r.processMonitorAddr, "process-monitor-addr", "", "Address for process metrics sidecar (e.g. :9091)")
```
Sushisource, line 49: Should this be just a port number? Seems like it would not make sense to ever bind to anything other than 0.0.0.0?

cmd/cli/project.go
```diff
@@ -0,0 +1,380 @@
...
+		Use:   "project",
+		Short: "Run a workflow load test",
+		Long: `Build a test project, spawn a client process, run load, and exit.
+
+The client is always spawned. Use --spawn-worker to also spawn a
+local worker process; otherwise a worker should already be running.
+
+Examples:
+  omes project --language go --project-dir ./projecttests/go/tests/helloworld --spawn-worker --iterations 100
+  omes project --language go --project-dir ./projecttests/go/tests/helloworld --iterations 100`,
```
Sushisource, line 37: I'm not sure the names of the commands feel obvious. `exec` isn't in any way obviously related to `project`, but, it is. Maybe `exec` should be `project-exec` and this can be `project-run`.

Same comment as with `exec` that the language flag feels unnecessary here.

cmd/cli/project.go
```diff
@@ -0,0 +1,380 @@
...
+	fs := cmd.Flags()
+	r.sdkOpts.AddCLIFlags(fs)
+	fs.Lookup("language").Usage = "Language to use for workflow tests (go only)"
+	fs.AddFlagSet(r.clientOpts.FlagSet())
+	fs.AddFlagSet(r.loggingOpts.FlagSet())
+	r.loadOpts.AddCLIFlags(cmd)
+	r.programOpts.AddFlags(fs)
+	r.projectRunOpts.AddCLIFlags(fs)
+	fs.StringVar(&r.runID, "run-id", "", "Run ID (auto-generated if not provided)")
+	fs.StringVar(&r.runFamily, "run-family", "", "Human-readable identifier for grouping related runs")
```
Sushisource, line 93: How? Looks like this goes into the exported data? Should say so

cmd/cli/project.go
```diff
@@ -0,0 +1,380 @@
...
+	fs.StringVar(&r.taskQueue, "task-queue", "", "Task queue name (default: omes-<run-id>)")
+	fs.IntVar(&r.clientPort, "client-port", 0, "Port for local client HTTP server (0 = auto)")
```
Sushisource, line 95: Local client HTTP server of... what?

There's a lot of flags here, but not much information on how this all actually fits together. The whole setup of how projects work, the architecture, etc, could really use a brief readme with a little diagram or something.

cmd/cli/project.go
```diff
@@ -0,0 +1,380 @@
...
+	fs.StringVar(&r.executor, "executor", "", "projecttests executor to run")
+	fs.StringVar(&r.configPath, "config", "", "Path to JSON config file for project test")
+	fs.StringSliceVar(&r.options, "option", nil, "Additional options for the executor, in key=value format")
```
Sushisource, line 98: Apparently this also accepts some magic `@` syntax to read from a file, but that's not mentioned here.

cmd/clioptions/client.go
```diff
@@ -86,9 +87,43 @@ func (c *ClientOptions) MustDial(metrics *metrics.Metrics, logger *zap.SugaredLo
 	return client
 }
 
+// DialLite connects to a Temporal server with TLS and auth but no metrics or custom data converter.
+// Useful for lightweight clients (e.g., post-run verification queries).
+func (c *ClientOptions) DialLite(logger *zap.SugaredLogger) (client.Client, error) {
```
Sushisource, line 92: Why is this necessary? Just avoiding extra metrics? If so, a more obvious name like `DialNoMetrics` would be good

cmd/clioptions/load.go
```diff
@@ -0,0 +1,26 @@
+package clioptions
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+type LoadOptions struct {
```
Sushisource, line 9: AFAICT this file is only used in the one place. Why is it its own file?

internal/utils/net.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return 0, err
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	listener.Close()
+	return port, nil
+}
+
+// DefaultPort assigns a random available port to *p if it is 0.
+func DefaultPort(p *int) error {
```
Sushisource, line 22: This is... weird. Why does this exist? Why not just expose `FindAvailablePort`?

internal/verification/verification.go
```diff
@@ -0,0 +1,150 @@
+package verification
```
Sushisource, line 1: Why is this in internal? Looks like it's just used by the project runner? Why not just have it be private functions in there?

loadgen/projectexecutor/executor.go
```diff
@@ -0,0 +1,125 @@
...
+
+const (
+	defaultClientHost = "localhost"
+	// Default timeout waiting for project service to be ready.
+	defaultClientReadyTimeout = 3 * time.Second
+	// Interval between readiness checks.
+	defaultClientReadyCheckInterval = 100 * time.Millisecond
+
+	// steady-rate project executor (wrapper over generic executor)
+	SteadyRateProjectExecutor = "steady-rate"
```
Sushisource, line 24: Steady rate at what cadence? I assume this just means it does iterations one by one? A better name might just be `serial-iterations`

metrics/metrics.go
```diff
@@ -211,27 +217,65 @@ func NewProcessCollector(pid int) (*processCollector, error) {
...
 }
 
 func (c *processCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- c.cpuDesc
 	ch <- c.memDesc
+	ch <- c.memPercentDesc
+	ch <- c.numCPUsDesc
 }
 
 func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
```
Sushisource, line 235: Should probably describe in docstring how this is doing some add-up-all-the-children thing.

metrics/prom_export.go
```diff
@@ -0,0 +1,294 @@
...
+
+const (
+	// JobWorkerApp is the Prometheus scrape job name for worker SDK/application metrics.
+	JobWorkerApp = "omes_worker_app"
+	// JobWorkerProcess is the Prometheus scrape job name for worker process sidecar metrics.
+	JobWorkerProcess = "omes_worker_process"
+	// DefaultStepInterval is the default step interval to scrape Prometheus if none was provided
+	DefaultStepInterval = 5 * time.Second
+	// exportPipelineParquetBatchSize controls parquet flush size for this export path.
+	exportPipelineParquetBatchSize = 5000
```
Sushisource, line 26: Batch size in whats? Bytes? What would be the reasons to change this value?

metrics/prom_query.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+)
+
+// DefaultDerivedQueries returns a curated list of PromQL queries that precompute
+// worker/process metrics into Omni-friendly metric lines.
+func DefaultDerivedQueries() []PromQuery {
```
Sushisource, line 15: This is very duplicative of `buildMetricQueries`

projecttests/dockerfiles/README.md
```diff
@@ -0,0 +1,151 @@
+# projecttests Docker Setup
```
Sushisource, line 1: This file is fairly useful (if a bit overly verbose), but it's not in a very useful place. Per my overall review comments, this should be combined with some kind of higher level readme that explains the whole structure and how to use it.

projecttests/go/tests/helloworld/client.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/temporalio/omes/projecttests/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/temporal"
+)
+
+var pool = harness.NewClientPool()
+
+func clientMain(ctx context.Context, config *harness.Config) error {
+	c, err := pool.GetOrDial("default", config.ConnectionOptions)
```
Sushisource, line 15: IMO rather than expecting everyone to use this pool, it would make more sense to just provide the client as part of the harness execute signature.

projecttests/README.md
```diff
@@ -0,0 +1,57 @@
+# Project Tests
+
+Self-contained Go programs for testing Temporal workflows under load. Each project is an independent Go module that implements its own workflows, activities, and execution logic, coordinated via gRPC.
+
```
Sushisource, line 4: This is more like the overview I was looking for but still lacks a real overview of the architecture, and still is written like Go will forever be the only language. Combining the docker readme with this, giving an architectural overview, and at least linking to it from the top level README would go a long way.

cmd/cli/project.go
```diff
@@ -0,0 +1,380 @@
...
+	fs.AddFlagSet(r.clientOpts.FlagSet())
+	fs.AddFlagSet(r.loggingOpts.FlagSet())
+	r.loadOpts.AddCLIFlags(cmd)
+	r.programOpts.AddFlags(fs)
+	r.projectRunOpts.AddCLIFlags(fs)
+	fs.StringVar(&r.runID, "run-id", "", "Run ID (auto-generated if not provided)")
+	fs.StringVar(&r.runFamily, "run-family", "", "Human-readable identifier for grouping related runs")
+	fs.StringVar(&r.taskQueue, "task-queue", "", "Task queue name (default: omes-<run-id>)")
+	fs.IntVar(&r.clientPort, "client-port", 0, "Port for local client HTTP server (0 = auto)")
+	fs.StringVar(&r.executor, "executor", "", "projecttests executor to run")
```
Sushisource, line 96: Seems like this should have a default?

PR #306, opened 2026-03-23
Fix process metrics to aggregate across child processes

THardy98:
The `sdkbuild` package launches Python workers via `uv run python -m main`,
so the monitored PID is the uv wrapper, not the actual Python worker.
Aggregate CPU and memory metrics across the process tree to capture the
full resource usage.

### DIFF
--- a/workers/dotnet/Temporalio.Omes/ClientActionsExecutor.cs
+++ b/workers/dotnet/Temporalio.Omes/ClientActionsExecutor.cs
@@ -16,7 +16,7 @@ public class ClientActionsExecutor
 
     public string? WorkflowId { get; set; }
 
-    public ClientActionsExecutor(ITemporalClient client, string workflowId, string taskQueue, bool errOnUnimplemented = false)
+    public ClientActionsExecutor(ITemporalClient client, string? workflowId, string taskQueue, bool errOnUnimplemented = false)
     {
         _client = client;
         WorkflowId = workflowId;


metrics/metrics.go
```diff
@@ -217,21 +222,40 @@ func NewProcessCollector(pid int) (*processCollector, error) {
...
 func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
-	// CPU percent
-	if cpuPercent, err := c.process.Percent(0); err == nil {
-		ch <- prometheus.MustNewConstMetric(c.cpuDesc, prometheus.GaugeValue, cpuPercent)
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	// Discover child processes once. Retries on each scrape until found,
+	// since the child (e.g. python under uv) may not exist yet at startup.
+	if !c.childrenDiscovered {
```
Sushisource, line 234: What if more children are spawned later? Seemingly this needs to be re-done every time... or at least periodically.

metrics/metrics.go
```diff
@@ -217,21 +222,40 @@ func NewProcessCollector(pid int) (*processCollector, error) {
...
 	ch <- c.memDesc
+	ch <- c.memPercentDesc
 }
 
 func (c *processCollector) Collect(ch chan<- prometheus.Metric) {
-	// CPU percent
-	if cpuPercent, err := c.process.Percent(0); err == nil {
-		ch <- prometheus.MustNewConstMetric(c.cpuDesc, prometheus.GaugeValue, cpuPercent)
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
Sushisource, line 230: Lock could be a lot tighter here I think, around just the write portion. Doesn't need to be held while calling `Children()`

metrics/metrics.go
THardy98: I'd originally written this to collect children processes on every `Collect` call, but I think I expect this to be pretty stable. This change is intended purely for the sdkbuild worker invocation command (specifically for Python), I don't expect the number of child processes spawned to change over time.

I'm not sure if we'll have other use cases for this, but I don't mind going to what I had before.

Sushisource: This isn't for any worker running generally? Why do we even care about collecting metrics for just the sdkbuild part?

metrics/metrics.go
THardy98: I'm not sure we can make it tighter.

It needs to be held during the `!c.childrenDiscovered` check, which includes the `Children()` call. Otherwise, 2 goroutines could both enter the outer `!c.childrenDiscovered` conditional, and both append (potentially the same) children processes.

It also needs to be held while iterating over `c.procs` later otherwise, a goroutine could potentially add to `c.procs` (assuming the lock is only around the previous conditional) while another, that has already released the lock, is iterating over it. You'd have an append during iteration.

Sushisource: Ah yeah fair enough

metrics/metrics.go
Sushisource: Got the context from slack

THardy98: We don't. It's just that when you build an sdkbuild program, the result is a command you can call to run that program. That command directly runs the worker, so the resultant process is the worker process. Except Python, the command it runs is:
```
// NewCommand makes a new uv command. The first argument needs to be the
// name of the module.
func (p *PythonProgram) NewCommand(ctx context.Context, args ...string) (*exec.Cmd, error) {
	args = append([]string{"run", "python", "-m"}, args...)
	cmd := exec.CommandContext(ctx, "uv", args...)
	cmd.Dir = p.dir
	setupCommandIO(cmd, nil, nil)
	return cmd, nil
}
```
It's a uv command that then runs the python command that starts the worker. So when we track the process metrics of the command produced by sdkbuild:
`cmd.Process.Pid`
it tracks the uv command, not the worker. The child process of the uv command is the actual python worker process. It was either track the child process or make the change in sdkbuild to call the python binary directly. I opted for this because what exists in sdkbuild for Python seems pretty normal (as in we use uv to manage the python build already and using uv run to start the python program is standard)

PR #310, opened 2026-03-24
Add cargo fmt to lint-and-format toolchain

stephanos:
## Why?

All workers already have enforced formatting.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/dev/clean.go
+++ b/cmd/dev/clean.go
@@ -56,7 +56,7 @@ Examples:
 	return cmd
 }
 
-func cleanLanguage(ctx context.Context, target string) error {
+func cleanTarget(ctx context.Context, target string) error {
 	targetDir, err := getTargetDir(target)
 	if err != nil {
 		return err
@@ -95,7 +95,7 @@ func runClean(ctx context.Context, targets []string, cleanAll bool) error {
 	}
 
 	for _, target := range targets {
-		if err := cleanLanguage(ctx, target); err != nil {
+		if err := cleanTarget(ctx, target); err != nil {
 			if cleanAll {
 				fmt.Printf("Warning: failed to clean %s: %v\n", target, err)
 			} else {


loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -5,13 +5,16 @@ use crate::protos::temporal::{
     api::common::v1::{Memo, Payload, Payloads},
     omes::kitchen_sink::{
         action, awaitable_choice, client_action, do_actions_update, do_query, do_signal,
```
stephanos, line 7: All changes here are from running the fmt

.github/workflows/ci.yml
```diff
@@ -45,18 +45,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [go, java, python, typescript, dotnet]
+        target: [go, java, python, typescript, dotnet, kitchensink-gen]
```
stephanos, line 48: `target` captures intent a bit better now IMO

cmd/dev/lint_and_format.go
```diff
@@ -162,6 +166,29 @@ func lintAndFormatTypescriptWorker(ctx context.Context, workerDir string) error
...
+	kitchenSinkGenDir := getKitchenSinkGenDir(repoDir)
+
+	fmt.Println("Formatting Rust kitchensink-gen...")
+	if err := runCommandInDir(ctx, kitchenSinkGenDir, "cargo", "fmt"); err != nil {
+		return err
+	}
+
+	fmt.Println("✅ Rust kitchensink-gen lint-and-format completed successfully!")
+	return nil
+}
```
stephanos, line 185: ^ this is the main change

.github/workflows/ci.yml
stephanos: (I couldn't help myself)

cmd/dev/install.go
```diff
@@ -226,18 +227,12 @@ func installProtoc(ctx context.Context) error {
 	return nil
 }
 
-func installUv(ctx context.Context) error {
-	version, err := getVersion("uv")
-	if err != nil {
-		return err
-	}
-	return installViaMise(ctx, "uv", version)
-}
```
stephanos, line 235: ^ this was dead code

cmd/dev/clean.go
```diff
@@ -12,65 +12,70 @@ import (
...
 	}
 
 	return cmd
 }
 
-func cleanLanguage(ctx context.Context, language, repoDir string) error {
-	workerDir := filepath.Join(repoDir, "workers", language)
-	if _, err := os.Stat(workerDir); err != nil {
-		return fmt.Errorf("worker directory does not exist: %s", workerDir)
+func cleanLanguage(ctx context.Context, target string) error {
```
THardy98: If we're going to change the `language` param to ` target` we might as well change the function name to `cleanTarget` as well

cmd/dev/install.go
```diff
@@ -199,7 +199,8 @@ func installRust(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	return installViaMise(ctx, "rust", version)
+	// Use --profile default to ensure rustfmt is included
+	return installViaMise(ctx, "rust", version, "--profile", "default")
```
THardy98, line 203: Why do we need `--profile default` ?

cmd/dev/install.go
```diff
@@ -226,18 +227,12 @@ func installProtoc(ctx context.Context) error {
...
-	}
-	return installViaMise(ctx, "uv", version)
-}
-
-func installViaMise(ctx context.Context, tool, version string) error {
+func installViaMise(ctx context.Context, tool, version string, extraArgs ...string) error {
 	fmt.Println("Installing", tool, version)
 
-	cmd := exec.CommandContext(ctx, "mise", "use", fmt.Sprintf("%s@%s", tool, version))
+	args := []string{"use", "-g", fmt.Sprintf("%s@%s", tool, version)}
```
THardy98: `-g` global install ?

cmd/dev/test.go
```diff
@@ -58,6 +58,12 @@ func runTestWorker(ctx context.Context, language string) error {
 		return err
 	}
 
+	// kitchensink-gen cannot be tested as a worker.
+	if language == "kitchensink-gen" {
+		fmt.Println("Skipping test for generator target:", language)
```
THardy98, line 63: Somewhat redundant print, but I have no real opinions on it

cmd/dev/install.go
stephanos: it says in the comment ;)

without it it doens't seem to install `rustfmt`. I had my CI check fail.

cmd/dev/install.go
stephanos: oh; good catch. Let me remove that again.

cmd/dev/test.go
stephanos: yeah just want to make sure it doesn't silently do nothing

cmd/dev/install.go
stephanos: ie despite the name the default profile is `minimal`

(see https://mise.jdx.dev/lang/rust.html)

PR #317, opened 2026-03-29
Add a `project` scenario

THardy98:
## What was changed
Added a new `project` scenario that allows running load against arbitrary projects.

The idea is to allow developers to write self-contained feature tests (e.g. for Nexus operations) called projects, independently of the kitchen-sink scenario pipeline. Some advantages over the existing scenarios:
- can specify client and worker configs (no need to plumb flags through to config)
- can write regular workflow code, no DSL
- abstracted from the omes framework (don't need to write a scenario)
- can independently version different projects (i.e. can run one project on an older SDK version than another)

Enabling this is a simple harness comprised of a gRPC API and some hook functions:
- `Init` RPC: one-time project initialization, provides connection parameters for the project to create clients
- `Execute` RPC: runs at each iteration of the executor, this is where load is driven
- `InitFunc` hook: optional hook that allows for one-time, project-specific initialization (i.e. registering Nexus endpoints), triggered during the `Init` RPC
- `ExecuteFunc` hook: required hook, this runs at each `Execute` RPC call, this determines what load you want to drive
- `RegisterClient` hook: registers a client to be used by your project
- `RegisterWorker` hook: register a a worker to be used by your project

Currently, harness implementations are limited to Go and .NET.
They include sample tests:
- Go: `HelloWorld`
- .NET: `HelloWorld`, `NexusSimpleWorkflow`

There is also docker support via `go-project.Dockerfile` and `dotnet-project.Dockerfile`.

Like existing worker builds, builds for projects also support "build-from-source" (i.e. `--version <path to local repo>`), so you can test against local changes.

See the `README.md` for details on how to use this scenario.

**Note**
- Currently, this is limited to the default generic executor (i.e. named `SteadyStateExecutor`), but still effective for throughput testing.

### DIFF
--- a/scenarios/project/build.go
+++ b/scenarios/project/build.go
@@ -130,6 +130,7 @@ func buildDotNet(ctx context.Context, opts BuildOptions) (sdkbuild.Program, erro
 		BaseDir:         baseDir,
 		DirName:         dirName,
 		Version:         opts.Version,
+		Configuration:   "Release",
 		ProgramContents: fmt.Sprintf("return await %s.RunAsync(args);", projectName),
 		CsprojContents: `<Project Sdk="Microsoft.NET.Sdk">
 			<PropertyGroup>


THardy98:
RFR

PR #318, opened 2026-03-30
Add a describe scenario to throughputstress test

prathyushpv:
## What was changed
Bench-go has a scenario that calls DescribeWorkflow api for each workflow execution. Adding that scenario to omes.

## Why?
This would give us test coverage of this API in the load test. Workflow's mutable state should already be in workflow cache. So it should be a light operation on server.


THardy98 (changes requested):
I'm not familiar with how bench-go tests this but it's not immediately clear to me why we're adding the describe call to throughput stress. If the goal is to stress the describe API, it may make more sense to put this as a separate scenario. Imo, it feels like a mix of concerns (we are changing the general test case, to accommodate a niche case).

I also looks like we aren't doing any assertions/verifications on it, so it's just a describe call ?

prathyushpv:
throughput stress is something we use for release validation. The idea is to stress most code paths during this test. That is why we have all these different features being tested in this scenario. The goal is not to stress describe API. It is just to execute this path when other parts of the system is also under stress.

THardy98:
> throughput stress is something we use for release validation. The idea is to stress most code paths during this test. That is why we have all these different features being tested in this scenario. The goal is not to stress describe API. It is just to execute this path when other parts of the system is also under stress.

This makes sense to me, but my concern is that it affects existing throughput measurements by default, without anyway of turning it off. It may make more sense for this to be a scenario option, which would allow you to A/B  describe as well, wdyt?

prathyushpv:
That makes sense! I have added a scenario flag to turn it off. Kept the default value to true. I think our cicd pipelines are the only consumer of this throughputstress scenario. Do you know if it is used anywhere else so that I should set the default value of this input to false?

THardy98:
> That makes sense! I have added a scenario flag to turn it off. Kept the default value to true. I think our cicd pipelines are the only consumer of this throughputstress scenario. Do you know if it is used anywhere else so that I should set the default value of this input to false?

SDK team usese throughputstress to perf benchmark workers, if it's not too much hassle, I'd default it to `false`

prathyushpv:
Oh I didn't know it was used before. Changed it to false

THardy98 (approved):

PR #319, opened 2026-03-31
Fixes: Java/Typescript source builds, .NET/Ruby kitchensink errors

THardy98:
## What was changed
- Fixes for Java and Typescript failures that occur only when building from source.
(features repo bump is for upstream Typescript/Java build fixes when building from source)
- Fix to surface .NET task failures, previously failed silently (masked a search attribute error with child workflows - fixed upstream in .NET SDK)
- Fix Ruby kitchen sink implementation, caught by added CI step below
- added CI step to smoke test kitchen sink implementations under throughput_stress

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
             --run-id ${{ github.run_id }} \
             --connect-timeout 1m \
             --iterations 5
-      - name: Run throughput_stress against ${{ matrix.sdk }} worker
+      - name: Smoke test throughput_stress to confirm build is usable against ${{ matrix.sdk }} worker
         timeout-minutes: 10
         run: |
           go run ./cmd run-scenario \


.github/workflows/ci.yml
```diff
@@ -128,6 +128,21 @@ jobs:
             --run-id ${{ github.run_id }} \
             --connect-timeout 1m \
             --iterations 5
+      - name: Run throughput_stress against ${{ matrix.sdk }} worker
```
tconley1428: Is this intended to be kept as part of the build worker step? Or was it for validation?

dockerfiles/java.Dockerfile
```diff
@@ -46,10 +46,12 @@ RUN CGO_ENABLED=0 ./temporal-omes prepare-worker --language java --dir-name prep
 # Copy the CLI and prepared feature to a "run" container. Distroless isn't used here since we run
 # through Gradle and it's more annoying than it's worth to get its deps to line up
 FROM --platform=linux/$TARGETARCH eclipse-temurin:21
+RUN apt-get update && apt-get install --no-install-recommends --assume-yes git && rm -rf /var/lib/apt/lists/*
```
mjameswh, line 49: Just curious, why?

.github/workflows/ci.yml
THardy98: Kept - it's a smoke test that validates worker kitchen sink implementations

PR #328, opened 2026-04-07
Standalone Python app harness, `project-server` support

THardy98:
## What was changed
Make Python harness standalone (it's own python project)

Adds `project-server` support to the Python app harness. 

The larger `worker.py` diff is mostly code movement:
- shared client/config setup moved to `client.py`
- logging helper moved to `helpers.py`

`project-server` behavior lives in `project.py`.
Added documentation to `harness/__init__.py` describing harness usage.

Slightly changed the worker factory shape so the client is passed explicitly, matching the style used by the project handlers.

Added a small testsuite for the harness in `python/tests`, covers:
- `project-server` init/execute behavior
- worker wiring touched by this refactor.

Changes to the install/lint/test commands to account for harness linting/testing

## Why?
Enable project-based tests


THardy98:
branched off: https://github.com/temporalio/omes/pull/327

EDIT: merged, branched off `main`

Sushisource (approved):
This looks good to me but it also looks like all the tests are operating on mocked stuff. It'd be good to have an integration test that is doing the whole thing with a simple `App`.

PR #336, opened 2026-04-22
Adding scenario for serverless workers

gcristea-temporal:
## What was changed

Added new scenario for support scale testing for serverless workers through Omes.
## Why?
The existing scenarios don't address specific edge cases that can arise for serverless workers. Specifically,
* fast scaling capabilities — scale up from zero quickly
* mixed activity types with different resource requirements
* serverless timeout sensitivity

## Checklist

1. Closes: **none**

2. How was this tested: **ran tests locally.**

3. Any docs updates needed?
**No docs update are required.**


THardy98 (approved):
looks reasonable to me!

scenarios/serverless_burst_test.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			ScenarioOptions: map[string]string{
+				sbFastActivityCountFlag: "5",
+				sbSlowActivityCountFlag: "2",
+				sbSlowActivityDuration:  "2s",
+			},
+		}, clioptions.LangGo)
+		require.NoError(t, err)
+		require.EqualValues(t, 3, completed.Load())
+	})
```
jaypipes, line 45: this may just be me not knowing Omes enough, but when I look at the test cases in this serverles_burst_test.go file, I have absolutely no idea what is being asserted.

Sure, we're requiring that the executor run without error and completed an expected number of executions. However, I fail to see where anything about the *serverless worker execution* is being asserted. In the comments in the serverless_burst.go file, I find this:

> Pair this scenario with a short or zero pre-burst idle period
> (provided externally, e.g. by running with scale-to-zero in between) to observe
> how quickly the WCI can invoke enough Lambda workers to drain the resulting backlog.

However, I don't see any assertions in the test cases that look at expected backlog size values or some sort of WCI-specific metric to compare an expected value or range of values?

jaypipes (commented):
@gcristea-temporal unfortunately I have no idea what these tests are asserting :)

scenarios/serverless_burst_test.go
gcristea-temporal: You are right that there isn't anything specific for serverlerss worker execution. The execution side will follow: new worker and worker image.

The tests in this change are unit tests for the new scenario. They validate that the interaction between the scenario and the rest of the framework works as intended.

jaypipes: ack. in that case, :+1: :)

jaypipes (approved):

PR #338, opened 2026-04-24
Add support for Worker Deployment Versions (Go worker)

veeral-patel:
## Summary

This PR adds `--deployment-name`, `--deployment-build-id`, and `--default-versioning-behavior` flags to the Omes worker CLI.

## Context

Omes currently only supports the legacy Build ID based worker versioning scheme for the Go worker. This PR adds flags for Deployment Versions, and initializes the worker accordingly.

We continue to support the legacy Build ID based path, but I've added a comment stating that those flags are deprecated. We can phase out this path entirely in the near future.

Out of scope: Python / TypeScript / Java / .NET worker wiring

## Files changed

| File | Change |
|---|---|
| `cmd/clioptions/worker.go` | Added `DeploymentName`, `DeploymentBuildID`, `DefaultVersioningBehavior` fields on `WorkerOptions` and registered `--deployment-name`, `--deployment-build-id`, `--default-versioning-behavior` flags. Clarified `--build-id` description as legacy and mutually exclusive with `--deployment-name`. |
| `workers/go/worker/worker.go` | Hoisted `worker.Options` construction out of the per-task-queue goroutine. When `--deployment-name` is set, builds `worker.DeploymentOptions` from `DeploymentName` + `DeploymentBuildID` + resolved `DefaultVersioningBehavior`; otherwise keeps the legacy `BuildID` + `UseBuildIDForVersioning` path (now commented as deprecated). Added `parseVersioningBehavior` helper (accepts `pinned`, `auto-upgrade`; defaults to `auto-upgrade` on empty). Added up-front validation for the three mutually-exclusive / required-together flag combinations. |

## Test plan

- [x] `go build ./...` passes
- [x] Baseline scenario (no versioning flags) completes 3 iterations in ~2s with embedded server
- [x] Deployment-versioned worker starts cleanly and logs `"Started Worker" ... "BuildID": "v1"` under `--deployment-name omes-test --deployment-build-id v1 --default-versioning-behavior auto-upgrade`
- [x] `--build-id` + `--deployment-name` → fatal `--build-id and --deployment-name are mutually exclusive; use --deployment-build-id with --deployment-name`
- [x] `--deployment-name` without `--deployment-build-id` → fatal `--deployment-build-id is required when --deployment-name is set`
- [x] `--deployment-build-id` without `--deployment-name` → fatal `--deployment-build-id requires --deployment-name`
- [x] `--default-versioning-behavior bogus` → fatal `invalid --default-versioning-behavior "bogus" (expected pinned or auto-upgrade)`
- [x] Legacy `--build-id` path compiles and reaches the deprecated code branch (runtime end-to-end depends on task-queue Build-ID assignment rules, unchanged from prior behavior)

### DIFF
--- a/cmd/clioptions/worker.go
+++ b/cmd/clioptions/worker.go
@@ -8,6 +8,7 @@ import (
 type WorkerOptions struct {
 	BuildID                      string
 	DeploymentName               string
+	DeploymentBuildID            string
 	DefaultVersioningBehavior    string
 	MaxConcurrentActivityPollers int
 	MaxConcurrentWorkflowPollers int
@@ -27,8 +28,9 @@ func (m *WorkerOptions) FlagSet() *pflag.FlagSet {
 		return m.fs
 	}
 	m.fs = pflag.NewFlagSet("worker_options", pflag.ExitOnError)
-	m.fs.StringVar(&m.BuildID, "build-id", "", "Build ID")
-	m.fs.StringVar(&m.DeploymentName, "deployment-name", "", "Worker Deployment name. When set, enables Worker Deployment Versioning and must be combined with --build-id")
+	m.fs.StringVar(&m.BuildID, "build-id", "", "Build ID for legacy Build-ID-based worker versioning. Mutually exclusive with --deployment-name")
+	m.fs.StringVar(&m.DeploymentName, "deployment-name", "", "Worker Deployment name. When set, enables Worker Deployment Versioning and must be combined with --deployment-build-id")
+	m.fs.StringVar(&m.DeploymentBuildID, "deployment-build-id", "", "Build ID within the Worker Deployment. Required when --deployment-name is set")
 	m.fs.StringVar(&m.DefaultVersioningBehavior, "default-versioning-behavior", "", "Default versioning behavior for workflows that don't set one at registration. One of: pinned, auto-upgrade. Defaults to auto-upgrade when --deployment-name is set")
 	m.fs.IntVar(&m.MaxConcurrentActivityPollers, "worker-max-concurrent-activity-pollers", 0, "Max concurrent activity pollers")
 	m.fs.IntVar(&m.MaxConcurrentWorkflowPollers, "worker-max-concurrent-workflow-pollers", 0, "Max concurrent workflow pollers")
--- a/cmd/clioptions/worker.go
+++ b/cmd/clioptions/worker.go
@@ -7,6 +7,8 @@ import (
 // WorkerOptions for setting up worker parameters
 type WorkerOptions struct {
 	BuildID                      string
+	DeploymentName               string
+	DefaultVersioningBehavior    string
 	MaxConcurrentActivityPollers int
 	MaxConcurrentWorkflowPollers int
 	ActivityPollerAutoscaleMax   int // overrides MaxConcurrentActivityPollers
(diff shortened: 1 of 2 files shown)


workers/go/worker/worker.go
```diff
@@ -68,7 +68,52 @@ func makePollerBehavior(simple, auto int) worker.PollerBehavior {
...
+			UseVersioning: true,
+			Version: worker.WorkerDeploymentVersion{
+				DeploymentName: options.DeploymentName,
+				BuildID:        options.BuildID,
+			},
+			DefaultVersioningBehavior: behavior,
+		}
+	} else if options.BuildID != "" {
+		workerOpts.BuildID = options.BuildID
+		workerOpts.UseBuildIDForVersioning = true
```
carlydf, line 123: I think these fields are deprecated, so IMO we just should stop using them. It's ok to break existing users of this

carlydf: hm I guess KEDA is an "external user" of this field. We should mark them deprecated in Omes, and later remove them when we remove the server-side API support

veeral-patel: Modified the PR so we now add new flags for --deployment-name and --deployment-build-id. I also added a comment to the existing BuildID and UseBuildIDForVersioning flags stating they're deprecated.

workers/go/worker/worker.go
```diff
@@ -68,7 +68,60 @@ func makePollerBehavior(simple, auto int) worker.PollerBehavior {
...
+	default:
+		return 0, fmt.Errorf("invalid --default-versioning-behavior %q (expected pinned or auto-upgrade)", s)
+	}
+}
+
 func runWorkers(client client.Client, taskQueues []string, options clioptions.WorkerOptions) error {
+	workerOpts := worker.Options{
+		MaxConcurrentActivityExecutionSize:     options.MaxConcurrentActivities,
+		MaxConcurrentWorkflowTaskExecutionSize: options.MaxConcurrentWorkflowTasks,
+		ActivityTaskPollerBehavior: makePollerBehavior(
```
carlydf, line 86: nit: lack of aligned indents makes this harder to read imo (usually IDEs do the indent alignment, and AI-generated code does not, maybe it's "go fmt" not sure?)

workers/go/worker/worker.go
carlydf: looks good, ty!

carlydf (approved):

cmd/clioptions/worker.go
```diff
@@ -25,7 +28,10 @@ func (m *WorkerOptions) FlagSet() *pflag.FlagSet {
 		return m.fs
 	}
 	m.fs = pflag.NewFlagSet("worker_options", pflag.ExitOnError)
-	m.fs.StringVar(&m.BuildID, "build-id", "", "Build ID")
+	m.fs.StringVar(&m.BuildID, "build-id", "", "Build ID for legacy Build-ID-based worker versioning. Mutually exclusive with --deployment-name")
```
THardy98: might be worth explicitly marking this as deprecated (especially given that `deployment-build-id` exists)

cmd/clioptions/worker.go
```diff
@@ -25,7 +28,10 @@ func (m *WorkerOptions) FlagSet() *pflag.FlagSet {
...
+	m.fs.StringVar(&m.DeploymentName, "deployment-name", "", "Worker Deployment name. When set, enables Worker Deployment Versioning and must be combined with --deployment-build-id")
+	m.fs.StringVar(&m.DeploymentBuildID, "deployment-build-id", "", "Build ID within the Worker Deployment. Required when --deployment-name is set")
+	m.fs.StringVar(&m.DefaultVersioningBehavior, "default-versioning-behavior", "", "Default versioning behavior for workflows that don't set one at registration. One of: pinned, auto-upgrade. Defaults to auto-upgrade when --deployment-name is set")
```
THardy98, line 34: do we want a default value here? (not sure if this is desirable, just a thought. Deferring to yall, feel free to ignore)

THardy98 (approved):
LGTM, couple nits worth addressing

cmd/clioptions/worker.go
veeral-patel: @THardy98 No, I don't think there's a reasonable default we could set for the deployment name and build ID! But thank you for flagging!

cmd/clioptions/worker.go
veeral-patel: Done, I've updated the CLI arg description for `build-id` to mention it's deprecated.

workers/go/worker/worker.go
veeral-patel: @carlydf I checked the files I modified in this PR with `go fmt` and they all passed the `go fmt` check!

PR #339, opened 2026-04-27
Standalone Nexus support

stephanos:
## What was changed

Add support for testing Standalone Nexus.


loadgen/kitchen_sink_executor_test.go
```diff
@@ -1044,6 +1069,15 @@ func testForSDK(
...
 						}
+						if clientSeq := action.GetExecActivity().GetClient().GetClientSequence(); clientSeq != nil {
+							for _, cas := range clientSeq.ActionSets {
+								for _, ca := range cas.Actions {
+									if sno := ca.GetDoStandaloneNexusOperation(); sno != nil && sno.Endpoint == "" {
+										sno.Endpoint = nexusEndpoint
+									}
+								}
+							}
+						}
```
stephanos, line 1120: The endpoint isn't available at the time the `testInput` is created and we therefore have to patch it here. We could change the test input to come from a closure and pass in the endpoint there; but not sure that's worth it.

.gitignore
```diff
@@ -8,6 +8,7 @@ mise.toml
 # Ignore Gradle project-specific cache directory
 .gradle
 .devserver/
+.devserver.lock
```
stephanos, line 11: forgot to add that in https://github.com/temporalio/omes/commit/6b5f7e575e8e1a48ebcf4d7a0773029cf70f2eac

go.sum
```diff
@@ -124,8 +124,12 @@ go.opentelemetry.io/otel/trace v1.39.0 h1:2d2vfpEDmCJ5zVYz7ijaJdOF59xLomrvj7bjt6
 go.opentelemetry.io/otel/trace v1.39.0/go.mod h1:88w4/PnZSazkGzz/w84VHpQafiU4EtqqlVdxWy+rNOA=
 go.temporal.io/api v1.62.11 h1:MWDaooDvOJCIRb1atqeZX2ErDPNTsNc3/mMEVEvvaVU=
 go.temporal.io/api v1.62.11/go.mod h1:iaxoP/9OXMJcQkETTECfwYq4cw/bj4nwov8b3ZLVnXM=
+go.temporal.io/api v1.62.12 h1:627rVnItegQmrszg1bH4vfyc/1uNo5qCereCNkvZefw=
```
S15, line 125: smol mod tidy

S15 (approved):

PR #340, opened 2026-04-28
Project harness for Go

THardy98:
## What was changed
Adds `go/harness/` package, effectively a port of the same Python harness (`python/harness`).

`go/harness` supports the same harness semantics, structure, and API as the existing Python harness. The test suite (`project_test.go` and `worker_test.go`) similarly mimic the existing Python harness test suite.

The Go worker uniquely supported additional flags for worker deployment versioning that were not available/used in other languages, that has correspondingly been added to the Go's worker harness.

**Note**: in contrast to the harness in other languages, the Go harness is **not** a standalone Go module. There is significant coupling between the existing go worker code and Omes which would make this either a much larger change, or cause a lot of duplication. The plan is to use the existing `workers/go/go.mod` as a single source to manage dependencies between the harness and the tests that consume it (at least, until this is resolved)

## Why?
- load testing ergonomics / DX
- Language parity


Sushisource (commented):
Am I missing where there's a Go test project?

THardy98:
> Am I missing where there's a Go test project?

Adding in a subsequent PR. Wanted to separate the harness implementation and the actual `helloworld` project. Thought the diff might get quite big otherwise.

(this applies for the other harness PRs as well for TS and Ruby)

Sushisource (approved):

PR #342, opened 2026-05-01
Project harness for Java

THardy98:
## What was changed
The same as all the other project harnesses.

Added a standalone project harness for Java, makes it more ergonomic to write omes load tests.

2. How was this tested:
Small test suite


workers/java/harness/src/main/java/io/temporal/omes/harness/HarnessHelpers.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+
+final class HarnessHelpers {
+  private HarnessHelpers() {}
+
+  static Logger configure(String logLevel, String logEncoding) {
+    Logger logger = LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
+    if (!(logger instanceof ch.qos.logback.classic.Logger)) {
+      return logger;
+    }
```
Sushisource, line 23: Would this ever not be true since we just control everything in here? I'd be fine with just erroring if someone sets a logger that isn't the kind of logger we support.

Sushisource (approved):

PR #345, opened 2026-05-07
Added support for go lambda worker

gcristea-temporal:
## What was changed
* New Lambda worker ([workers/go/lambda/worker.go](workers/go/lambda/worker.go)) — reads all configuration from environment variables rather than CLI flags. Connects to Temporal with optional TLS/API-key credentials fetched from AWS Secrets Manager, sets `VersioningBehaviorPinned` by default, and registers the full set of kitchen sink workflows and activities via `lambdaworker.RunWorker`.
* Build system integration ([workers/build.go](workers/build.go), [cmd/cli/prepare_worker.go](cmd/cli/prepare_worker.go)) — adds a `--lambda` flag to `prepare-worker`. When set, the generated main.go calls `lambda.Main()`  instead of `worker.Main()`, producing a Lambda-compatible binary.
* Environment variable binding ([cmd/clioptions/envvars.go](cmd/clioptions/envvars.go)) — adds `BindEnvVars` / `FlagNameToEnvVar` helpers that map `pflag` flags to `TEMPORAL_OMES_*` environment variables, allowing the Lambda worker to be configured without CLI args.
* Scenario update ([scenarios/serverless_burst.go](scenarios/serverless_burst.go)) — wires `AutoUpgradeVersioningOverride` into the kitchen sink workflow options so the serverless burst scenario works with deployment-versioned Lambda workers.
* Dependencies ([workers/go/go.mod](workers/go/go.mod)) — pulls in `go.temporal.io/sdk/contrib/aws/lambdaworker`, `go.temporal.io/sdk/contrib/envconfig`, and the AWS SDK v2 packages needed for Secrets Manager.

## Why?

The goal was to use Omes for scale testing of serverless workers.
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


THardy98 (approved):
LGTM - ideally i'd like to avoid creating a new build/entrypoint just for lambda workers but not going to block on it.

Going forward, use cases like this might be good candidates for the added project harnesses

PR #348, opened 2026-05-11
Ability to run dev server from git ref

stephanos:
## What was changed

Added new package `devserver` which allows to start a local Temporal server from any git ref (SHA, tag, branch).

## Why?

Right now only the CLI can start a Temporal devserver, but it has a very long release cycle.

(1) Allow testing new features as soon as they land in Temporal Server without having to wait for a CLI release.
(2) Allow testing new features in Temporal Server's [mixed brain suite](https://github.com/temporalio/temporal/tree/main/tests/mixedbrain).

Long-term this will move to the CLI, but that's a larger and later effort. In the meantime it can incubate here and unblock the aforementioned use cases.


cmd/dev/util.go
```diff
@@ -122,48 +118,6 @@ func getTargetDir(target string) (string, error) {
 	return filepath.Join(repoDir, "workers", target), nil
 }
 
-// loadVersions parses versions.env file and returns a map of version variables
```
stephanos, line 125: moved this to `versions` package for re-use from `devserver` package

devserver/config.go
```diff
@@ -0,0 +1,168 @@
...
+			listeners = append(listeners, l)
+			ports[i] = base + i
+		}
+		for _, l := range listeners {
+			_ = l.Close()
+		}
+		return ports, nil
+	}
+	return [portCount]int{},
+		fmt.Errorf("allocatePorts: no free contiguous range in [%d, %d) after %d attempts: %w", safePortMin, safePortMax, maxPortAllocationAttempts, lastErr)
```
THardy98: why do we need a contiguous range ?

devserver/config.go
```diff
@@ -0,0 +1,168 @@
...
+				}
+				continue attempt
+			}
+			listeners = append(listeners, l)
+			ports[i] = base + i
+		}
+		for _, l := range listeners {
+			_ = l.Close()
+		}
+		return ports, nil
```
THardy98: nit (non-blocking): might be easier to read if this were a helper function where we could `defer` closing the listeners (don't need `attempt` or `continue`, check multiple spots to close listeners etc.). Maybe also prefer `net.JoinHostPort` over `net.Listen`.

I prompted Codex for this, and it suggested:
```
func tryAllocatePorts(host string, base int) ([portCount]int, error) {
	var ports [portCount]int

	listeners := make([]net.Listener, 0, portCount)
	defer func() {
		for _, l := range listeners {
			_ = l.Close()
		}
	}()

	for i := 0; i < portCount; i++ {
		port := base + i
		addr := net.JoinHostPort(host, strconv.Itoa(port))

		l, err := net.Listen("tcp", addr)
		if err != nil {
			return ports, err
			}

		listeners = append(listeners, l)
		ports[i] = port
	}
```
which imo is a bit nicer

devserver/config.go
```diff
@@ -0,0 +1,168 @@
...
+	driver := cmp.Or(p.Driver, "sqlite")
+	add("DB", driver)
+	switch driver {
+	case "sqlite":
+		// Use in-memory SQLite.
+		add("SQLITE_MODE", "memory")
+		add("SQLITE_CACHE", "shared")
+	default:
+		return nil, fmt.Errorf("unsupported persistence driver %q (want sqlite)", driver)
+	}
```
THardy98: we only support `sqlite` to start?

The `Driver` param in `Options` only documents that `sqlite` is the default

devserver/devserver.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Start clones+builds the requested ref and runs the server, returning once
+// the frontend is reachable. The caller must invoke Stop to release resources.
+func Start(ctx context.Context, opts Options) (*Server, error) {
+	// Validate + apply defaults.
+	if opts.Ref == "" && opts.SourceDir == "" {
+		return nil, errors.New("devserver: Ref or SourceDir is required")
+	}
+	opts.Logger = cmp.Or(opts.Logger, zap.NewNop().Sugar())
```
THardy98, line 91: Should these be mutually exclusive? `Ref` implies clone + build, `SourceDir` implies just build (doesn't need a gitref)

THardy98 (approved):
overall lgtm, left a few comments/nits

devserver/config.go
stephanos: 👍 ah I had listed all explicitly and removed that, but then forgot to remove the default

devserver/config.go
stephanos: You Codex is clearly smarter than my Codex ;)

devserver/devserver.go
stephanos: 👍 agreed

devserver/config.go
stephanos: 👍 we don't really

THardy98: kk, then i don't think we need to restart the whole port allocation if a single port binding fails

we can just try on a different port and keep the bindings we have already

not a big deal, but thought i'd mention

stephanos: 👍 I've made it so it only retries individual port allocation now

PR #350, opened 2026-05-12
Add Nexus attach-callbacks action to throughput stress

prathyushpv:
## What was changed
Add Nexus attach-callbacks action to throughputstress similar to bench-go.

## Why?
To exercise this feature in load test and make omes throughtput stress on par with bench-go.


scenarios/throughput_stress.go
```diff
@@ -654,6 +657,40 @@ func (t *tpsExecutor) createNexusWaitForCancelAction() *Action {
...
+// `numOps` concurrent Nexus operations against the same handler workflow ID. The first
+// operation starts the handler workflow; the rest attach as completion callbacks. The handler
+// sleeps `handlerBlockDuration` (via before_actions) so all callbacks have time to attach
+// before it completes and fans out completion to each.
+func (t *tpsExecutor) createNexusAttachCallbacksAction() *Action {
+	const numOps = 3
+	const handlerBlockDuration = 5 * time.Second
+	handlerWfID := "nexus-attach-handler-" + uuid.NewString()
+	actions := make([]*Action, 0, numOps)
+	for i := 0; i < numOps; i++ {
```
stephanos: 🤖 `nit` **[complexity]**: Use `for range numOps` (Go 1.22+).

**Issue:** The loop variable `i` is unused. Go 1.25 (per go.mod) supports range-over-integer.

**Suggestion:** `for range numOps {`

workers/go/kitchensink/kitchen_sink.go
```diff
@@ -503,6 +518,29 @@ func handleNexusOperation(ctx workflow.Context, nexusOp *kitchensink.ExecuteNexu
 		})
 }
 
+func handleSendSignal(ctx workflow.Context, ws *KSWorkflowState, action *kitchensink.SendSignalAction) error {
+	return withAwaitableChoiceCustom(ctx, ws, func(ctx workflow.Context) workflow.Future {
+		return workflow.SignalExternalWorkflow(ctx, action.WorkflowId, action.RunId, action.SignalName, nil)
```
stephanos: 🤖 `nit` **[bug]**: `SendSignalAction.args` and `.headers` are silently dropped.

**Issue:** The proto defines `repeated Payload args = 4` and `map headers = 5` on `SendSignalAction`, but this implementation passes `nil` as the signal payload, discarding any args the caller specified. The Python `handle_send_signal` does the same (`ext_handle.signal(action.signal_name)` with no args). For the current "unblock" scenario this is fine, but any future use of `SendSignalAction` with signal data would silently lose it.

**Suggestion:** If keeping it simple for now, add a brief comment noting the limitation (e.g. `// TODO: forward action.Args once raw-payload passthrough is supported`). Alternatively, forward the first arg as a `RawValue`:
```go
var arg interface{}
if len(action.Args) > 0 {
    arg = action.Args[0]
}
return workflow.SignalExternalWorkflow(ctx, action.WorkflowId, action.RunId, action.SignalName, arg)
```

stephanos (approved):
Approval contingent on discussed, planned refactoring

THardy98 (changes requested):
Looks good, but we should add support for this in Java, TS, and I believe .NET as well.

THardy98 (approved):

THardy98 (commented):
would appreciate an issue to track doing this for the remaining languages (and can tag this PR as reference)

PR #351, opened 2026-05-12
Add standalone activity to throughput stress (opt-in)

prathyushpv:
## What was changed

- Adds a `DoStandaloneActivity` ClientAction to the kitchen-sink DSL that invokes the `payload` activity directly via the low-level `StartActivityExecution` / `PollActivityExecution` RPCs, bypassing normal workflow activity scheduling.
- Wires it into `throughput_stress` behind a new `--option enable-standalone-activity=true` flag (default off).
- Threads `Namespace` through `ClientActionsExecutor` (required for the namespace-scoped RPCs).
- Go SDK implements the variant; other SDKs (Java, Python, TS, .NET, Ruby) raise a non-retryable `ApplicationFailure` if the variant is encountered.

## Why?

Mirrors the standalone activity pattern in bench-go's `throughputstress` scenario so we can benchmark workflow-independent activities (`StartActivityExecution` / `PollActivityExecution`) in omes.


workers/dotnet/WorkerLib/KitchenSink/ClientActionsExecutor.cs
```diff
@@ -70,6 +70,13 @@ private async Task ExecuteClientAction(ClientAction action)
             var handle = _client.GetWorkflowHandle(WorkflowId!);
             await handle.DescribeAsync();
         }
+        else if (action.DoStandaloneActivity != null)
+        {
+            throw new ApplicationFailureException(
+                "do_standalone_activity is not implemented for .NET",
+                "UnsupportedOperation",
+                nonRetryable: true);
+        }
```
THardy98, line 76: There's a flag that exists called `err-if-unimplemented` or something like that. This is probably a place where we want to check if that's set. If it's false, then we should probably just log instead of throw.

(applies to the other langs where we haven't implemented this as well)

THardy98 (changes requested):
Generally - would prefer if we kept parity between languages for the operations they support in kitchensink, especially since the code snippet to implement is quite small:

```
func (e *ClientActionsExecutor) executeStandaloneActivity(ctx context.Context, sa *DoStandaloneActivity) error {
	act := sa.GetActivity()
	if act == nil {
		return fmt.Errorf("DoStandaloneActivity.activity is required")
	}

	actType, args := ActivityNameAndArgs(act)

	handle, err := e.Client.ExecuteActivity(ctx, client.StartActivityOptions{
		ID:                     fmt.Sprintf("standalone-%s-%d", e.WorkflowOptions.ID, time.Now().UnixNano()),
		TaskQueue:              act.TaskQueue,
		ScheduleToCloseTimeout: act.ScheduleToCloseTimeout.AsDuration(),
		ScheduleToStartTimeout: act.ScheduleToStartTimeout.AsDuration(),
		StartToCloseTimeout:    act.StartToCloseTimeout.AsDuration(),
		HeartbeatTimeout:       act.HeartbeatTimeout.AsDuration(),
		RetryPolicy:            ConvertFromPBRetryPolicy(act.RetryPolicy),
	}, actType, args...)
	if err != nil {
		return fmt.Errorf("failed to start standalone activity: %w", err)
	}
	return handle.Get(ctx, nil)
}
```
across the languages.

If it's pressing to get this in asap, filing an issue would be great so we can track where parity is missing

PR #361, opened 2026-05-22
Consolidate kitchen sink Nexus operations

stephanos:
Replaces the special-purpose Nexus handler workflow and separate operations with one kitchen-sink `execute` operation.

This is a refactoring first and foremost.


workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -511,41 +495,49 @@ enum ActivityCancellationType {
...
-
-// Input for the Nexus handler workflow that backs echo-sync and echo-async operations
-message NexusHandlerInput {
-  string input = 1;
-  repeated ActionSet before_actions = 2;
-  string handler_workflow_id = 3;
-  temporal.api.enums.v1.WorkflowIdConflictPolicy handler_workflow_id_conflict_policy = 4;
-  // If true, the handler workflow waits on the "unblock" signal before returning.
-  bool wait_for_signal = 5;
+  oneof variant {
```
stephanos, line 507: This is the main change.

workers/ruby/protos/kitchen_sink/kitchen_sink_pb.rb
```diff
@@ -1,17 +1,17 @@
...
+# source: kitchen_sink.proto
 
 require 'google/protobuf'
 
 require 'google/protobuf/duration_pb'
 require 'google/protobuf/empty_pb'
-require 'temporalio/api/common/v1/message'
-require 'temporalio/api/failure/v1/message'
-require 'temporalio/api/enums/v1/workflow'
+require 'temporal/api/common/v1/message_pb'
```
stephanos: 🤖

<details>
<summary><strong>high</strong> — Regenerated Ruby bindings require <code>temporal/api/...\_pb</code>, which the <code>temporalio</code> gem does not ship, so the Ruby worker will fail to load.</summary>

The three `require` lines changed from the gem's paths to raw protoc-ruby defaults:

```diff
-require 'temporalio/api/common/v1/message'
-require 'temporalio/api/failure/v1/message'
-require 'temporalio/api/enums/v1/workflow'
+require 'temporal/api/common/v1/message_pb'
+require 'temporal/api/failure/v1/message_pb'
+require 'temporal/api/enums/v1/workflow_pb'
```

The `temporalio` gem (pinned to `1.6.0` in `mise.toml` / `Gemfile`) exposes its protos at `lib/temporalio/api/common/v1/message.rb`; there is no `temporal/api/common/v1/message_pb.rb` on the load path, and `workers/ruby/protos/` contains only `kitchen_sink_pb.rb`. `workers/ruby/workerlib/kitchensink/kitchen_sink.rb:3` requires this file, so the Ruby worker raises `LoadError` at startup — and the Ruby target is built and tested in CI (`.github/workflows/ci.yml`).

Even ignoring the `LoadError`: `add_serialized_file` below still declares dependencies on `temporal/api/common/v1/message.proto`, which are only registered in the descriptor pool by loading the gem's `temporalio/api/...` files. The new requires cannot register them.

The `# source:` header also flipped from `kitchen_sink/kitchen_sink.proto` to `kitchen_sink.proto`, which suggests this file was regenerated from a different working directory / `-I` root than the one used previously — likely the root cause of both diffs.

</details>

**Suggestion:** Regenerate from `workers/proto` (so the descriptor path stays `kitchen_sink/kitchen_sink.proto`) with the same protoc invocation used previously, or restore the gem-relative requires:

```suggestion
require 'temporalio/api/common/v1/message'
```

loadgen/kitchensink/client_action_executor.go
```diff
@@ -215,7 +215,15 @@ func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Cont
 		return fmt.Errorf("NewNexusClient: %w", err)
 	}
 
-	handle, err := nexusClient.ExecuteOperation(ctx, sno.Operation, &NexusHandlerInput{}, client.StartNexusOperationOptions{
+	var input any = ""
+	if sno.Operation == "echo-async" {
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — Standalone Nexus input shape is inferred from a hardcoded operation name, so any other workflow-run operation silently gets the wrong input.</summary>

`DoStandaloneNexusOperation` carries only an `operation` string, so this branch guesses the payload type from the literal `"echo-async"`. Every other operation name gets `input = ""`.

Failure scenario: a scenario sets `DoStandaloneNexusOperation{Operation: "echo-async-v2"}` (or `echo-async` is renamed, or a second workflow-run operation is added). `""` is sent to a handler whose input is `*StartWorkflow`, the handler deserializes an empty/garbage `StartWorkflow`, and the resulting kitchen-sink workflow starts with an empty `WorkflowInput` — i.e. no `ReturnResult` action — so it blocks forever with no execution timeout (see `EchoAsyncOperation`, which only sets `WorkflowExecutionTimeout` when `workflow_id` is set). The standalone op then just times out on its 90s `ScheduleToCloseTimeout` with an opaque error.

This is exactly the ambiguity the new `ExecuteNexusOperation.variant` oneof removes for the in-workflow path; the standalone client-action path still relies on the magic string.

</details>

**Suggestion:** Give `DoStandaloneNexusOperation` the same `oneof variant { OperationCall call; StartWorkflow start_workflow; }` so the caller states the input shape explicitly. If that's out of scope for this PR, at minimum make the fallback explicit rather than silently defaulting to `""` — e.g. `default: return fmt.Errorf("unsupported standalone nexus operation %q", sno.Operation)` on a `switch sno.Operation`.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,37 +531,42 @@ enum ActivityCancellationType {
...
+    // Invoke a Nexus operation with a string input.
+    OperationCall call = 3;
```

<details>
<summary><strong>small</strong> — Deleted field numbers aren't <code>reserved</code>, and the oneof reuses tags 3 and 7 with incompatible types.</summary>

`ExecuteNexusOperation` drops `operation = 2`, `expected_output = 6`, `handler_workflow_id = 8`, `handler_workflow_id_conflict_policy = 9`, and `wait_for_signal = 10` without a `reserved` declaration, so nothing stops a future change from reusing those numbers for a different meaning.

Worse, the new oneof reuses tags that previously held different types:

- `3` was `string input`, now `OperationCall call` — both are wire type 2, so old bytes parse into an `OperationCall` as garbage rather than erroring.
- `7` was `repeated ActionSet before_actions`, now `StartWorkflow start_workflow` — also silently mis-parses.

Everything here is built from one commit, so this isn't breaking today, but it removes the safety net that would normally surface a stale payload (persisted scenario input, a workflow history mid-upgrade, a stale worker image) as a decode error instead of nonsense field values.

</details>

**Suggestion:** Reserve the removed tags and pick fresh numbers for the oneof:

```proto
message ExecuteNexusOperation {
  reserved 2, 3, 6, 7, 8, 9, 10;
  reserved "operation", "input", "expected_output", "before_actions",
           "handler_workflow_id", "handler_workflow_id_conflict_policy", "wait_for_signal";

  string endpoint = 1;
  map<string, string> headers = 4;
  AwaitableChoice awaitable_choice = 5;
  oneof variant {
    OperationCall call = 11;
    StartWorkflow start_workflow = 12;
  }
}
```

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -712,71 +707,23 @@ impl<'a> Arbitrary<'a> for ExecuteChildWorkflowAction {
     }
 }
 
-static NEXUS_OPERATIONS: [&str; 2] = ["echo-sync", "echo-async"];
-
-impl ExecuteNexusOperation {
-    fn echo_sync(u: &mut Unstructured<'_>) -> arbitrary::Result<Self> {
+impl<'a> Arbitrary<'a> for ExecuteNexusOperation {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        // The fuzzer only generates echo-sync; workflows started through Nexus need
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — Fuzzer now emits only <code>echo-sync</code>/<code>WaitFinish</code>, dropping all generated coverage of async Nexus operations and Nexus cancellation.</summary>

Previously `Arbitrary for ExecuteNexusOperation` chose between `echo-sync` and `echo-async`, and the async branch generated arbitrary `AwaitableChoice` values — including `CancelBeforeStarted` / `CancelAfterStarted` / `CancelAfterCompleted` — plus random `before_actions`. After this change the generator produces one shape, so the fuzzer no longer exercises async Nexus start/cancel paths at all. Those paths only survive in the hand-written `throughput_stress` actions and `kitchen_sink_executor_test.go`.

The stated reason ("workflows started through Nexus need coordinated completion that does not compose under random action generation") holds for arbitrary nested action sets, but not for a self-completing handler: `StartWorkflow{ workflow_input: { initial_actions: [ReturnResult] } }` terminates on its own exactly like the old `echo-async`, which is precisely what `createNexusEchoAsyncAction` and the standalone executor now do.

</details>

**Suggestion:** Keep an async arm with a fixed, self-completing workflow input so cancellation choices stay in the generated space:

```rust
if u.ratio(1, 2)? {
    // echo-async with a self-completing handler: safe under any awaitable choice.
    variant: Some(execute_nexus_operation::Variant::StartWorkflow(StartWorkflow {
        operation: "echo-async".to_string(),
        workflow_input: Some(WorkflowInput {
            initial_actions: vec![mk_action_set([/* ReturnResult */])],
            ..Default::default()
        }),
        ..Default::default()
    })),
    awaitable_choice: Some(u.arbitrary()?),
    ...
}
```

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -610,49 +622,34 @@ type ReturnOrErr struct {
...
-	}, nil
-})
+// EchoAsyncOperation starts a KitchenSinkWorkflow through Nexus.
+var EchoAsyncOperation = temporalnexus.MustNewWorkflowRunOperationWithOptions(
+	temporalnexus.WorkflowRunOperationOptions[*kitchensink.StartWorkflow, *common.Payload]{
+		Name: "echo-async",
+		Handler: func(ctx context.Context, input *kitchensink.StartWorkflow, opts nexus.StartOperationOptions) (temporalnexus.WorkflowHandle[*common.Payload], error) {
+			input = cmp.Or(input, &kitchensink.StartWorkflow{})
+			workflowOptions := client.StartWorkflowOptions{ID: cmp.Or(input.GetWorkflowId(), opts.RequestID)}
+			if input.GetWorkflowId() != "" {
```

<details>
<summary><strong>small</strong> — The 60-minute execution timeout only applies when an explicit workflow ID is given, which is now the riskier of the two branches.</summary>

`WorkflowExecutionTimeout` is set only inside `if input.GetWorkflowId() != ""`. The old `NexusHandlerWorkflow` made that asymmetry harmless: without a workflow ID it ran `before_actions` and returned. Now the request-ID branch starts a full `KitchenSinkWorkflow` driven by caller-supplied actions, so an input that never produces a `ReturnResult` (e.g. `createNexusWaitForCancelAction`'s `AwaitWorkflowState("never", "resolves")`, or an empty `WorkflowInput`) leaves an unbounded workflow behind if the Nexus operation is abandoned rather than cancelled — the exact leak the comment on the old code ("Cap the handler so we don't leave dangling workflows if a stress run fails") was guarding against.

`workers/python/nexus_service.py` has the same split (`execution_timeout=timedelta(minutes=60)` only in the `input.workflow_id` branch).

</details>

**Suggestion:** Apply the cap unconditionally and keep only the conflict policy in the ID-specific branch:

```suggestion
			workflowOptions := client.StartWorkflowOptions{
				ID:                       cmp.Or(input.GetWorkflowId(), opts.RequestID),
				WorkflowExecutionTimeout: 60 * time.Minute,
			}
			if input.GetWorkflowId() != "" {
				workflowOptions.WorkflowIDConflictPolicy = input.GetWorkflowIdConflictPolicy()
			}
```

loadgen/kitchensink/client_action_executor.go
```diff
@@ -129,8 +130,8 @@ func (e *ClientActionsExecutor) executeClientAction(ctx context.Context, action
 	} else if action.GetNestedActions() != nil {
 		err = e.executeClientActionSet(ctx, action.GetNestedActions())
 		return err
-	} else if sano := action.GetDoStandaloneNexusOperation(); sano != nil {
-		return e.executeStandaloneNexusOperation(ctx, sano)
+	} else if standalone := action.GetDoStandaloneNexusOperation(); standalone != nil {
```
stephanos: fix: keep `sano` name

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,46 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
 	return run, err
 }
 
-func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, sno *DoStandaloneNexusOperation) error {
+func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, standalone *DoStandaloneNexusOperation) error {
```
stephanos: fix: make `standalone` to `sano`

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,46 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
...
-	handle, err := nexusClient.ExecuteOperation(ctx, sno.Operation, &NexusHandlerInput{}, client.StartNexusOperationOptions{
+	handle, err := nexusClient.ExecuteOperation(ctx, nexusOp.GetOperation(), nexusOp.GetInput(), client.StartNexusOperationOptions{
 		ID:                     operationID,
 		ScheduleToCloseTimeout: 90 * time.Second,
 	})
 	if err != nil {
 		return fmt.Errorf("ExecuteOperation: %w", err)
 	}
 
+	if expectedOutput := nexusOp.GetExpectedOutput(); expectedOutput != nil {
```
stephanos, line 234: To match the workflow-based one.

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,46 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
...
+		var result commonpb.Payload
+		if err := handle.Get(ctx, &result); err != nil {
+			return fmt.Errorf("Get Nexus operation: %w", err)
```
stephanos: Keep "standalone" qualifier in error messages in this method

loadgen/kitchensink/client_action_executor_test.go
```diff
@@ -0,0 +1,17 @@
+package kitchensink
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNexusStartWorkflowRequest(t *testing.T) {
```
stephanos: fix; drop this entire file again

loadgen/kitchen_sink_executor_test.go
```diff
@@ -985,11 +991,40 @@ func TestKitchenSink(t *testing.T) {
...
 			},
 			historyMatcher: PartialHistoryMatcher(`
-				NexusOperationScheduled {"operation":"echo-async"}
+				NexusOperationScheduled {"operation":"execute"}
 				NexusOperationStarted
 				NexusOperationCompleted`),
 			expectedUnsupportedErrs: nexusUnsupportedSDKs,
 		},
+		{
+			name: "NexusOperation/Sync/ExpectedOutputMismatch",
```
stephanos: This needs to be in the same `NexusOperation/Sync` group ie next to them

loadgen/kitchen_sink_executor_test.go
```diff
@@ -985,11 +991,40 @@ func TestKitchenSink(t *testing.T) {
...
+								},
+							},
+						}),
+				},
+			},
+			historyMatcher: PartialHistoryMatcher(`
+				NexusOperationScheduled {"operation":"execute"}
+				NexusOperationCompleted`),
+			expectedUnsupportedErrs: nexusUnsupportedSDKs,
+			expectedWorkflowError:   `expected output`,
```
stephanos: fix; can this be more specific?

clioptions/client.go
```diff
@@ -176,6 +176,10 @@ func (p *PassThroughPayloadConverter) FromPayload(payload *common.Payload, value
 	if err != nil {
 		return fmt.Errorf("unable to decode raw payload: %w", err)
 	}
+	if output, ok := valuePtr.(*common.Payload); ok {
+		*output = *innerPayload
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — `*output = *innerPayload` copies a protobuf message by value, which the protobuf-go contract forbids and `go vet` flags.</summary>

`common.Payload` embeds `protoimpl.MessageState`, which contains `pragma.DoNotCopy` (`[0]sync.Mutex`) plus the cached `atomicMessageInfo` pointer. Assigning the struct by value:

- is reported by `go vet`'s `copylocks` analyzer ("assignment copies lock value"), so anyone running `go vet ./...` or `golangci-lint` (govet) on this repo now gets a failure — `cmd/dev lint-and-format` only runs `go fmt`/`go build`, so CI won't catch it;
- copies internal reflection state that protobuf-go explicitly documents as non-copyable, and silently discards nothing today only because source and destination happen to be the same concrete type.

This is on the hot path for every Nexus result decode added by this PR (`handleNexusOperation` and `executeStandaloneNexusOperation` both do `fut.Get(ctx, &commonpb.Payload{...})`), so it is worth getting right rather than relying on the copy happening to work.

</details>

**Suggestion:**

```suggestion
	if output, ok := valuePtr.(*common.Payload); ok {
		proto.Reset(output)
		proto.Merge(output, innerPayload)
		return nil
	}
```

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -394,6 +392,8 @@ message SendSignalAction {
   // Headers to attach to the signal
   map<string, temporal.api.common.v1.Payload> headers = 5;
   AwaitableChoice awaitable_choice = 6;
+  // When set, deliver this typed payload to the do_actions_signal handler.
+  DoSignal.DoSignalActions do_actions = 7;
```

<details>
<summary><strong>med</strong> — The new <code>SendSignalAction.do_actions</code> field is silently ignored by the Java worker, which will send a signal with an empty name instead of failing loudly.</summary>

`do_actions` overrides the signal name (the Go worker rewrites it to `do_actions_signal`, the Python worker calls `ext_handle.signal("do_actions_signal", ...)`), and `scenarios/throughput_stress.go`'s `createNexusAttachCallbacksAction` now relies on this by setting **only** `do_actions` and leaving `signal_name` empty.

`workers/java/.../KitchenSinkWorkflowImpl.java:203-220` still does:

```java
Async.procedure(stub::signal, sendSignal.getSignalName(), sendSignal.getArgsList());
```

so a `SendSignalAction` carrying `do_actions` becomes a signal with name `""` and no actions — a confusing server-side error or a silently no-op signal, rather than the clear `"X is not supported"` `ApplicationFailure` this repo uses everywhere else for unimplemented features (e.g. `ExecuteNexusOperation is not supported` a few lines below).

Today this is unreachable because the only producer is behind Nexus, which Java already rejects. But the field is now part of the shared kitchen-sink contract, so the next scenario that uses `do_actions` outside a Nexus context will hit it.

</details>

**Suggestion:** Add an explicit guard in the Java worker's `hasSendSignal()` branch (and the same for any other SDK that implements `send_signal`) that throws `ApplicationFailure.newNonRetryableFailure("SendSignalAction.do_actions is not supported", "UnsupportedOperation")` when `sendSignal.hasDoActions()`.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,37 +529,52 @@ enum ActivityCancellationType {
...
-message NexusHandlerInput {
-  string input = 1;
-  repeated ActionSet before_actions = 2;
-  string handler_workflow_id = 3;
-  temporal.api.enums.v1.WorkflowIdConflictPolicy handler_workflow_id_conflict_policy = 4;
-  // If true, the handler workflow waits on the "unblock" signal before returning.
-  bool wait_for_signal = 5;
+  NexusOperationRequest input = 3;
+  // Service name. Defaults to the kitchen sink service when empty.
+  string service = 7;
```

<details>
<summary><strong>small</strong> — Field numbers are re-typed and dropped without <code>reserved</code> declarations.</summary>

Across the two messages changed here, tags are reused with incompatible types and removed tags are left free:

- `DoStandaloneNexusOperation`: `1` went from `string endpoint` to `ExecuteNexusOperation operation`; `2` (`service`) and `3` (`operation`) were deleted.
- `ExecuteNexusOperation`: `3` went from `string input` to `NexusOperationRequest input`; `6` from `string expected_output` to `Payload expected_output`; `7` from `repeated ActionSet before_actions` to `string service`; `8`/`9`/`10` (`handler_workflow_id`, `handler_workflow_id_conflict_policy`, `wait_for_signal`) were deleted.

Because loadgen and workers are built from the same tree, the retype is safe for live runs. The lasting risk is the deleted tags: nothing stops a future change from reusing `8`–`10` (or `2`/`3`), at which point any serialized `TestInput` written by `kitchen-sink-gen --proto-output` from an older build silently decodes into the wrong fields instead of failing.

</details>

**Suggestion:** Add `reserved` entries for the removed tags (and their names), e.g. `reserved 8, 9, 10; reserved "handler_workflow_id", "handler_workflow_id_conflict_policy", "wait_for_signal";` in `ExecuteNexusOperation` and `reserved 2, 3; reserved "service", "operation";` in `DoStandaloneNexusOperation`.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -610,69 +609,79 @@ type ReturnOrErr struct {
...
+) (temporalnexus.TemporalOperationResult[*common.Payload], error) {
+	input = cmp.Or(input, &kitchensink.ExecuteActivityAction{})
 	activityOpts := client.StartActivityOptions{
 		// Reuse the Nexus request ID so retries attach to the original activity.
 		ID:                       "nexus-standalone-activity-" + opts.RequestID,
-		StartToCloseTimeout:      30 * time.Second,
+		TaskQueue:                input.GetTaskQueue(),
+		ScheduleToCloseTimeout:   input.GetScheduleToCloseTimeout().AsDuration(),
+		ScheduleToStartTimeout:   input.GetScheduleToStartTimeout().AsDuration(),
+		StartToCloseTimeout:      input.GetStartToCloseTimeout().AsDuration(),
```

<details>
<summary><strong>small</strong> — Dropping the hardcoded 30s <code>StartToCloseTimeout</code> makes a request with no timeouts fail as "standalone activities are disabled", which is the wrong diagnosis.</summary>

Previously this always set `StartToCloseTimeout: 30 * time.Second`. Now every timeout comes from the request, so a `start_activity` action that omits them (nothing validates this — the proto has no required fields, and neither `kitchen-sink-gen` nor `kitchen_sink_executor_test.go` exercises this path) sends all-zero timeouts. The server rejects that with `InvalidArgument`, which the error mapping right below converts into a non-retryable:

> cannot start standalone activity; standalone activities (activity.enableStandalone) and activity completion callbacks (activity.enableCallbacks) must be enabled

So a missing-timeout bug in the input gets reported as a server misconfiguration. The only in-repo caller (`nexusStandaloneActivityRequest` in `scenarios/throughput_stress.go`) does set 30s, so this is latent rather than currently broken.

</details>

**Suggestion:** Keep a fallback so the request stays valid regardless of input:

```suggestion
		StartToCloseTimeout:      cmp.Or(input.GetStartToCloseTimeout().AsDuration(), 30*time.Second),
```

(or validate up front and return an explicit `nexus.HandlerErrorTypeBadRequest` naming the missing timeout).

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,46 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
...
 
-func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, sno *DoStandaloneNexusOperation) error {
+func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, standalone *DoStandaloneNexusOperation) error {
+	nexusOp := standalone.GetOperation()
+	if nexusOp == nil {
+		return fmt.Errorf("DoStandaloneNexusOperation.operation is required")
+	}
+	service := nexusOp.GetService()
+	if service == "" {
+		service = "kitchen-sink"
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — The service name stays a bare literal even though this PR introduced a const for the operation name.</summary>

`KitchenSinkNexusOperationName` was added to `loadgen/kitchensink/helpers.go` for exactly this reason, but `"kitchen-sink"` is still spelled out in six places: here, `scenarios/throughput_stress.go:936`, `loadgen/kitchen_sink_executor_test.go:1100` and `:1133`, `loadgen/kitchen-sink-gen/src/main.rs:718`/`:733`, plus the two worker-side constants (`KitchenSinkServiceName`, `KITCHEN_SINK_SERVICE_NAME`).

</details>

**Suggestion:** Export a `KitchenSinkNexusServiceName = "kitchen-sink"` next to `KitchenSinkNexusOperationName` in `loadgen/kitchensink/helpers.go` and use it here and in the other loadgen/scenario call sites:

```suggestion
		service = KitchenSinkNexusServiceName
```

loadgen/kitchensink/helpers.go
```diff
@@ -88,6 +91,16 @@ func ClientActions(clientActions ...*ClientAction) *ClientSequence {
 	}
 }
 
+func NexusStartWorkflowRequest(workflowID string, options *NexusWorkflowStartOptions) *NexusOperationRequest {
```
stephanos: fix; let's inline this

scenarios/throughput_stress.go
```diff
@@ -582,20 +582,26 @@ func (t *tpsExecutor) createActionsChunk(
 		// Add Nexus operations, if configured.
 		if t.config.NexusEnabled {
 			asyncActions = append(asyncActions, t.createNexusEchoSyncAction())
-			asyncActions = append(asyncActions, t.createNexusEchoAsyncAction())
+			asyncActions = append(asyncActions, t.createNexusStartWorkflowAction())
 			asyncActions = append(asyncActions, t.createNexusWaitForCancelAction())
 			asyncActions = append(asyncActions, t.createNexusAttachCallbacksAction())
 			if t.config.IncludeStandaloneNexus {
 				asyncActions = append(asyncActions,
```
stephanos, line 589: fix: add a small comment above each of the two to say it's sync or async

workers/go/workerlib/kitchensink/kitchen_sink_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	ks "github.com/temporalio/omes/loadgen/kitchensink"
+)
+
+func TestNexusWorkflowOptionsAlwaysSetExecutionTimeout(t *testing.T) {
```
stephanos: fix; drop this file again

workers/python/nexus_service.py
```diff
@@ -1,63 +1,72 @@
...
+    ) -> nexus.TemporalOperationResult[Payload]:
+        action = input.WhichOneof("action")
+        if action == "echo":
+            return nexus.TemporalOperationResult.sync(
+                Payload(
+                    metadata={"encoding": b"json/plain"},
+                    data=json.dumps(input.echo).encode(),
+                )
+            )
+        if action != "workflow_action" or not input.workflow_action.HasField("start"):
```
stephanos: fix: instead; why not make the error case the final return after all/each action check

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,37 +529,52 @@ enum ActivityCancellationType {
...
-
-// Input for the Nexus handler workflow that backs echo-sync and echo-async operations
-message NexusHandlerInput {
-  string input = 1;
-  repeated ActionSet before_actions = 2;
-  string handler_workflow_id = 3;
-  temporal.api.enums.v1.WorkflowIdConflictPolicy handler_workflow_id_conflict_policy = 4;
-  // If true, the handler workflow waits on the "unblock" signal before returning.
-  bool wait_for_signal = 5;
+  NexusOperationRequest input = 3;
```
stephanos: fix: let's move this after field 2; and sort them by field number genrally

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,37 +529,52 @@ enum ActivityCancellationType {
...
+    // Start a standalone activity.
+    ExecuteActivityAction start_activity = 3;
+  }
+}
+
+message NexusWorkflowAction {
+  string workflow_id = 1;
+  string run_id = 2;
+  // Options used when the action may start a workflow.
+  NexusWorkflowStartOptions start_options = 3;
```
stephanos, line 558: This will be used with `Signal` in the next PR; that's why it's not the type for `start`.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,37 +529,52 @@ enum ActivityCancellationType {
   ABANDON = 2;
 }
 
-// Execute a Nexus operation
+// Execute a Nexus operation.
 message ExecuteNexusOperation {
```
stephanos, line 531: These message changes are the main change of the PR.

scenarios/throughput_stress.go
```diff
@@ -892,19 +911,32 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
...
-				Operation: "standalone-activity",
+				Operation: KitchenSinkNexusOperationName,
+				Input:     nexusStandaloneActivityRequest(),
 			},
 		},
 	}
 }
 
-func (t *tpsExecutor) createStandaloneNexusOperationAction(operation string) *Action {
+func nexusStandaloneActivityRequest() *NexusOperationRequest {
```
stephanos: fix: inline nexusStandaloneActivityRequest

workers/python/kitchen_sink.py
```diff
@@ -409,21 +402,3 @@ def convert_act_cancel_type(
         return temporalio.workflow.ActivityCancellationType.ABANDON
     else:
         raise NotImplementedError("Unknown cancellation type " + str(ctype))
-
-
-@workflow.defn
-class NexusHandlerWorkflow:
```
stephanos, line 413: Nexus is now using a kitchensink workflow when starting one.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -61,9 +61,7 @@ message ClientAction {
 // DoStandaloneNexusOperation starts a Nexus operation outside of any workflow context using
 // StartNexusOperationExecution and polls for its completion with PollNexusOperationExecution.
 message DoStandaloneNexusOperation {
-  string endpoint = 1;
-  string service = 2;
-  string operation = 3;
+  ExecuteNexusOperation operation = 1;
```
stephanos, line 64: Breaking proto changes are fine here.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -494,7 +487,11 @@ func handleNexusOperation(ctx workflow.Context, nexusOp *kitchensink.ExecuteNexu
 
 func handleSendSignal(ctx workflow.Context, ws *KSWorkflowState, action *kitchensink.SendSignalAction) error {
 	return withAwaitableChoiceCustom(ctx, ws, func(ctx workflow.Context) workflow.Future {
-		return workflow.SignalExternalWorkflow(ctx, action.WorkflowId, action.RunId, action.SignalName, nil)
```
stephanos, line 497: Wasn't using `Args` before ...

scenarios/throughput_stress.go
```diff
@@ -871,8 +922,13 @@ func (t *tpsExecutor) createNexusAttachCallbacksAction() *Action {
...
 				{Variant: &Action_SendSignal{
 					SendSignal: &SendSignalAction{
-						WorkflowId: handlerWfID,
-						SignalName: "unblock",
+						WorkflowId: workflowID,
+						SignalName: "do_actions_signal",
+						Args: []*common.Payload{ConvertToPayload(&DoSignal_DoSignalActions{
+							Variant: &DoSignal_DoSignalActions_DoActions{
+								DoActions: SingleActionSet(NewEmptyReturnResultAction()),
+							},
```

<details>
<summary><strong>high</strong> — Using the <code>do_actions</code> signal variant leaves the handler workflow running forever on the Python worker, hanging the whole attach-callbacks action.</summary>

The Nexus handler is now a `kitchenSink` workflow started with an empty `WorkflowInput{}`, so the only thing that can complete it is a `ReturnResultAction` delivered by the `do_actions_signal` sent here.

The two SDKs that support Nexus (`nexusUnsupportedSDKs` excludes only Java/Ruby/TypeScript/.NET) treat the two `DoSignalActions` variants differently:

- Go (`workers/go/workerlib/kitchensink/kitchen_sink.go:99-122`) folds both variants into one path and forwards the result to `retOrErrChan`, which completes `run`:
  ```go
  actionSet := sigActions.GetDoActionsInMain()
  if actionSet == nil { actionSet = sigActions.GetDoActions() }
  workflow.Go(ctx, func(ctx workflow.Context) {
      ret, err := state.handleActionSet(ctx, actionSet)
      if ret != nil || err != nil { retOrErrChan.Send(ctx, ReturnOrErr{ret, err}) }
  })
  ```
- Python (`workers/python/kitchen_sink.py:43-48`) only routes `do_actions_in_main` back to the main loop; for `do_actions` it awaits the action set and **discards the return value**:
  ```python
  if signal_actions.HasField("do_actions_in_main"):
      self.action_set_queue.put_nowait(signal_actions.do_actions_in_main)
  else:
      await self.handle_action_set(signal_actions.do_actions)   # return value dropped
  ```

`KitchenSinkWorkflow.run` then stays parked on `await self.action_set_queue.get()`. The handler workflow never returns, so the three `wait_started` operations never resolve and the `AwaitPendingActions` step at the end of this nested set blocks until the caller workflow times out.

The previous implementation avoided this because it signalled `"unblock"` on the dedicated `NexusHandlerWorkflow`, which Python implemented explicitly.

</details>

**Suggestion:** Use the variant both workers route to the main run loop:

```suggestion
						Args: []*common.Payload{ConvertToPayload(&DoSignal_DoSignalActions{
							Variant: &DoSignal_DoSignalActions_DoActionsInMain{
								DoActionsInMain: SingleActionSet(NewEmptyReturnResultAction()),
							},
						})},
```

(Alternatively, make Python's `do_actions` branch propagate its return value the way Go does — but that is a broader behavioural change to the signal contract.)

loadgen/kitchen_sink_executor_test.go
```diff
@@ -944,8 +944,38 @@ func TestKitchenSink(t *testing.T) {
...
+		{
+			name: "NexusOperation/Sync/ExpectedOutputMismatch",
+			testInput: &TestInput{
+				WorkflowInput: &WorkflowInput{
+					InitialActions: ListActionSet(
+						&Action{
+							Variant: &Action_NexusOperation{
+								NexusOperation: &ExecuteNexusOperation{
+									Operation:      KitchenSinkNexusOperationName,
+									ExpectedOutput: ConvertToPayload("goodbye"),
```

<details>
<summary><strong>med</strong> — Only the <em>mismatch</em> branch of <code>expected_output</code> is tested; nothing covers a successful match, so a payload-shape regression would pass CI while breaking every Nexus op in throughput-stress.</summary>

`expected_output` changed from `string` to `temporal.api.common.v1.Payload`, and matching is now exact proto equality (`expectedOutput.Equal(&result)` in Go, `result != nexus_op.expected_output` in Python). That equality only holds because of three separate, non-obvious pieces added in this PR:

1. `clioptions/client.go:173-185` — `PassThroughPayloadConverter.FromPayload` gained a `*common.Payload` special case so the inner payload survives the `_passthrough` wrap/unwrap round trip byte-for-byte.
2. `loadgen/kitchensink/helpers.go:17` — `jsonPayloadConverter` switched from `NewProtoJSONPayloadConverter()` to `GetDefaultDataConverter()`, so `ConvertToPayload("hello")` produces `{encoding: json/plain, data: "\"hello\""}` instead of `nil` (the ProtoJSON converter returns `nil, nil` for non-proto values).
3. Each worker must reproduce that payload exactly — Go via `ConvertToPayload`, Python via a hand-written `Payload(metadata={"encoding": b"json/plain"}, data=json.dumps(...).encode())`, and `kitchen-sink-gen`'s `json_payload()` via `serde_json::to_vec`.

If any of those drifts (an extra metadata key, a different encoding, `messageType` being added), the match silently becomes never-true. This test still passes — it asserts the workflow *fails* — while `createNexusEchoSyncAction` (`scenarios/throughput_stress.go:818`), which sets `ExpectedOutput: ConvertToPayload("hello")`, would fail on every iteration of the main load scenario.

</details>

**Suggestion:** Add `ExpectedOutput: ConvertToPayload("hello")` to the adjacent `NexusOperation/Sync` case so the happy path is asserted alongside the mismatch case (it currently sets no `ExpectedOutput` at all).

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,46 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
...
+		Endpoint: nexusOp.GetEndpoint(),
+		Service:  service,
 	})
 	if err != nil {
-		return fmt.Errorf("NewNexusClient: %w", err)
+		return fmt.Errorf("New standalone Nexus client: %w", err)
 	}
 
-	handle, err := nexusClient.ExecuteOperation(ctx, sno.Operation, &NexusHandlerInput{}, client.StartNexusOperationOptions{
+	handle, err := nexusClient.ExecuteOperation(ctx, nexusOp.GetOperation(), nexusOp.GetInput(), client.StartNexusOperationOptions{
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — Reusing <code>ExecuteNexusOperation</code> here advertises <code>awaitable_choice</code> and <code>headers</code>, but the standalone executor silently drops both.</summary>

`DoStandaloneNexusOperation` used to carry only `endpoint`/`service`/`operation`, so there was nothing to ignore. Now it embeds the full `ExecuteNexusOperation`, and this method consumes only `endpoint`, `service`, `operation`, `input`, and `expected_output`:

- `awaitable_choice` is ignored — this path always behaves as `wait_finish`. A caller setting `Abandon`, `WaitStarted`, or `CancelAfterStarted` gets fully-awaited semantics with no error, which is exactly the kind of silent mismatch the kitchen sink exists to catch.
- `headers` is ignored — `client.StartNexusOperationOptions` has no header field, so it cannot be honoured here at all (the in-workflow Go path drops it too, though Python's `handle_nexus_operation` does forward it).

Nothing sets these today (`createStandaloneNexusOperationAction` leaves both unset), so this is latent rather than currently broken — but the shared message now makes them look supported.

</details>

**Suggestion:** Reject what this path cannot honour, so a future scenario fails loudly instead of quietly running different semantics than it asked for:

```go
if ac := nexusOp.GetAwaitableChoice(); ac != nil && ac.GetWaitFinish() == nil {
	return fmt.Errorf("DoStandaloneNexusOperation only supports the wait_finish awaitable choice")
}
if len(nexusOp.GetHeaders()) > 0 {
	return fmt.Errorf("DoStandaloneNexusOperation does not support headers")
}
```

Alternatively, note the ignored fields in the `DoStandaloneNexusOperation` proto comment.

workers/go/go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/nexus-rpc/sdk-go v0.7.0
 	github.com/prometheus/client_golang v1.18.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.11.1
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — <code>stretchr/testify</code> is promoted to a direct dependency, but nothing in the <code>workers/go</code> module imports it.</summary>

`grep -rn "stretchr/testify" --include="*.go" workers/go` returns nothing — the only test files left in the module are `harness/project_test.go` and `harness/worker_test.go`, neither of which uses it. This hunk is leftover from `workers/go/workerlib/kitchensink/kitchen_sink_test.go`, which you've already noted should be dropped; `go mod tidy` will move testify back to the indirect block.

</details>

**Suggestion:** Run `go mod tidy` in `workers/go` after removing the test file so this hunk disappears from the PR.

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,45 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
 	return run, err
 }
 
-func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, sno *DoStandaloneNexusOperation) error {
+func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Context, sano *DoStandaloneNexusOperation) error {
+	nexusOp := sano.GetOperation()
```
stephanos, line 210: `DoStandaloneNexusOperation` is now a thin wrapper around the actual operation (see proto file).

loadgen/kitchen_sink_executor_test.go
```diff
@@ -944,8 +944,38 @@ func TestKitchenSink(t *testing.T) {
...
+						}),
+				},
+			},
+			historyMatcher: PartialHistoryMatcher(`
+				NexusOperationScheduled {"operation":"execute"}
+				NexusOperationCompleted`),
+			expectedUnsupportedErrs: nexusUnsupportedSDKs,
+		},
+		{
+			name: "NexusOperation/Sync/ExpectedOutputMismatch",
```
stephanos, line 967: Only net new test here.

scenarios/throughput_stress.go
```diff
@@ -581,21 +579,41 @@ func (t *tpsExecutor) createActionsChunk(
 
 		// Add Nexus operations, if configured.
 		if t.config.NexusEnabled {
+			// Invoke a synchronous Nexus operation.
```
stephanos: fix: remove comment

scenarios/throughput_stress.go
```diff
@@ -581,21 +579,41 @@ func (t *tpsExecutor) createActionsChunk(
...
 			asyncActions = append(asyncActions, t.createNexusEchoSyncAction())
-			asyncActions = append(asyncActions, t.createNexusEchoAsyncAction())
+			// Invoke an asynchronous Nexus operation backed by a workflow.
```
stephanos: fix: remove comment

scenarios/throughput_stress.go
```diff
@@ -581,21 +579,41 @@ func (t *tpsExecutor) createActionsChunk(
...
+			asyncActions = append(asyncActions, t.createNexusStartWorkflowAction())
 			asyncActions = append(asyncActions, t.createNexusWaitForCancelAction())
-			asyncActions = append(asyncActions, t.createNexusAttachCallbacksAction())
 			if t.config.IncludeStandaloneNexus {
 				asyncActions = append(asyncActions,
-					t.createStandaloneNexusOperationAction("echo-async"),
-					t.createStandaloneNexusOperationAction("echo-sync"),
+					// async
```
stephanos: fix: remove comment

scenarios/throughput_stress.go
```diff
@@ -581,21 +579,41 @@ func (t *tpsExecutor) createActionsChunk(
...
+								StartOptions: &NexusWorkflowStartOptions{
+									WorkflowInput: &WorkflowInput{
+										InitialActions: ListActionSet(NewEmptyReturnResultAction()),
+									},
+								},
+								Action: &NexusWorkflowAction_Start{Start: &emptypb.Empty{}},
+							},
+						},
+					}),
+					// sync
```
stephanos: fix: remove comment

scenarios/throughput_stress.go
```diff
@@ -581,21 +579,41 @@ func (t *tpsExecutor) createActionsChunk(
...
 				)
 			}
 			if t.config.IncludeNexusStandaloneActivity {
 				asyncActions = append(asyncActions, t.createNexusStandaloneActivityAction())
 				if t.config.IncludeStandaloneNexus {
 					asyncActions = append(asyncActions,
-						t.createStandaloneNexusOperationAction("standalone-activity"),
+						t.createStandaloneNexusOperationAction(&NexusOperationRequest{
+							Action: &NexusOperationRequest_StartActivity{StartActivity: &ExecuteActivityAction{
+								ActivityType: &ExecuteActivityAction_Noop{Noop: &emptypb.Empty{}},
```
stephanos: fix: is `Noop: &emptypb.Empty{}` needed?

scenarios/throughput_stress.go
```diff
@@ -794,22 +812,34 @@ func (t *tpsExecutor) createNexusEchoSyncAction() *Action {
 		Variant: &Action_NexusOperation{
 			NexusOperation: &ExecuteNexusOperation{
 				Endpoint:       t.config.NexusEndpoint,
-				Operation:      "echo-sync",
-				Input:          "hello",
-				ExpectedOutput: "hello",
+				Operation:      KitchenSinkNexusOperationName,
+				ExpectedOutput: ConvertToPayload("hello"),
```
stephanos: fix: let's put `ExpectedOutput` always below `Input` everywhere

scenarios/throughput_stress.go
```diff
@@ -834,77 +875,32 @@ func (t *tpsExecutor) createNexusWaitForCancelAction() *Action {
 	}
 }
 
-// createNexusAttachCallbacksAction exercises Nexus USE_EXISTING callback coalescing:
-// fire N ops with wait_started, signal the handler to unblock, then await its completion.
-func (t *tpsExecutor) createNexusAttachCallbacksAction() *Action {
```
stephanos, line 839: fix: why was this removed?

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -610,69 +603,77 @@ type ReturnOrErr struct {
...
 
-func NexusHandlerWorkflow(ctx workflow.Context, input *kitchensink.NexusHandlerInput) (string, error) {
-	state := KSWorkflowState{
-		workflowState: &kitchensink.WorkflowState{},
-	}
-	for _, actionSet := range input.BeforeActions {
-		if _, err := state.handleActionSet(ctx, actionSet); err != nil {
-			return "", err
+// ExecuteNexusOperation dispatches kitchen sink Nexus actions.
+var ExecuteNexusOperation = temporalnexus.MustNewTemporalOperation(
```
stephanos: fix: do we need this var?

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -610,69 +603,77 @@ type ReturnOrErr struct {
...
-	return input.Input, nil
+	return temporalnexus.TemporalOperationResult[*common.Payload]{}, nexus.HandlerErrorf(
+		nexus.HandlerErrorTypeBadRequest, "Nexus operation request has no supported action set")
 }
 
-// EchoSyncOperation returns the input synchronously without starting a workflow.
-var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Context, input *kitchensink.NexusHandlerInput, opts nexus.StartOperationOptions) (string, error) {
-	if len(input.BeforeActions) > 0 {
-		return "", nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "before_actions not supported in echo-sync")
+func nexusWorkflowOptions(input *kitchensink.NexusWorkflowAction, requestID string) client.StartWorkflowOptions {
```
stephanos: fix: should this be inlined?

THardy98:
Is this RFR? diff is massive

loadgen/kitchensink/client_action_executor.go
```diff
@@ -205,27 +206,45 @@ func (e *ClientActionsExecutor) executeUpdateAction(ctx context.Context, upd *Do
...
 	operationID := fmt.Sprintf("standalone-nexus-%s-%s", e.WorkflowOptions.ID, uuid.NewString())
 	nexusClient, err := e.Client.NewNexusClient(client.NexusClientOptions{
-		Endpoint: sno.Endpoint,
-		Service:  sno.Service,
+		Endpoint: nexusOp.GetEndpoint(),
+		Service:  KitchenSinkNexusServiceName,
 	})
 	if err != nil {
-		return fmt.Errorf("NewNexusClient: %w", err)
+		return fmt.Errorf("New standalone Nexus client: %w", err)
```
mavemuri: minor-nit/question: iirc capitalized errors are discouraged - but unsure if there a repo-level override

scenarios/throughput_stress.go
```diff
@@ -820,10 +849,21 @@ func (t *tpsExecutor) createNexusWaitForCancelAction() *Action {
 		Variant: &Action_NexusOperation{
 			NexusOperation: &ExecuteNexusOperation{
 				Endpoint:  t.config.NexusEndpoint,
-				Operation: "echo-async",
-				BeforeActions: ListActionSet(
-					NewAwaitWorkflowStateAction("never", "resolves"),
-				),
+				Operation: KitchenSinkNexusOperationName,
```
mavemuri, line 852: Question: does it make sense to leave room here to supply custom operations as well? eg. operation inter-op kinds of test cases etc. Not sure if the singular `KitchenSinkNexusOperationName` is too broad

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -529,35 +527,48 @@ enum ActivityCancellationType {
...
+
+// Input for the kitchen sink Nexus operation.
+message NexusOperationRequest {
+  oneof action {
+    // Return this value synchronously.
+    string echo = 1;
+    // Act on a kitchenSink workflow.
+    NexusWorkflowAction workflow_action = 2;
+    // Start a standalone activity.
+    ExecuteActivityAction start_activity = 3;
```
mavemuri, line 550: Question: do we need anything for plain workflowRun/activity too?

mavemuri (approved):

loadgen/kitchensink/client_action_executor.go
stephanos: Good point; addressed!

scenarios/throughput_stress.go
stephanos: `KitchenSinkNexusOperationName` maps to the new `kitchensink.KitchenSinkNexusOperation`; if we need new operation handlers we can just add new ones, right?

workers/proto/kitchen_sink/kitchen_sink.proto
stephanos: Can you define "plain"? This now allows to start a workflow and a standalone activity; what's missing?

Note that this is scoped to actions from Nexus handlers; regular workflows and activities are defined somewhere else.

mavemuri: Right I did mean for nexus-triggered ops. I might be reading this wrong - NexusWorkflowAction - thought this was for operations on top of workflows - signal/query/update, not for a nexus op workflow run itself

scenarios/throughput_stress.go
mavemuri: Yes was thinking if for that reason maybe Operation should come from t.config instead 

But not a blocker

workers/proto/kitchen_sink/kitchen_sink.proto
stephanos: `workflow_action.start` is the plain Nexus-backed workflow-run case: the handler starts `KitchenSinkWorkflow` and returns that run. There aren’t any signal/query/update actions here yet, see next PR.

Open to naming suggestions if that helps to make anything clearer!

scenarios/throughput_stress.go
stephanos: I think this way is preferable since the main thing is to ensure that the operation name that the client and the worker use matches. We could add the option to rename it, but I'm not sure yet how that'd be valuable. The kitchensink workflow type is the same; ie it's hard-coded in the code, too.

stephanos:
@THardy98 Well, 95% of this is generated code that changed ... I marked the [main change of this PR here](https://github.com/temporalio/omes/pull/361#discussion_r3939153205). This is the smallest I could make the diff. I moved all other changes to stacked PRs to keep this one "small".

I can provide more context or guidance, if that's helpful.

PR #370, opened 2026-05-28
Promote cmd/clioptions to top-level clioptions/

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * #375
 * #374
 * #373
 * #372
 * #371
 * __->__#370

--- --- ---

### Promote cmd/clioptions to top-level clioptions/

clioptions is the most-imported package in the repo (used across both the
root module and the workers/go module), but it lived under cmd/, which by
convention is reserved for main packages (binary entry points). Move it to
a top-level clioptions/ and rewrite the import path accordingly.

Testing: `go build ./...` and `go vet ./...` pass for both the root and
workers/go modules.


lilydoar:

PR #371, opened 2026-05-28
Restructure cmd/ into conventional cmd/omes + cmd/dev layout

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * #375
 * #374
 * #373
 * #372
 * __->__#371
 * #370

--- --- ---

### Restructure cmd/ into conventional cmd/omes + cmd/dev layout

User-facing change: the CLI is now invoked as `./cmd/omes` (previously
`./cmd`) -- e.g. `go run ./cmd/omes run-scenario ...`. README, CI, the worker
Dockerfiles, and cmd/dev/test.go are updated; any external scripts that call
`go run ./cmd` must switch to `./cmd/omes`.

Per Go convention, cmd/ is a container of binary subdirectories, each named
after the binary it produces. Previously the omes binary's main lived at the
cmd/ root (built via ./cmd, producing a binary named 'cmd' that Docker had to
rename with -o), with its implementation factored into a separate package cli
plus a one-line cmd/main.go shim purely to keep the cmd/ root tidy.

Move the implementation into cmd/omes as package main (func main), delete the
cli package indirection and the shim. The omes binary now builds via
./cmd/omes and is named 'omes' automatically.

Testing: `go build ./...` passes; `go run ./cmd/omes --help` and
`go run ./cmd/omes list-scenarios` work.

PR #372, opened 2026-05-28
Extract workers test helpers into internal/workertest

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * #375
 * #374
 * #373
 * __->__#372
 * #371
 * #370

--- --- ---

### Extract workers test helpers into internal/workertest

workers/test_env.go, test_workers.go, and test_historyrequire.go were
package workers (production) files: the test_ prefix is not special to Go
(only _test.go is excluded from builds), so this test-support code compiled
into the shipped binary and forced production-only edges from package workers
into loadgen/devserver/versions that only the test harness needs.

Move them to a dedicated internal/workertest package (env.go, workerpool.go,
historyrequire.go). The helpers reference a few production symbols (Builder,
Runner, BaseDir, and the previously-unexported logWriter), so export logWriter
as LogWriter with a NewLogWriter constructor and qualify the rest. Update the
consumers (loadgen and scenario tests) to import internal/workertest.

Testing: `go build ./...` and `go vet ./...` pass; the loadgen and scenario
test packages compile against the new import path.

### DIFF
--- a/internal/workertest/env.go
+++ b/internal/workertest/env.go
@@ -1,4 +1,4 @@
-package workers
+package workertest
 
 import (
 	"context"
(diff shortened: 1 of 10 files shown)


lilydoar:

PR #373, opened 2026-05-28
Move workers orchestration package to internal/workerctl

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * #375
 * #374
 * __->__#373
 * #372
 * #371
 * #370

--- --- ---

### Move workers orchestration package to internal/workerctl

The root workers package (build.go, log.go, run.go) is application logic for
building and running worker subprocesses, with no consumer outside the root
module. Application code that isn't part of a public API belongs under
internal/, where the import boundary is compiler-enforced. It also could not
stay at workers/, since that directory is the documented polyglot
worker-source home (workers/<lang>/...).

Move it to internal/workerctl (package workerctl) and update importers. The
workers/ directory now contains only the per-language worker source trees.

Testing: `go build ./...` and `go vet ./...` pass for both modules.

### DIFF
--- a/cmd/omes/prepare_worker.go
+++ b/cmd/omes/prepare_worker.go
@@ -6,7 +6,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/temporalio/omes/clioptions"
-	"github.com/temporalio/omes/workers"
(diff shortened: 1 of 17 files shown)


lilydoar:

PR #374, opened 2026-05-28
Move devserver to internal/devserver

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * #375
 * __->__#374
 * #373
 * #372
 * #371
 * #370

--- --- ---

### Move devserver to internal/devserver

devserver (the embedded Temporal dev-server launcher) has no consumer outside
the root module. Application code that isn't part of a public API belongs
under internal/, where the import boundary is compiler-enforced. Move it and
update the two importers.

Testing: `go build ./...` passes; the internal/devserver unit tests
(TestAllocatePorts, TestDefaultOutputDirUsesRepoRoot, TestWriteDynamicConfig,
...) pass.

PR #375, opened 2026-05-28
Move versions to internal/versions

lilydoar:
Stacked PRs:
 * #378
 * #377
 * #376
 * __->__#375
 * #374
 * #373
 * #372
 * #371
 * #370

--- --- ---

### Move versions to internal/versions

versions (the versions.env reader) has no consumer outside the root module.
Application code that isn't part of a public API belongs under internal/,
where the import boundary is compiler-enforced.

versions.go locates versions.env at runtime by walking up from its own source
path (runtime.Caller), so the added directory level requires one more
filepath.Dir to still reach the repo root. Verified at runtime that
versions.env still resolves after the move.


lilydoar:

PR #376, opened 2026-05-28
Narrow cli.Dockerfile's workers COPY to what's actually needed

lilydoar:
Stacked PRs:
 * #378
 * #377
 * __->__#376
 * #375
 * #374
 * #373
 * #372
 * #371
 * #370

--- --- ---

### Narrow cli.Dockerfile's workers COPY to what's actually needed

The CLI image copied the entire workers/ tree (all language worker sources)
but only needs workers/proto (for the kitchen-sink-gen protoc build) and
workers/go/harness/api (the go.mod replace target for the ./cmd/omes build).
Copy just those two, trimming the build context and making the dependency
explicit. Verified by building the CLI image with podman.


lilydoar:

PR #377, opened 2026-05-28
Cache Go dependency downloads in cli.Dockerfile

lilydoar:
Stacked PRs:
 * #378
 * __->__#377
 * #376
 * #375
 * #374
 * #373
 * #372
 * #371
 * #370

--- --- ---

### Cache Go dependency downloads in cli.Dockerfile

Previously the Dockerfile copied source then go.mod/go.sum and built in one
shot, so any source change invalidated the layer and re-downloaded every
module. Copy go.mod/go.sum (plus the workers/go/harness/api replace target,
required for the module graph to resolve) first and run `go mod download` as
its own layer, then copy source and build. Dependency downloads now cache
independently of source edits. Verified by building the CLI image with podman.


lilydoar:

PR #378, opened 2026-05-28
Normalize the CLI-build block across all Dockerfiles

lilydoar:
Stacked PRs:
 * __->__#378
 * #377
 * #376
 * #375
 * #374
 * #373
 * #372
 * #371
 * #370

--- --- ---

### Normalize the CLI-build block across all Dockerfiles

All seven Dockerfiles build the omes CLI with the same set of inputs, but the
duplicated block had drifted: java had metrics/scenarios in a different order,
dotnet folded an unrelated workers/proto/harness copy into it, typescript
appended versions.env, and the build invoked go via different paths
(go vs /usr/local/go/bin/go). That divergence is what let the earlier COPY-set
regression hide.

Unify them on one byte-identical block (the cached go-mod-download pattern):
- ensure go is on PATH in every image so the build line is identical
- move per-language extras out of the shared block into each file's
  language-specific section (cli: workers/proto for kitchen-sink-gen;
  typescript: versions.env; dotnet: workers/proto/harness)

The shared block is now identical across all seven (verified by hash). Built
cli, go, and ruby images with podman to cover both base-image patterns
(golang-native and go-installed-on-PATH).

### DIFF
--- a/dockerfiles/cli.Dockerfile
+++ b/dockerfiles/cli.Dockerfile
@@ -32,21 +32,18 @@ COPY go.mod go.sum ./
 COPY workers/go/harness/api ./workers/go/harness/api
 RUN go mod download
 
-# Copy CLI source. kitchen-sink-gen needs the proto tree; the rest of workers/
-# (language workers) is not needed here.
+# Copy CLI source and build the CLI.
 COPY cmd ./cmd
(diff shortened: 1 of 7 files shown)


lilydoar:

PR #384, opened 2026-05-29
Move workers orchestration package to internal/workerctl

lilydoar:
Stacked PRs:
 * #388
 * #387
 * #386
 * #385
 * __->__#384

--- --- ---

### Move workers orchestration package to internal/workerctl

The root workers package (build.go, log.go, run.go) is application logic for
building and running worker subprocesses, with no consumer outside the root
module. Application code that isn't part of a public API belongs under
internal/, where the import boundary is compiler-enforced. It also could not
stay at workers/, since that directory is the documented polyglot
worker-source home (workers/<lang>/...).

Move it to internal/workerctl (package workerctl) and update importers. The
workers/ directory now contains only the per-language worker source trees.

Testing: `go build ./...` and `go vet ./...` pass for both modules.

### DIFF
--- a/cmd/omes/prepare_worker.go
+++ b/cmd/omes/prepare_worker.go
@@ -6,7 +6,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/temporalio/omes/clioptions"
-	"github.com/temporalio/omes/workers"
+	"github.com/temporalio/omes/internal/workerctl"
 )
 
 func prepareWorkerCmd() *cobra.Command {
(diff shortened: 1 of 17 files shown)


THardy98 (approved):
Very happy to see this change. What is the `ctl` in `workerctl` ?

THardy98 (changes requested):
I just merged:
https://github.com/temporalio/omes/pull/395

which adds `env.go` and `env_test.go`, you may want move this into `workerctl` (or where you think is the best fit).

They are also tiny so I'm okay if we fold them into an existing file - sorry for the additional work.

lilydoar:
> Very happy to see this change. What is the `ctl` in `workerctl` ?

"ctl" stands for "control" in cli abbreviations. Same thing you'd see in a tool like `kubectl`.

PR #403, opened 2026-06-10
Add standalone-activity support to throughput_stress (RE-353)

lilydoar:
## What was changed

Adds an `include-standalone-activity` scenario option to `throughput_stress` that starts activities **outside any workflow context** via `StartActivityExecution` and waits for the outcome with `PollActivityExecution` — mirroring the `include-standalone-nexus` pattern (#339) and bench-go's standalone-activity throughputstress step (bench-go #357/#358).

- `kitchen_sink.proto`: new `ClientAction` variant `DoStandaloneActivity` (field 7).
- Go client-action executor: Start + Poll implementation (30s start-to-close, 3 attempts), targeting the existing `noop` activity on the workflow's task queue. Runs for both the loadgen runner and the Go worker (shared executor).
- Java/Python/Ruby/TS/.NET workers raise `UnsupportedOperation`, exactly like standalone Nexus.
- Bindings regenerated via `go run ./cmd/dev build-proto` (Go, Java, Python, .NET).

## Why

bench-go grew an `EnableStandaloneActivity` toggle (used by saas-cicd chasm e2e pipelines, with server dynamic config `activity.enableStandalone`) after the bench→omes parity scan closed. omes had no standalone-activity coverage at all — this closes that gap so the saas-cicd e2e cutover ([RE-264](https://temporalio.atlassian.net/browse/RE-264)) isn't blocked on it. [RE-353](https://temporalio.atlassian.net/browse/RE-353), part of epic RE-257.

## Notes for review

- **Ruby generated proto** (`workers/ruby/.../kitchen_sink_pb.rb`) is not covered by `dev build-proto` — if it's generated by a separate step, it needs a follow-up regen (the Ruby executor stub is in place).
- Server must support standalone activities; on unsupported servers the action fails — bench-go added a "short-circuit to success" for that case (bench-go `bdd2879`), which I deliberately did *not* replicate: the option is opt-in, so failing loudly on unsupported servers seems preferable. Open to changing.
- Validated: `go build`, `go test ./scenarios/ ./loadgen/kitchensink/` pass. Not yet exercised against a server with `activity.enableStandalone` — needs a verification run before un-drafting.

[RE-264]: https://temporalio.atlassian.net/browse/RE-264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lilydoar:
**Verification run (local, server 1.31.0 via `temporal server start-dev`):**

```
go run ./cmd/omes run-scenario-with-worker --scenario throughput_stress --language go \
  --run-id verify353 --iterations 2 --option include-standalone-activity=true \
  --option internal-iterations=2 --server-address localhost:7233 --namespace default
```

Scenario completed cleanly, and `temporal activity list` confirms the standalone path was exercised end-to-end — 4 standalone activities (2 iterations × 2 internal iterations), all **Completed**, type `noop`, IDs matching the executor's scheme:

```
Completed  standalone-activity-w-verify353-7176d07159bb5746-1-a9dbbf7e-...  noop
Completed  standalone-activity-w-verify353-7176d07159bb5746-2-273892df-...  noop
...
```

One caveat: `start-dev` accepts standalone-activity APIs whether or not `--dynamic-config-value activity.enableStandalone=true` is passed (a control run without the flag also succeeded), so this validates the omes side but not the server-side gating — cloud/e2e environments still need `activity.enableStandalone` set, as the README section notes.

Also pushed: regenerated Ruby kitchen-sink bindings (`workers/ruby/protos/.../kitchen_sink_pb.rb`) — `build-proto` doesn't cover Ruby, so this was done with `protoc --ruby_out` plus the same `temporalio/api/...` require rewrite the file already used.

GregoryTravis (approved):
Approving for Java and Ruby

lilydoar:
Reworked this PR to fold in the advantages of #351 (thanks @prathyushpv) while keeping this PR's `include-standalone-activity` flag and README conventions. After comparing both PRs, #351's approach is strictly better on several axes, so I adopted it here rather than maintaining two competing implementations.

**What changed from the original version of this PR:**

- **proto:** `DoStandaloneActivity` now carries `ExecuteActivityAction activity = 1` instead of a bare `string activity_type`. This reuses the existing DSL message, so the activity variant, task queue, timeouts, and retry policy are all configurable rather than hardcoded.
- **Go executor:** switched from the raw `StartActivityExecution`/`PollActivityExecution` RPCs to the high-level SDK `client.ExecuteActivity` + `StartActivityOptions` + `handle.Get`. Idiomatic, ~15 fewer lines, and drops the manual `Namespace` threading.
- **Dedup:** extracted `ActivityNameAndArgs` + `ConvertFromPBRetryPolicy` into shared helpers and pointed the Go worker's `launchActivity` at them.
- **Load shape:** `throughput_stress` now runs a `payload` activity (256B in/out) instead of `noop`, matching bench-go.
- **Cross-language coverage:** implemented the standalone-activity client action for **Python, TypeScript, .NET, and Java** workers (each via its SDK's high-level standalone-activity client API + a shared `ActivityDispatch` helper). Ruby remains stubbed. Previously this PR only supported the Go worker.
- **Java SDK** bumped to 1.35.0 (required for `ActivityClient`/`StartActivityOptions`).

**Verified builds:** Go (`build` + `test`), Python (`mypy`), TypeScript (`tsc --noEmit`), .NET (`dotnet build`), Java (`gradlew build` incl. Spotless).

**Still outstanding before un-drafting:** not yet exercised against a live server with `activity.enableStandalone` — that verification run is the remaining step.

workers/dotnet/WorkerLib/KitchenSink/ClientActionsExecutor.cs
```diff
@@ -79,6 +79,10 @@ private async Task ExecuteClientAction(ClientAction action)
             throw new ApplicationFailureException(
                 "DoStandaloneNexusOperation is not supported", "UnsupportedOperation", nonRetryable: true);
         }
+        else if (action.DoStandaloneActivity != null)
```
THardy98, line 87: do we also need to check if this is explicitly false

loadgen/kitchensink/client_action_executor.go
THardy98, line 1: Just as an aside - it's a bit odd that this file is located in `loadgen/kitchensink`, the`ClientActionsExecutor` is worker logic, it doesn't generate any load.

(not really related to your PR, but just noting it)

workers/ruby/workerlib/kitchensink/client_action_executor.rb
```diff
@@ -51,6 +51,11 @@ def execute_client_action(action)
         'DoStandaloneNexusOperation is not supported',
         non_retryable: true
       )
+    when :do_standalone_activity
+      raise Temporalio::Error::ApplicationError.new(
+        'DoStandaloneActivity is not supported',
+        non_retryable: true
+      )
```
THardy98: standalone activities is supported in ruby SDK version 1.5.0 (see https://github.com/temporalio/sdk-ruby/pull/443)

in any case, if there's a feature that's not supported by a particular SDK, we should check `err-if-unimplemented` here and log instead of throwing if set

THardy98 (changes requested):
Couple small things but Ruby support is blocking

workers/dotnet/WorkerLib/KitchenSink/ClientActionsExecutor.cs
lilydoar: `DoStandaloneActivity` is a message-typed oneof member, so the generated property is the message or null — != null is the "is this variant set" check (there's no false/default case to guard as there would be for a bool), and it matches every other branch in this method.

Basically, this isn't a bool, it is either null, or a valid value

workers/ruby/workerlib/kitchensink/client_action_executor.rb
lilydoar: Added the checks for `err-if-unimplemented`

THardy98: any reason why we didn't add Ruby support ?

lilydoar: The reason it was stubbed: the Ruby worker's Gemfile.lock/gemspec were pinned to temporalio 1.3.0, which predates the standalone-activity API. But that was actually a latent inconsistency from #404 — that PR bumped versions.env to RUBY_SDK_VERSION=1.5.0 but left the lock/gemspec at 1.3.0.

Adding it now

prathyushpv (approved):

THardy98 (approved):
Would prefer if we could add

PR #408, opened 2026-06-18
Add failure-tolerant run mode (ContinueOnIterationFailure)

lilydoar:
## What

Adds an opt-in, failure-tolerant run mode to the generic executor. `ContinueOnIterationFailure` changes **only when a run stops** — it runs every iteration to completion instead of aborting on the first terminal failure — and surfaces per-iteration success/failure counts so a caller can layer its own policy on top. The **verdict contract is unchanged**: if any iteration terminally fails, the run still returns a failure.

- **`loadgen` — `RunConfiguration.ContinueOnIterationFailure`** (default `false`, behavior unchanged when unset): when set, an iteration that terminally fails (after exhausting `MaxIterationAttempts`) no longer aborts the run; the executor keeps starting new iterations until the duration/iteration limit is reached. The failure is not ignored — it is counted, and the run **still fails at the end** if any iteration failed.
- **`loadgen` — `RunConfiguration.OnIterationFailure` hook**: invoked when an iteration terminally fails, mirroring the existing `OnCompletion` hook. It is how a caller tallies failures. The executor also tracks succeeded/failed counts and returns a failure verdict (`run completed with N of M iterations failed`) at end-of-run when the failed tally is nonzero.
- **`scenarios/throughput_stress`**: tallies terminal failures into a new `FailedIterations` field carried in the snapshot (alongside `CompletedIterations`). It skips the binary zero-failed-workflow gate (`VerifyNoFailedWorkflows`) when tolerating failures — that check yields no count, so the counted iteration-failure tally is the authoritative signal; the block is only reached when no iteration failed anyway, since a tolerated failure makes the executor return a failure verdict and the scenario returns early.

**Why the counts, if the run still fails?** The default verdict (any failure ⇒ fail) is preserved so nothing silently passes. A caller that wants a *more granular* verdict — e.g. "the load stayed up and fewer than X% of iterations failed" — reads `failedIterations`/`completedIterations` from the snapshot and applies its own threshold on top. omes provides the mechanism and the counts; the tolerance policy lives with the caller.

## Why

Part of the bench-go → omes migration ([RE-257](https://temporalio.atlassian.net/browse/RE-257)). The release-validation suite intentionally disrupts the cell (rollout / rolling-restart / chaos / scale-to-zero), so the standing load hits transient iteration failures that must not abort the whole run the way the current all-or-nothing behavior does. This lets the run continue through the disruption and report the failure counts, while still failing by default — the foundation for a failure-rate verdict.

## Testing

- `go test ./loadgen/ -run 'TestRun|TestExecutor'` — all executor tests pass, including a new `TestRunContinueOnIterationFailure`: with `ContinueOnIterationFailure` set and half the iterations failing, every iteration runs and both hooks fire for the right iterations, **and the run still returns a failure** (`3 of 6 iterations failed`).
- `go build ./loadgen/... ./scenarios/... ./cmd/...` — clean. `go vet` shows only pre-existing, unrelated notes.
- Default-path behavior is unchanged: with `ContinueOnIterationFailure` unset, a terminal failure aborts the run exactly as before (covered by existing `TestRunFailIterations`/`TestRunFailDuration`).

## Consumer / follow-up

The saas-cicd side that consumes this — reading `failedIterations` into the load result and applying the release-validation verdict from the counts — is a companion PR. Note the saas-cicd branch currently builds against this omes only under the `protoregistry.conflictPolicy=warn` ldflag workaround for the nexus-rpc proto annotation collision (ext 8233) with the bergundy annotations embedded in `saas-infra-plane/api`; the durable fix (saas-infra-plane regenerating against `nexus-rpc/nexus-proto-annotations`) is tracked separately.

[RE-257]: https://temporalio.atlassian.net/browse/RE-257

### DIFF
--- a/loadgen/generic_executor.go
+++ b/loadgen/generic_executor.go
@@ -3,6 +3,7 @@ package loadgen
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"go.temporal.io/sdk/client"
@@ -21,6 +22,10 @@ type genericRun struct {
 	logger   *zap.SugaredLogger
 	// Timer capturing E2E execution of each scenario run iteration.
 	executeTimer client.MetricsTimer
+	// Iteration outcome tallies, used for the end-of-run summary. failed counts
+	// only terminal failures tolerated via ContinueOnIterationFailure.
+	completed atomic.Int64
+	failed    atomic.Int64
 }
 
 func (g *GenericExecutor) Run(ctx context.Context, info ScenarioInfo) error {
@@ -130,11 +135,28 @@ func (g *genericRun) Run(ctx context.Context) error {
 			defer func() {
 				g.executeTimer.Record(time.Since(iterStart))
 
+				// Swallow the error for the spawn loop only, so a tolerated failure
+				// doesn't stop new iterations from starting. It is not ignored: it is
+				// counted (g.failed) and surfaced via OnIterationFailure, and the run
+				// still fails at the end if the tally is nonzero.
+				iterErr := err
+				if iterErr != nil && g.config.ContinueOnIterationFailure {
+					err = nil
+				}
+
 				select {
 				case <-ctx.Done():
 				case doneCh <- err:
-					if err == nil && g.config.OnCompletion != nil {
-						g.config.OnCompletion(ctx, run)
+					if iterErr == nil {
+						g.completed.Add(1)
+						if g.config.OnCompletion != nil {
+							g.config.OnCompletion(ctx, run)
+						}
+					} else {
+						g.failed.Add(1)
+						if g.config.OnIterationFailure != nil {
+							g.config.OnIterationFailure(ctx, run, iterErr)
+						}
 					}
 				}
 			}()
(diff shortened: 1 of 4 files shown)


loadgen/generic_executor.go
```diff
@@ -130,11 +135,28 @@ func (g *genericRun) Run(ctx context.Context) error {
...
 				g.executeTimer.Record(time.Since(iterStart))
 
+				// Swallow the error for the spawn loop only, so a tolerated failure
+				// doesn't stop new iterations from starting. It is not ignored: it is
+				// counted (g.failed) and surfaced via OnIterationFailure, and the run
+				// still fails at the end if the tally is nonzero.
+				iterErr := err
+				if iterErr != nil && g.config.ContinueOnIterationFailure {
+					err = nil
+				}
```
bshore-temporal, line 145: The errors might be helpful in diagnosing erratic / flapping / non-deterministic failures.  Can they be accumulated, maybe as a multi-error?  if the count is too high, maybe something like "the last 20 errors" could be surfaced.  This could be a future feature.

bshore-temporal (approved):
👍

loadgen/generic_executor.go
lilydoar: https://temporalio.atlassian.net/browse/RE-514

PR #410, opened 2026-06-23
Add nexus-operation-with-standalone-activity to throughput_stress

tekkaya:
## What was changed

Adds an opt-in `throughput_stress` feature option, `--option include-nexus-standalone-activity`, that exercises **a standalone activity backed Nexus operation** and resolves the operation via the activity's **completion callback**. The single operation is driven through **two invocation paths**:

- **In-workflow Nexus operation** — the kitchen sink workflow invokes it via the `ExecuteNexusOperation` action (operation `standalone-activity`).
- **Standalone Nexus operation** — invoked outside any workflow via the `DoStandaloneNexusOperation` client action. Added only when standalone Nexus is already part of the run.

## Why

`throughput_stress` already covered standalone activities and standalone Nexus operations separately, but not the standalone activity backed Nexus operation which is resolved by the activity's completion callback. 

## Notes for review

- **Server requirements:** standalone activities (`activity.enableStandalone`), activity completion callbacks (`activity.enableCallbacks`), and a Nexus callback URL (`component.nexusoperations.useSystemCallbackURL`, which defaults on). These are dynamic-config gates, so they must be enabled on every server version the load reaches — including all versions in a rolling-upgrade (mixed-version) run.
- **Verification:** both modules build + vet; `TestThroughputStressNexusStandaloneActivity` passes end-to-end against the dev server (standalone activities + activity completion callbacks enabled); the full existing `TestThroughputStress*` suite still passes.

### DIFF
--- a/docs/throughput-stress.md
+++ b/docs/throughput-stress.md
@@ -89,3 +89,16 @@ is any.
 | `=false` | either | no Nexus load; standalone Nexus not included |
 
 Asking for `include-standalone-nexus=true` while Nexus is off is a contradiction, and fails the run.
+
+## Nexus operation with a standalone activity
+
+`include-nexus-standalone-activity` adds a standalone activity backed Nexus operation.
+It is driven two ways each iteration: As an in-workflow Nexus operation, and — when standalone Nexus is part of the run —
+as a standalone Nexus operation.
+
+This is **opt-in and off by default**; pass `--option include-nexus-standalone-activity=true`.
+It requires `nexus-enabled` and also needs server support for standalone activities and activity
+completion callbacks (dynamic config `activity.enableStandalone` and `activity.enableCallbacks`) and a
+Nexus callback URL; if those are off the operation fails clearly rather than being skipped.
+
+Currently only supported and run by Go workers.
--- a/workers/go/apps/worker/worker.go
+++ b/workers/go/apps/worker/worker.go
@@ -34,6 +34,7 @@ func buildWorker(client sdkclient.Client, context harness.WorkerContext) sdkwork
 	for _, op := range []nexus.RegisterableOperation{
 		kitchensink.EchoSyncOperation,
 		kitchensink.EchoAsyncOperation,
+		kitchensink.StandaloneActivityNexusOperation,
 	} {
 		if err := service.Register(op); err != nil {
 			panic(err)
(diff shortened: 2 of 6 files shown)


workers/go/apps/worker/worker.go
```diff
@@ -34,6 +34,7 @@ func buildWorker(client sdkclient.Client, context harness.WorkerContext) sdkwork
 	for _, op := range []nexus.RegisterableOperation{
 		kitchensink.EchoSyncOperation,
 		kitchensink.EchoAsyncOperation,
+		kitchensink.StandaloneActivityNexusOperation,
```
S15, line 37: The lambda worker registers the other two, should it also register `StandaloneActivityNexusOperation`?

S15, line 37: I think the python worker will also try and run the workflow version of SAA+Nexus, and fail.

Quinn-With-Two-Ns: I think the Go lambda worker has the same issue

docs/throughput-stress.md
```diff
@@ -77,3 +77,16 @@ is any.
...
 
 Asking for `include-standalone-nexus=true` while Nexus is off is a contradiction, and fails the run.
+
+## Nexus operation with a standalone activity
+
+`include-nexus-standalone-activity` adds a standalone activity backed Nexus operation.
+It is driven two ways each iteration: As an in-workflow Nexus operation, and — when standalone Nexus is part of the run —
+as a standalone Nexus operation.
+
+Like standalone Nexus it only adds to a run already generating Nexus load, so it follows `nexus-enabled`.
```
Quinn-With-Two-Ns: If the server doesn't support attaching callbacks does the Nexus Operation handler skip or fail clearly?

workers/go/apps/worker/worker.go
tekkaya: Good catch, thanks. Registered `StandaloneActivityNexusOperation` on the lambda worker alongside the existing echo operations.

tekkaya: Correct, the standalone path skips non-Go workers, but the in-workflow path does not, so a Python worker would fail. I fixed this by making `include-nexus-standalone-activity` opt-in and off by default. I also tested it with a Python worker: the default configuration works as expected; explicitly enabling the option fails with “operation not found” error.

docs/throughput-stress.md
tekkaya: Correct, it will fail rather than skip; omes has no graceful-skip path for it. With the updated changes, the handler maps failures caused by disabled standalone activities (`Unimplemented`) or disabled activity completion callbacks (`InvalidArgument`) to a non-retryable BadRequest handler error. The error also mentions the required dynamic configs: activity.enableStandalone or activity.enableCallbacks.

Quinn-With-Two-Ns (approved):

PR #411, opened 2026-06-26
Error-handling cleanup: surface swallowed failures, branch on values not strings

lilydoar:
## What

A focused error-handling cleanup across the omes Go code, prompted by an audit against our Go error-handling conventions (don't drop errors, log XOR return, branch on values not strings, propagate at boundaries). Each fix is its own commit. Scope was limited to **High** and **Medium** severity findings — bugs and correctness issues — not the broad stylistic layer (noise-prefix messages, capitalized error strings).

## Changes (one commit each)

**1. Return dropped search-attribute registration error (`scenarios/ebb_and_flow.go`)**
`spawnWorkflowWithActivities` called `RegisterDefaultSearchAttributes(ctx)` and ignored its `error`. A registration failure was lost and only surfaced later as an obscure workflow-start error. The sibling caller in `generic_executor.go` already checks this return value.

**2. Surface failed `ebbAndFlowTrack` workflows instead of swallowing them (`scenarios/ebb_and_flow.go`)**
`wf.Get` failures were logged and then `return nil`, so they never reached the `MaxConsecutiveErrors` guard, and a failed workflow was still counted as completed — inflating `TotalCompletedWorkflows`, which post-scenario verification relies on. Now the error is returned and the completion count is gated on success. The backlog still drains on either outcome, so the rate controller's behavior is unchanged.

**3. Fix dropped terminate error and pointer log (`loadgen/scenario.go`)**
In `ExecuteKitchenSinkWorkflow`'s client-actions goroutine, the failure log printed the `atomic.Pointer[error]` value (an address) rather than the error, and the result of the cleanup `TerminateWorkflow` call was silently discarded. Now it logs the actual error, and logs the terminate failure (the goroutine has no return path to propagate it).

**4. Stop double-handling terminal iteration errors (`loadgen/generic_executor.go`)**
The per-iteration goroutine logged a terminal failure *and* sent it on `doneCh`, where `Run` returns it wrapped for its caller to report — the same error handled twice. The terminal branch now propagates only. The retry-branch log is kept on purpose: that error is superseded by the retry and never returned, so the log is its only record.

**5. Branch on gRPC status code, not error string (`loadgen/scenario.go`)**
`RegisterDefaultSearchAttributes` decided whether to ignore an "already exists" failure via `strings.Contains(err.Error(), ...)`. Replaced with `status.Code(err) == codes.AlreadyExists`, matching `ensureNexusEndpoint` in the same file. The second match string was a typo (`"unavailble"`) that could never match the real message, so no working behavior is lost.

**6. Wrap cert-load error with `%w` (`clioptions/client.go`)**
`loadTLSConfig` used `%s` to format the cert-load error, severing the unwrap chain at this library-ish boundary. Switched to `%w`, matching the two sibling wraps in the same function.

## Testing

- `go build` clean across all first-party packages.
- `go test ./clioptions/ ./scenarios/` — pass.
- `go test ./loadgen/` worker-free unit tests (cover the `generic_executor` change) — pass.
- `SDK=go go test ./loadgen/ -run TestKitchenSink` — pass (`ok … 25.5s`); exercises the `ExecuteKitchenSinkWorkflow` and `RegisterDefaultSearchAttributes` paths against a real devserver.

Note: a no-`SDK` `go test ./loadgen/` fails locally because `TestKitchenSink` runs the full SDK matrix (Go/Java/Python/Ruby/TS/.NET) and only the Go worker toolchain is built here. That's environmental and pre-existing — CI sets `SDK` per the test's own skip logic.

### DIFF
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -404,22 +404,10 @@ func (s *ScenarioInfo) RegisterDefaultSearchAttributes(ctx context.Context) erro
 		},
 		Namespace: s.Namespace,
 	})
-	// Throw an error if the attributes could not be registered, but ignore already exists errs
-	alreadyExistsStrings := []string{
-		"already exists",
-		"attributes mapping unavailble",
-	}
-	if err != nil {
-		isAlreadyExistsErr := false
-		for _, s := range alreadyExistsStrings {
-			if strings.Contains(err.Error(), s) {
-				isAlreadyExistsErr = true
-				break
-			}
-		}
-		if !isAlreadyExistsErr {
-			return fmt.Errorf("failed to register search attributes: %w", err)
-		}
+	// Throw an error if the attributes could not be registered, but ignore the
+	// case where they already exist (re-registration on a reused namespace).
+	if err != nil && status.Code(err) != codes.AlreadyExists {
+		return fmt.Errorf("failed to register search attributes: %w", err)
 	}
 	return nil
 }
--- a/loadgen/scenario.go
+++ b/loadgen/scenario.go
@@ -546,14 +546,13 @@ func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSi
 			err := executor.ExecuteClientSequence(cancelCtx, clientSeq)
 			if err != nil {
 				clientActionsErrPtr.Store(&err)
-				r.Logger.Error("Client actions failed: ", clientActionsErrPtr)
+				r.Logger.Error("Client actions failed: ", err)
 				cancel()
 
 				// TODO: Remove or change to "always terminate when exiting early" flag
-				err := r.Client.TerminateWorkflow(
-					ctx, options.StartOptions.ID, "", "client actions failed", nil)
-				if err != nil {
-					return
+				if terminateErr := r.Client.TerminateWorkflow(
+					ctx, options.StartOptions.ID, "", "client actions failed", nil); terminateErr != nil {
+					r.Logger.Errorf("Failed to terminate workflow after client actions failure: %v", terminateErr)
 				}
 			}
 		}()
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -376,10 +376,12 @@ func (e *ebbAndFlowExecutor) spawnWorkflowWithActivities(
 	// Wait for workflow completion
 	var result ebbandflow.WorkflowOutput
 	err = wf.Get(ctx, &result)
-	if err != nil {
-		e.Logger.Errorf("ebbAndFlowTrack workflow failed for iteration %d: %v", iteration, err)
-	}
+	// The batch has settled either way, so drain it from the backlog the
+	// controller tracks; only a successful workflow counts as completed.
 	e.completedActivities.Add(activities)
+	if err != nil {
+		return fmt.Errorf("ebbAndFlowTrack workflow failed for iteration %d: %w", iteration, err)
+	}
 	e.incrementTotalCompletedWorkflow()
 
 	return nil
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -230,7 +230,9 @@ func (e *ebbAndFlowExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo)
 			config.MinBacklog, config.MaxBacklog, config.Period, e.Configuration.Duration)
 	}
 
-	e.RegisterDefaultSearchAttributes(ctx)
+	if err := e.RegisterDefaultSearchAttributes(ctx); err != nil {
+		return fmt.Errorf("failed to register search attributes: %w", err)
+	}
 
 	var started, completed, backlog, target int64
 	lastIteration := time.Now()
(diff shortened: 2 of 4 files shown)


loadgen/scenario.go
```diff
@@ -306,22 +306,10 @@ func (s *ScenarioInfo) RegisterDefaultSearchAttributes(ctx context.Context) erro
...
-				break
-			}
-		}
-		if !isAlreadyExistsErr {
-			return fmt.Errorf("failed to register search attributes: %w", err)
-		}
+	// Throw an error if the attributes could not be registered, but ignore the
+	// case where they already exist (re-registration on a reused namespace).
+	if err != nil && status.Code(err) != codes.AlreadyExists {
+		return fmt.Errorf("failed to register search attributes: %w", err)
```
THardy98, line 410: I had assumed that there maybe were cases (i.e. old server versions) where we don't return `codes.AlreadyExists`, hence the string matching.

@stephanos I _think_ you wrote this? Are we ok to do away with the string matching now?

scenarios/ebb_and_flow.go
```diff
@@ -303,10 +305,12 @@ func (e *ebbAndFlowExecutor) spawnWorkflowWithActivities(
...
 	err = wf.Get(ctx, &result)
+	// The batch has settled either way, so drain it from the backlog the
+	// controller tracks; only a successful workflow counts as completed.
+	e.completedActivities.Add(activities)
 	if err != nil {
-		e.Logger.Errorf("ebbAndFlowTrack workflow failed for iteration %d: %v", iteration, err)
+		return fmt.Errorf("ebbAndFlowTrack workflow failed for iteration %d: %w", iteration, err)
 	}
-	e.completedActivities.Add(activities)
 	e.incrementTotalCompletedWorkflow()
```
THardy98, line 385: I'm not sure if this is desired.

I don't have the context here (I think @dnr wrote this scenario), but it could be intended to continue through failed workflows.

scenarios/ebb_and_flow.go
```diff
@@ -281,7 +281,9 @@ func (e *ebbAndFlowExecutor) spawnWorkflowWithActivities(
 
 	// Start workflow.
 	run := e.NewRun(int(iteration))
-	e.RegisterDefaultSearchAttributes(ctx)
+	if err := e.RegisterDefaultSearchAttributes(ctx); err != nil {
+		return fmt.Errorf("failed to register search attributes for iteration %d: %w", iteration, err)
+	}
```
THardy98: Likewise here, unsure if the preference would be to log or to stop the scenario with an error

loadgen/scenario.go
```diff
@@ -428,14 +416,13 @@ func (r *Run) ExecuteKitchenSinkWorkflow(ctx context.Context, options *KitchenSi
...
 				cancel()
 
 				// TODO: Remove or change to "always terminate when exiting early" flag
-				err := r.Client.TerminateWorkflow(
-					ctx, options.StartOptions.ID, "", "client actions failed", nil)
-				if err != nil {
-					return
+				if terminateErr := r.Client.TerminateWorkflow(
+					ctx, options.StartOptions.ID, "", "client actions failed", nil); terminateErr != nil {
+					r.Logger.Errorf("Failed to terminate workflow after client actions failure: %v", terminateErr)
```
THardy98, line 543: why do we log here instead of `return` ?

lilydoar: This block runs inside the go func() client-actions goroutine, so a bare return would just exit the goroutine and silently drop the terminate error — which is exactly what the old code did (if err != nil { return }). There's no error channel back to ExecuteKitchenSinkWorkflow.

The error we actually care about — the client-actions failure — does get propagated: it's stored in clientActionsErrPtr (line 418) and returned by the main path at line 435. The TerminateWorkflow call is best-effort cleanup after that failure has already happened, so its result is secondary

loadgen/scenario.go
lilydoar: One thing worth flagging regardless of the codes.AlreadyExists decision: the second string in the old list, "attributes mapping unavailble", is a typo — the real server message is ...unavailable. So strings.Contains never matched it, meaning only the "already exists" branch was ever actually doing anything. Removing that second string drops no working behavior.

scenarios/ebb_and_flow.go
lilydoar: It doesn't stop the scenario — returning the error feeds it into errCh, which the Run loop tolerates via MaxConsecutiveErrors (default 10): a failure increments the counter, any success resets it, and we only abort after that many consecutive failures. So we do continue through the occasional failed workflow, just not silently.

The stronger signal that failures shouldn't be swallowed is VerifyNoFailedWorkflows at the end of Run — the scenario already fails if any workflow ended up FAILED/TERMINATED, so the old log-and-continue was inconsistent with that check. It also counted failed runs as completed (incrementTotalCompletedWorkflow), which inflated TotalCompletedWorkflows (used by the == 0 guard and MinVisibilityCountEventually). Note I still drain completedActivities on failure so the backlog-oscillation math stays correct — only the completed count is gated on success now.

scenarios/ebb_and_flow.go
lilydoar: Same mechanism as above — this returns into errCh and is tolerated by MaxConsecutiveErrors, so a transient blip won't stop the scenario. RegisterDefaultSearchAttributes already swallows the benign AlreadyExists case internally, so it only returns on a genuinely unexpected failure.

Failing (loudly) is the right call here because the whole verification path depends on it: the workflows tag themselves with OmesExecutionIDSearchAttribute and both post-scenario checks (MinVisibilityCountEventually, VerifyNoFailedWorkflows) query by it. If registration silently fails, that surfaces later as a confusing visibility error instead of pointing at the real cause.

PR #419, opened 2026-07-14
loadgen: add configurable weighted payload-size distribution to throughput_stress

tdeebswihart:
## What was changed

Adds a `--option payload-distribution-json` knob to the `throughput_stress` scenario to pull payload sizes from a `DistributionField`. This replaces the existing hard-coded 256B payload size which I've kept as the default.

NOTE: Activity input bytes are now pseudo-random data pulled from the rng so we can actually test CDS/Walker behavior with large payloads as we compress everything we can.

## Why?

We need to test temporal with more realistic workloads. This will let us configure those

### DIFF
--- a/loadgen/helper_payload.go
+++ b/loadgen/helper_payload.go
@@ -1,0 +1,51 @@
+package loadgen
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"math/rand"
+)
+
+// PayloadConfig configures activity payload sizes using the shared distribution
+// framework. Parsed from a JSON string passed via --option (with @file.json support).
+type PayloadConfig struct {
+	// Size is the distribution (in bytes) of activity payload sizes. See DistributionField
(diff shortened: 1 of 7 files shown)


loadgen/helper_payload.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"math/rand"
+)
+
+// PayloadConfig configures activity payload sizes using the shared distribution
+// framework. Parsed from a JSON string passed via --option (with @file.json support).
+type PayloadConfig struct {
```
tdeebswihart, line 11: This _is_ a one-field struct so could alternately just be a `PayloadSize *DistributionField[int32]` on the main workload struct

THardy98: imo, don't mind having a `PayloadConfig`

THardy98 (approved):

PR #434, opened 2026-07-29
docs: add scenario-authoring guide

lilydoar:
## What

Adds `docs/authoring-scenarios.md` — the authoring counterpart to `docs/running.md`. Scenario authoring was previously only a short README section, which covered the mechanics but not the parts that surprise people.

Covers:
- Registration from `init()`, and that **the file name is the scenario name** — so renaming a file renames the scenario and moves its task queue.
- Choosing between `KitchenSinkExecutor` and a custom `Executor`, and what you take on when you cross that line (your own option parsing/validation, and your own loop unless you delegate to `GenericExecutor`).
- The two configuration channels — framework run flags vs `--option` — and conventions for documenting and validating options.
- The current sharp edges, stated plainly because they change how carefully you document: the typed option accessors **panic** on a malformed value, and an **unknown/misspelled option key is silently ignored**.

Also slims the README's authoring section to a short summary pointing at the guide, and adds it to the README documentation index.

## Notes

Docs only. Run commands and the run-flag list stay in `docs/running.md` and are referenced rather than repeated, so there's one home per fact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ benchmark features and situations. Backwards compatibility may not be maintained
 
 - **[Running & configuring omes](./docs/running.md)** — run commands, load configuration, worker
   profiles, and running against a specific SDK version.
+- **[Authoring scenarios](./docs/authoring-scenarios.md)** — writing a new scenario, choosing an
+  executor, exposing options, and authoring conventions.
 
 ## Why the weird name?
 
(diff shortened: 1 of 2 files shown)


THardy98 (commented):
I'd appreciate adding a section that includes using the `project/app` scenario. It's effectively a way to write simple scenarios, where you effectively just want to run a workload at a configurable level of concurrency.

See the `helloworld` examples in the the worker directories for examples.

I think this is important to include as it's a much lower barrier to entry - you define in Temporal semantics (i.e. you write your workflow/nexus/activities as you normally would) without needing to know Omes at all. It's effectively like writing a sample that you want to run concurrently

lilydoar:
Good call — added in ff5fe0d. You're right that leading with the framework path sends newcomers to the harder of the two, so the project path now comes first.

The guide opens with a table choosing between them, framed on your point: a project is plain Temporal code (you write the workflow/activities/Nexus ops as you normally would) and omes just runs your `Execute` handler repeatedly at a configured concurrency — no omes internals required. A scenario is for when you need a load *shape* the steady-rate executor can't express.

The new **Writing a project** section covers:
- where apps live (`workers/<lang>/apps/<name>`) and registration in `registry.*`, noting `helloworld` exists in Go, Python, TypeScript, .NET, and Ruby as the thing to copy
- the `harness.App` / `ProjectHandlers` surface, using the Go `helloworld` app as the example, plus the optional `Init` handler
- giving each iteration a unique workflow ID (`ExecutionID` + `Iteration`) so concurrent iterations don't collide — easy to miss when adapting a sample that ran once
- a minimal try-it invocation, with the README's Project section referenced for the rest (worker/scenario separately, Docker) rather than duplicated
- the steady-rate limitation stated plainly, since that's the actual reason to graduate to a scenario

One thing I'd like your read on: I described the boundary as "steady rate — run *x* iterations or for *y* duration," taken from the README's current-limitations note. If projects have since grown past that, or if there's a nuance worth calling out, I'll correct it.

THardy98 (approved):

PR #435, opened 2026-07-29
loadgen: make option declarations the source of truth

lilydoar:
> **Stacked on #434** — review/merge that first. This PR's base is `lilydoar/omes-docs-authoring`, so the diff here is only this change. GitHub retargets the base to `main` when #434 merges.

Scenario options were an untyped map. A scenario read `--option` values lazily with an inline default, so nothing knew which options existed:

- A malformed value **panicked** from inside the run (`--option children-per-workflow=abc` → stack trace).
- An unknown or misspelled name was **silently ignored** — the run proceeded at the default, quietly producing a differently-shaped load than asked for.
- `list-scenarios` had nothing to show, so options were undiscoverable without reading source. Several scenarios worked around that by listing options in prose in their `Description`, which then drifted from the code.
- A default lived in **two** places — the declaration and the inline argument at every read site — with nothing keeping them in sync.

Three commits, each reviewable on its own.

## 1. Declare options with pflag — `d5a52c6`

```go
Options: func(o *loadgen.OptionSet) {
    o.Int("children-per-workflow", 30, "Child workflows per iteration.")
    o.Int("task-queue-count", 0, "Task queues to spread iterations across.")
    o.MarkRequired("task-queue-count")
},
```

`OptionSet` embeds a `pflag.FlagSet` — already a dependency for omes's own CLI flags — so every typed registrar, the type checking, and default rendering come from pflag rather than a bespoke type and parser. `loadgen` adds only `MarkRequired`, and phrases errors in terms of `--option name=value` instead of pflag's `--flag`.

```
$ omes run-scenario --scenario throughput_stress --run-id t --option sleep-tim=5s
invalid --option values for scenario "throughput_stress":
unknown option "sleep-tim"; this scenario accepts: continue-as-new-after-iterations, include-describe, ...

$ omes run-scenario --scenario project --run-id t
invalid --option values for scenario "project":
option "language" is required
option "project-name" is required
```

- Unknown names rejected, listing what the scenario does accept.
- Malformed values rejected **before the client dial**, with every problem reported at once instead of one per attempt.
- `list-scenarios` shows each option's type, default, requiredness, and usage.
- A rejected option prints as a plain usage error via a typed `InvalidOptionsError`, not a fatal-level stack trace. Genuine run failures keep the existing fatal path.
- `Scenario.DefaultConfiguration` supersedes the `HasDefaultConfiguration` executor interface, which **no in-repo scenario implemented** — which is why `list-scenarios` never printed default configuration for any of them. The interface still works.

All 13 option-using scenarios are migrated, and the prose option lists in their `Description`s are removed now that `list-scenarios` renders the real thing.

## 2. Make the declarations honest — `9846967`

Declaring surfaced four options whose declared default could not describe what the code did. Each is fixed at the source rather than documented around.

**`internal-iterations-timeout` has a computed default** (run duration + 1 minute), which no static default can express. `0` stays the sentinel and the usage string now says what `0` means.

**`nexus-endpoint` meant two opposite things** — in `throughput_stress` an empty value disabled Nexus; in `fuzzer` it meant "create an endpoint for this run". The enabled/disabled concept is now its own flag:

- `throughput_stress` gains `nexus-enabled` (default `false`, preserving Nexus being off unless asked for).
- `nexus-endpoint` means the same thing everywhere: empty creates one for this run.
- `include-standalone-nexus` requires `nexus-enabled` rather than a non-empty endpoint — what it always meant.
- Supplying an endpoint while `nexus-enabled` is false is an error instead of being silently ignored.

Nexus stays structural in `fuzzer`, so it gets no `nexus-enabled` flag; a switch that cannot be turned off is worse than none. Creating the endpoint needs a context, so `throughput_stress` resolves it in `Run` rather than `Configure`, via a new `ScenarioInfo.EnsureNexusEndpoint`.

**`phase-time` is deleted** — a deprecated alias whose value became `period`'s default, the one case of an option feeding another option's default. Both defaults were already `60s`, so nobody relying on the default is affected, and anyone passing it explicitly now gets an error naming `period` instead of silence.

**`no-output-file` becomes `output-file`** — a path rather than a negated boolean. It was previously checked by key *presence*, so `no-output-file=false` skipped the file anyway. Default is the same file, empty writes nothing, and the path is now configurable.

## 3. Read from the declarations, drop the map — `b7e2740`

Declarations are now the only home for a default. `ScenarioInfo` carries the resolved `OptionSet` instead of a `map[string]string`, and the accessors take no default:

```go
children := info.OptionInt("children-per-workflow")
```

Values arrive from pflag already typed, so scenarios no longer parse strings themselves. Two sites that had been hand-parsing are plain reads now: `throughput_stress`'s `visibility-count-timeout` (was `ParseDuration` over a `cmp.Or`-defaulted string, with its own error branch) and `project`'s `project-server-ready-timeout`.

Where a default genuinely cannot be a constant, the sentinel is resolved explicitly in code rather than hidden in an accessor argument.

Tests build their options with `loadgen.MustResolveScenarioOptions`, which resolves through the scenario's real declarations — so a test that misspells an option now fails instead of quietly exercising a default. That caught a `throughput_stress` fixture setting `nexus-endpoint` without `nexus-enabled`.

## Breaking changes

- **Declaration is mandatory.** A scenario accepts exactly what it declares, so passing an option to a scenario that declares none is an error naming the scenario. Scenarios outside this repository must declare their options to read them — they fail to *compile* rather than silently reading zero values.
- **`ScenarioInfo.ScenarioOptions` and the `ScenarioOption*` accessors are gone**, replaced by `ScenarioInfo.Options` and `Option*`.
- **Renamed/removed options:** `no-output-file` → `output-file`; `phase-time` deleted; `nexus-endpoint` on `throughput_stress` now requires `nexus-enabled`.

## Downstream

`saas-cicd` vendors omes as a library and builds `ScenarioInfo` itself, so it needs three matching adjustments — the resolved-options call, `nexus-enabled` alongside the endpoint it already injects, and `output-file` instead of `no-output-file`. Those are written on `lilydoar/omes-option-api-bump` and must land together with the omes pin bump, since they don't compile against the currently pinned omes.

## Verification

`go test ./loadgen ./scenarios/...` passes; build and `gofmt` clean. End-to-end checks against a real CLI and an embedded server: typo rejection, type errors, required options, the `nexus-enabled` guard, the deleted `phase-time` alias, a scenario declaring no options rejecting one, and a run with no options completing on declared defaults.

`TestKitchenSink` is excluded above because it fails identically on unmodified `main` in my environment — it builds a real dev server plus language-SDK workers. CI is the gate for it.

### DIFF
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -76,11 +76,20 @@ var _ loadgen.Resumable = (*ebbAndFlowExecutor)(nil)
 
 func init() {
 	loadgen.MustRegisterScenario(loadgen.Scenario{
-		Description: "Oscillates backlog between min and max.\n" +
-			"Options:\n" +
-			"  min-backlog, max-backlog, period, sleep-duration, max-rate,\n" +
-			"  control-interval, max-consecutive-errors, backlog-log-interval.\n" +
-			"Duration must be set.",
+		Description: "Oscillates backlog between min and max. Duration must be set.",
+		Options: func(o *loadgen.OptionSet) {
+			o.Int(MinBacklogFlag, 0, "Minimum total backlog to target.")
+			o.Int(MaxBacklogFlag, 30, "Maximum total backlog to target.")
+			o.Duration(PeriodFlag, 60*time.Second, "Period of backlog-size oscillation.")
+			o.Duration("phase-time", 60*time.Second, "Deprecated alias for "+PeriodFlag+".")
+			o.Duration(SleepDurationFlag, time.Millisecond, "How long each activity sleeps.")
+			o.Int(MaxRateFlag, 1000, "Maximum workflows spawned per control interval.")
+			o.Duration(ControlIntervalFlag, 100*time.Millisecond, "How often the backlog is controlled.")
+			o.Int(MaxConsecutiveErrorsFlag, 10, "Consecutive errors tolerated before stopping.")
+			o.Duration(BacklogLogIntervalFlag, 30*time.Second, "How often backlog stats are logged.")
+			o.Duration(VisibilityVerificationTimeoutFlag, 30*time.Second, "Timeout for the visibility count check.")
+			o.String(SleepActivityJsonFlag, "", "JSON sleep-activity configuration; use @<file> to read from a file.")
+		},
 		ExecutorFn: func() loadgen.Executor { return newEbbAndFlowExecutor() },
 	})
 }
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -81,7 +81,6 @@ func init() {
 			o.Int(MinBacklogFlag, 0, "Minimum total backlog to target.")
 			o.Int(MaxBacklogFlag, 30, "Maximum total backlog to target.")
 			o.Duration(PeriodFlag, 60*time.Second, "Period of backlog-size oscillation.")
-			o.Duration("phase-time", 60*time.Second, "Deprecated alias for "+PeriodFlag+".")
 			o.Duration(SleepDurationFlag, time.Millisecond, "How long each activity sleeps.")
 			o.Int(MaxRateFlag, 1000, "Maximum workflows spawned per control interval.")
 			o.Duration(ControlIntervalFlag, 100*time.Millisecond, "How often the backlog is controlled.")
@@ -118,9 +117,7 @@ func (e *ebbAndFlowExecutor) Configure(info loadgen.ScenarioInfo) error {
 		return fmt.Errorf("max-backlog must be greater than min-backlog, got max=%d min=%d", config.MaxBacklog, config.MinBacklog)
 	}
 
-	// TODO: backwards-compatibility, remove later
-	pt := info.ScenarioOptionDuration("phase-time", 60*time.Second)
-	config.Period = info.ScenarioOptionDuration(PeriodFlag, pt)
+	config.Period = info.ScenarioOptionDuration(PeriodFlag, 60*time.Second)
 	if config.Period <= 0 {
 		return fmt.Errorf("period must be greater than 0, got %v", config.Period)
 	}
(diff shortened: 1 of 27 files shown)


scenarios/ebb_and_flow.go
```diff
@@ -81,7 +81,6 @@ func init() {
 			o.Int(MinBacklogFlag, 0, "Minimum total backlog to target.")
 			o.Int(MaxBacklogFlag, 30, "Maximum total backlog to target.")
 			o.Duration(PeriodFlag, 60*time.Second, "Period of backlog-size oscillation.")
-			o.Duration("phase-time", 60*time.Second, "Deprecated alias for "+PeriodFlag+".")
```
THardy98, line 84: Marked deprecated, but worth verifying if this is safe

scenarios/ebb_and_flow.go
```diff
@@ -118,9 +117,7 @@ func (e *ebbAndFlowExecutor) Configure(info loadgen.ScenarioInfo) error {
 		return fmt.Errorf("max-backlog must be greater than min-backlog, got max=%d min=%d", config.MaxBacklog, config.MinBacklog)
 	}
 
-	// TODO: backwards-compatibility, remove later
-	pt := info.ScenarioOptionDuration("phase-time", 60*time.Second)
-	config.Period = info.ScenarioOptionDuration(PeriodFlag, pt)
+	config.Period = info.ScenarioOptionDuration(PeriodFlag, 60*time.Second)
```
THardy98, line 120: ditto

THardy98 (commented):
Looks good on the whole, nice improvement.

The addition of unit tests are nice, but it may be worth thinking of a way to validate that the options change preserves existing scenario behaviour (i.e. we haven't introduced a regression by changing the options process) through an integration/E2E test.

Something like a dry-run of our scenarios with options supplied would be nice.

I see `Options: loadgen.MustResolveScenarioOptions` diff in the existing scenario tests but am unsure if that's enough.

THardy98 (approved):
Am approving given that CI passes as is, but would keep in mind the idea of adding something akin to a dry-run in the future

PR #436, opened 2026-08-01
Rate-based control in ebb-and-flow scenario

carlydf:
## What was changed
Revives @dnr PR #297 which is stale.

Add rate-based control to ebb-and-flow scenario.

## Why?
Test different task creation rates.

## Checklist

2. How was this tested:
unit test + manual test with ongoing partition autoscaling work

### DIFF
--- a/scenarios/ebb_and_flow.go
+++ b/scenarios/ebb_and_flow.go
@@ -112,7 +112,7 @@ func init() {
 			o.Float64(MinAddRateFlag, 0, "Minimum activities added per second; enables rate mode with max-add-rate.")
 			o.Float64(MaxAddRateFlag, 0, "Maximum activities added per second; enables rate mode with min-add-rate.")
 			o.Duration(RatePeriodFlag, 10*time.Minute, "Period of add-rate oscillation.")
-			o.Duration(BacklogPeriodFlag, 0, "Period of backlog-size oscillation in rate mode; defaults to period.")
+			o.Duration(BacklogPeriodFlag, 0, "Period of backlog-size oscillation; defaults to period.")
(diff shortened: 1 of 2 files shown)


scenarios/ebb_and_flow.go
```diff
@@ -88,6 +109,11 @@ func init() {
 			o.Duration(BacklogLogIntervalFlag, 30*time.Second, "How often backlog stats are logged.")
 			o.Duration(VisibilityVerificationTimeoutFlag, 30*time.Second, "Timeout for the visibility count check.")
 			o.String(SleepActivityJsonFlag, "", "JSON sleep-activity configuration; use @<file> to read from a file.")
+			o.Float64(MinAddRateFlag, 0, "Minimum activities added per second; enables rate mode with max-add-rate.")
+			o.Float64(MaxAddRateFlag, 0, "Maximum activities added per second; enables rate mode with min-add-rate.")
+			o.Duration(RatePeriodFlag, 10*time.Minute, "Period of add-rate oscillation.")
+			o.Duration(BacklogPeriodFlag, 0, "Period of backlog-size oscillation in rate mode; defaults to period.")
```
dnr: not just in rate mode.. the point is I want to rename the period flag to backlog-period, so we don't have period for backlog mode and rate-period for rate mode. but we can't rename it in one step since oss-cicd uses it (right?)

dnr (approved):

scenarios/ebb_and_flow.go
carlydf: yeah, I imagine we'd have to introduce the new flag, make OSS CICD use it, then remove / deprecate the old flag

carlydf: changed the comment to not say "in rate mode"

PR #439, opened 2026-08-04
loadgen: gate feature options on namespace capabilities

tdeebswihart:
## What was changed

I've added a new kind of option for Omes: _features_. `o.Feature(name, usage, func(caps *namespacev1.NamespaceInfo_Capabilities) bool)` declares a
boolean scenario option that is automatically enabled if the corresponding feature is enabled for the testing namespace. 

`include-standalone-nexus` and `include-standalone-activity` in throughput_stress have been ported to use it; this way those features are automatically exercised if the namespace we're testing supports it.

This also fixes a Java worker bug it uncovered: `ClientActionExecutor.executeStandaloneActivity` used the result-less `ActivityClient.execute` overload, which asks the SDK to deserialize the result into `java.lang.Void`. The raw payload converter rejects that for any activity returning bytes, and the enclosing client-action activity retries without bound, so the iteration hung to the run timeout instead of failing. It now requests `byte[]`.

## Why?

Our feature release process is a bit of a mess right now, and one of the manual steps is "Remember to update oss-cicd as well as VCR to exercise your feature but only after its actually supported".

That's a lot and easy to get wrong, as the feature would need to be enabled _everywhere_ that the corresponding version of omes runs. Which means:
- Every test cell
- Release validation, but only when that release is being validated
- etc

Automatically enabling load exercise when the namespace reports support is a simple solution to this.

## How was this tested
- unit tests for the full option resolution matrix (unset/explicit crossed with supported/unsupported, joined errors for multiple unsupported features, `list-scenarios` rendering of the dynamic default)
- The existing throughput_stress integration test now exercises the probe against the dev server.
- a new integration test that checks this functionality

### DIFF
--- a/docs/authoring-scenarios.md
+++ b/docs/authoring-scenarios.md
@@ -163,6 +163,10 @@ loadgen.MustRegisterScenario(loadgen.Scenario{
 
 		o.Int("task-queue-count", 0, "Number of task queues to spread iterations across.")
 		o.MarkRequired("task-queue-count")
+
+		// Gated on a namespace capability; see "Feature options" below.
+		o.Feature("include-standalone-activity", "Include standalone activities.",
+			func(c *namespacev1.NamespaceInfo_Capabilities) bool { return c.GetStandaloneActivities() })
 	},
 	ExecutorFn: /* ... */,
 })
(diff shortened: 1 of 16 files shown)


docs/authoring-scenarios.md
```diff
@@ -212,6 +216,33 @@ too:
...
+| `=true` | yes | enabled |
+| `=true` | no | run fails with a usage error |
+| unset | probe failed after retries | run fails |
+
+Resolution needs a dialed client, so it happens in `loadgen.ResolveFeatureOptions` after the run
+connects, not at declaration or `ResolveOptions` time. A test that calls `Configure` directly
+without also calling `ResolveFeatureOptions` sees the feature at its pflag default (`false`), not
+its capability-resolved value.
+
+Reach for `o.Feature` only when a matching field exists on `NamespaceInfo_Capabilities` (from
```
lilydoar, line 243: Are these Namespace_Capabilities a piece of the OSS project? Not cloud side?

tdeebswihart: They're an OSS API but also available in the cloud

THardy98 (approved):

PR #442, opened 2026-08-05
Report a missed capability probe instead of reading false

lilydoar:
Stacked PRs:
 * #444
 * #443
 * __->__#442
 * #441

--- --- ---

### Report a missed capability probe instead of reading false

Feature options are resolved by ResolveFeatureOptions, which only the omes CLI
calls. Code driving a scenario as a library — VCR and oss-cicd build a
ScenarioInfo and call executor.Run directly — never resolves them, so every
feature option reads false and the run omits the feature it was meant to
exercise, indistinguishable from a run that worked.

The OptionSet now tracks resolution, and reading an unresolved feature reports
the option and the call that fixes it. An explicitly supplied value is
authoritative without a probe, so only an untouched option is reported, and a
failed resolution leaves the set unresolved. OptionSet.Set records a value as
deliberately chosen, which is what lets a library caller or a test set an option
directly and have it treated like a user passing --option.

Also:

- ResolveFeatureOptions no longer panics on a nil logger, which library callers
  are the ones to pass.
- A created Nexus endpoint is logged distinctly from a reused one, so an operator
  can see the run added something to the namespace. FuzzExecutor goes through the
  same path instead of duplicating it.
- Declaring an option both required and capability-gated is rejected: required
  means the user must supply it, gated means the namespace does.

### DIFF
--- a/docs/authoring-scenarios.md
+++ b/docs/authoring-scenarios.md
@@ -248,6 +248,15 @@ Resolution needs a dialed client, so it happens in `loadgen.ResolveFeatureOption
 connects, not at declaration or `ResolveOptions` time. A test that calls `Configure` directly
 without also calling `ResolveFeatureOptions` sees the feature at its pflag default (`false`), not
 its capability-resolved value.
+**If you drive a scenario as a library rather than through the omes CLI, you must call
+`loadgen.ResolveFeatureOptions` yourself** once the client is dialed. The CLI does it for you; code
+that builds a `ScenarioInfo` and calls `executor.Run` directly does not get it for free. Skipping it
+does not fail the run — every feature option reads `false`, so the load quietly omits the feature —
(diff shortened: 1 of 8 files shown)


docs/authoring-scenarios.md
```diff
@@ -239,6 +239,15 @@ Resolution needs a dialed client, so it happens in `loadgen.ResolveFeatureOption
 connects, not at declaration or `ResolveOptions` time. A test that calls `Configure` directly
 without also calling `ResolveFeatureOptions` sees the feature at its pflag default (`false`), not
 its capability-resolved value.
+**If you drive a scenario as a library rather than through the omes CLI, you must call
+`loadgen.ResolveFeatureOptions` yourself** once the client is dialed. The CLI does it for you; code
```
tdeebswihart, line 252: 👀

tdeebswihart (approved):

PR #444, opened 2026-08-05
Share the Nexus options and gate them on reported capability

lilydoar:
nexus-enabled and nexus-endpoint were declared per scenario and had diverged:
throughput_stress declared both, FuzzExecutor declared only the endpoint with a
different description, and fuzzer re-derived the endpoint name FuzzExecutor
derives in order to create it. Each scenario also implemented "empty endpoint
means create one" and its own relationship between a Nexus sub-feature and Nexus
itself.

DeclareNexusOptions declares them once. ResolveNexusConfig settles whether the
run generates Nexus load and against which endpoint, creating one if needed;
IncludeNexusSubFeature holds the sub-feature rule; NexusEndpointName gives the
creation site and the generator one derivation of the name.

Nexus is on by default: omes is a testing tool, so the useful default is to
exercise what the server supports, and nexus-enabled=false turns it off. Not a
bare true — it is gated on the server reporting Nexus, so it turns itself off
against a server without it, and asking for it explicitly there fails.

Gating it widened the feature mechanism. Predicates read namespace capabilities
from DescribeNamespace, while Nexus support comes from GetSystemInfo; they now
receive a Capabilities carrying both, leaving one implementation of the rule
instead of two.

Nexus being off takes standalone Nexus with it, which throughput_stress applies
in Run as well as Configure: whether the server supports Nexus is only known once
the client is dialed.

The per-language smoke tests pin nexus-enabled=false. They check that a built
worker can run the scenario at all, so the load shape should not follow what the
server happens to support; Nexus operations additionally need the worker to
handle them, which the java, dotnet, and typescript workers do not do within the
smoke test's budget. Nexus stays covered by the Go tests, which run it against a
dev server.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,10 @@ jobs:
             --run-id ${{ github.run_id }} \
             --connect-timeout 1m \
             --iterations 5
+      # This checks that a built worker can run the scenario at all, so the load
+      # shape is pinned rather than following what the server supports. Nexus is
+      # on by default and needs the worker to handle Nexus operations, which not
+      # every language does; Nexus itself is covered by the Go tests.
       - name: Smoke test throughput_stress to confirm build is usable against ${{ matrix.sdk }} worker
         timeout-minutes: 10
         run: |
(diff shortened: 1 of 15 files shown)


tdeebswihart (approved):
I like this, thank you!

PR #445, opened 2026-08-07
Concurrent writes in viz stress test

alexpaschenko:
Enables concurrent writes in viz stress test - otherwise max effective wfRPS is just that of the single thread (about 60-ish in practice).

Testing: built, pushed, and ran omes-deploy with this new image, tested with up to 5K wfRPS.

Dockerfile change is orthogonal to concurrency but necessary for the builds to pass.


alexpaschenko:
Deepak will merge `dk/vis-stress` into `main` first.

PR #446, opened 2026-08-10
dev: sync SDK dependencies with mise

THardy98:
## What changed?

This adds per-SDK and aggregate `mise` tasks that synchronize the repository's SDK dependency declarations and lockfiles with the versions pinned in `mise.toml`. The tasks delegate updates to each ecosystem's package manager where possible, with small fail-fast helpers for the Ruby and Java declarations that their package managers cannot update directly.

The TypeScript and Python compatibility-range lower bounds now advance with their tested SDK versions.
Java uses one shared Gradle property for its SDK and testing dependencies, while Ruby retains its published compatibility range and an exact development pin. 

CI runs the aggregate synchronization task and fails when it leaves a repository diff, and the README documents the workflow.

## Why?

`mise.toml` already defines the SDK versions used by CI and worker-image builds, but changing those pins did not update files such as `package.json`, `pyproject.toml`, `go.mod`, or their lockfiles. Keeping this workflow in `mise` makes the version pins actionable across the repository without expanding the Go development CLI.

After this change, an SDK bump consists of updating the relevant `mise` pin, running `mise run sync-sdk`, and reviewing the package-manager changes. There is no runtime rollout or manual migration.

## How did you test it?

- [ ] built
- [x] run locally and tested manually
- [x] covered by existing tests
- [ ] added new unit test(s)
- [ ] added new functional test(s)

Validation included `mise tasks validate`, two consecutive aggregate sync runs with identical diff hashes, a complete declared/resolved-version audit, fail-fast and idempotence checks for the Ruby and Java helpers, Gradle configuration, and the affected existing worker test suites plus focused Go tests.

## Potential risks

Package-manager resolution can update transitive lockfile entries alongside the requested SDK. The CI drift check also depends on the configured package registries being available. The helper scripts intentionally fail if the Ruby or Java declaration shape changes, requiring the synchronization logic to be updated alongside such a refactor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,9 +336,9 @@ mise run sync-sdk:typescript # Sync one SDK
 mise run sync-sdk            # Sync every SDK
 ```
 
-The sync tasks preserve declared compatibility ranges while updating exact dependency declarations and
-lockfiles to the versions in `mise.toml`. Review and commit the generated changes; CI reruns the aggregate
-task and fails if the committed dependency files have drifted.
+The sync tasks use each ecosystem's package manager to update declared SDK requirements and lockfiles to
+the versions in `mise.toml`. Review and commit the generated changes; CI reruns the aggregate task and
+fails if the committed dependency files have drifted.
 
 ## Fuzzer trophy case
 
--- a/README.md
+++ b/README.md
@@ -329,6 +329,17 @@ Or target specific languages: `go run ./cmd/dev build go java python`
 
 All development tool, SDK, and server versions are defined in `mise.toml`.
 
+After changing an SDK version in `[_.sdk]`, synchronize the corresponding worker dependency files:
+
+```sh
+mise run sync-sdk:typescript # Sync one SDK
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -329,6 +329,17 @@ Or target specific languages: `go run ./cmd/dev build go java python`
 
 All development tool, SDK, and server versions are defined in `mise.toml`.
 
+After changing an SDK version in `[_.sdk]`, synchronize the corresponding worker dependency files:
```
lilydoar, line 332: Does it make sense to use mise's sources/outputs args here?

https://mise.jdx.dev/tasks/architecture.html#source-and-output-tracking

This way, if I run `sync-sdk` task, it will run/not run each subtask based on which files have changed.

lilydoar (approved):
lgtm. Left one comment about a potential enhancement.

README.md
THardy98: will leave this for now, `sync-sdk` tasks are quite quick in any case

PR #451, opened 2026-08-18
Add standalone activity operator command load

fretz12:
## What was changed
Added capability-gated standalone activity operator-command load to throughput_stress, exercising pause/unpause, reset, and update-options in rotation. The Go worker temporarily uses direct RPCs. Unsupported SDK workers skip the action unless strict mode is enabled. Tests and documentation were added. Also added check for existing endpoint to allow serviceerror.AlreadyExists, since client converts the raw response

## Why?
To load-test server support before operator-command APIs are available across all SDKs, while avoiding failures against unsupported servers.


loadgen/kitchensink/operator_commands_test.go
```diff
@@ -0,0 +1,356 @@
...
+						StartToCloseTimeout: durationpb.New(15 * time.Second),
+						HeartbeatTimeout:    durationpb.New(3 * time.Second),
+					},
+				})
+
+			require.NoError(t, err)
+			require.Equal(t, "worker-task-queue", testClient.startedTaskQueue)
+			require.Equal(t, tt.wantCalls, service.calls)
+			require.Len(t, service.requestIDs, tt.wantRequestIDs)
+			require.Len(t, uniqueStrings(service.requestIDs), tt.wantRequestIDs)
```
THardy98: nit: i think we want `tt` as the 2nd arg, not the 3rd

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -568,10 +568,41 @@ func Heartbeat(ctx context.Context, config *kitchensink.ExecuteActivityAction_He
...
+		activity.RecordHeartbeat(ctx)
+		return nil
+	}
+	return runWithHeartbeats(ctx, config.SuccessDuration.AsDuration(), interval,
+		func() { activity.RecordHeartbeat(ctx) })
+}
+
+// runWithHeartbeats records immediately and periodically until duration elapses.
+// It preserves the context's cancellation cause so callers retain the original reason.
+func runWithHeartbeats(
```
THardy98, line 583: nit: named and docstring mention heartbeats but the function itself is agnostic to heartbeating

More of a comment, but either we can rename the function or make it explicitly expect a heartbeat fn arg.

I don't have a strong opinion on changing this necessarily, moreso mentioning it

loadgen/kitchensink/operator_commands_test.go
```diff
@@ -0,0 +1,356 @@
...
+	activitypb "go.temporal.io/api/activity/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	workflowservicepb "go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func TestStandaloneActivityOperatorCommandTypesStartOnWorker(t *testing.T) {
```
THardy98: A bit misleading, I don't think this actually starts or runs anything on a worker. IIUC, we're asserting an expected command sequence from the client

loadgen/kitchensink/operator_commands_test.go
THardy98, line 1: We're doing a lot of mocking here on the `operatorCommandTestService`. Do you have any concerns about validating these cases against actual server? (i.e. like in `kitchen_sink_executor_test.go`)

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1063,6 +1072,41 @@ func TestKitchenSink(t *testing.T) {
...
+		{
+			name: "ExecActivity/Client/StandaloneActivityOperatorCommands",
+			testInput: &TestInput{
+				WorkflowInput: &WorkflowInput{
+					InitialActions: ListActionSet(
+						ClientActivity(
+							ClientActions(&ClientAction{
+								Variant: &ClientAction_DoStandaloneActivityOperatorCommands{
+									DoStandaloneActivityOperatorCommands: &DoStandaloneActivityOperatorCommands{
+										CommandType: DoStandaloneActivityOperatorCommands_COMMAND_TYPE_UPDATE,
```
THardy98: Do we not want to test against `PAUSE` / `RESET` as well?

Might be worth adding some cases for these operations as well, since this is where our integration testing lies

scenarios/throughput_stress.go
```diff
@@ -554,6 +566,17 @@ func (t *tpsExecutor) createActionsChunk(
 		if t.config.IncludeStandaloneActivity {
 			asyncActions = append(asyncActions, t.createStandaloneActivityAction(loadgen.TaskQueueForRun(run.RunID), rng))
 		}
+		if t.config.IncludeStandaloneActivityOperatorCommands {
+			// Keep command rotation continuous across outer iterations and Continue-As-New.
+			internalIteration := t.config.InternalIterations - remainingInternalIters + i
+			commandOrdinal := (run.Iteration-1)*t.config.InternalIterations + internalIteration
```
THardy98: Not the easiest bit of code to read.

Codex is suggesting something like:
```
  completedInPriorRuns := (run.Iteration - 1) * t.config.InternalIterations
  completedInCurrentRun := t.config.InternalIterations - remainingInternalIters
  commandOrdinal := completedInPriorRuns + completedInCurrentRun + i
```
which I do agree is a bit easier to read.

Or a helper function

scenarios/throughput_stress_test.go
```diff
@@ -299,6 +304,107 @@ func TestThroughputStressPayloadSequenceAcrossContinueAsNew(t *testing.T) {
...
+		ExecutionID: "exec",
+		Logger:      zap.NewNop().Sugar(),
+		Options: loadgen.MustResolveScenarioOptions("throughput_stress", map[string]string{
+			IncludeStandaloneActivityOperatorCommandsFlag: "true",
+			NexusEnabledFlag: "false",
+		}),
+	}
+	require.NoError(t, executor.Configure(info))
+
+	sets := executor.createActions(info.NewRun(1))
```
THardy98: You might want to test against `NewRun(<number greater than one>)` so you test that we're calculating the command ordinal correctly across (outer) runs.

workers/go/workerlib/kitchensink/kitchen_sink.go
THardy98, line 1: A bit of a pain - but could you add this diff to the other workers as well? `heartbeat_interval` is configurable from the proto without needing SAA operator support and I'd like to avoid drift between the workers.

THardy98 (changes requested):

loadgen/kitchen_sink_executor_test.go
fretz12: agreed, added

loadgen/kitchensink/operator_commands_test.go
fretz12: Yea on second thought, the e2e test provides the real value, and we already have extensive server side coverage. The operator load itself isn't super complicated, so I think I'm good with removing it to relieve maintenance burden.

loadgen/kitchensink/operator_commands_test.go
fretz12: file removed

loadgen/kitchensink/operator_commands_test.go
fretz12: file removed

scenarios/throughput_stress.go
fretz12: refactored to a helper. It'll be reused with a follow on batch operator load

scenarios/throughput_stress_test.go
fretz12: good point, changed to 2

workers/go/workerlib/kitchensink/kitchen_sink.go
fretz12: removed the fn arg

workers/go/workerlib/kitchensink/kitchen_sink.go
fretz12: makes sense, done

loadgen/scenario.go
```diff
@@ -474,7 +476,8 @@ func ensureNexusEndpoint(ctx context.Context, cl client.Client, namespace, runID
 			},
 		})
 	if err != nil {
-		if status.Code(err) == codes.AlreadyExists {
+		var alreadyExists *serviceerror.AlreadyExists
+		if errors.As(err, &alreadyExists) || status.Code(err) == codes.AlreadyExists {
```
fretz12: the client actually converts it to serviceerror.AlreadyExists, as caught during nightly test

THardy98 (approved):

PR #460, opened 2026-09-02
Adding tests for WorkflowQuery as a Nexus Op

mavemuri:
## What was changed
Added tests for Query as a Nexus Op 

## Why?
part of the SDK ergonomics efforts

## Checklist

1. Closes NEXUS-692

2. How was this tested: tested locally 

3. Any docs updates needed? No

### DIFF
--- a/internal/workertest/historyrequire.go
+++ b/internal/workertest/historyrequire.go
@@ -278,6 +278,20 @@ func looselyEqual(x, y any) bool {
 			return mapIsSuperset(x, yMap)
 		}
 		return false
+	case []any:
+		// Match the expected list as a prefix, element by element, so a spec can
+		// assert only what it knows. Eg, when asserting links, if runID is
+		// not known, it should be possible to assert on just WID+NS
+		yList, ok := y.([]any)
+		if !ok || len(yList) > len(x) {
+			return false
+		}
+		for i, yv := range yList {
+			if !looselyEqual(x[i], yv) {
+				return false
+			}
+		}
+		return true
 	}
 	return reflect.DeepEqual(x, y)
 }
--- a/internal/workertest/historyrequire.go
+++ b/internal/workertest/historyrequire.go
@@ -262,6 +262,8 @@ func mapIsSuperset(actual, expected map[string]any) bool {
 	return true
 }
 
+// looselyEqual is used by [mapIsSuperset] and in turn for
+// partial/full history matching
 func looselyEqual(x, y any) bool {
 	switch x := x.(type) {
 	case float64:
@@ -279,11 +281,11 @@ func looselyEqual(x, y any) bool {
 		}
 		return false
 	case []any:
-		// Match the expected list as a prefix, element by element, so a spec can
+		// Match the unordered expected list so a spec can
 		// assert only what it knows. Eg, when asserting links, if runID is
 		// not known, it should be possible to assert on just WID+NS
 		yList, ok := y.([]any)
-		if !ok || len(yList) > len(x) {
+		if !ok || len(yList) != len(x) {
 			return false
 		}
 		for i, yv := range yList {
--- a/loadgen/kitchen_sink_executor_test.go
+++ b/loadgen/kitchen_sink_executor_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 	"github.com/temporalio/omes/clioptions"
 	. "github.com/temporalio/omes/internal/workertest"
@@ -18,16 +19,9 @@ import (
 	"go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/history/v1"
 	"go.temporal.io/api/workflowservice/v1"
-	"go.temporal.io/sdk/client"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/emptypb"
-)
-
-const (
-	namespace = "default"
-
-	queryTargetWorkflowID = "nexus-query-target"
 )
 
 var (
@@ -80,6 +74,17 @@ type testCase struct {
 	historyMatcher          HistoryMatcher
 	expectedUnsupportedErrs map[clioptions.Language]string
 	expectedWorkflowError   string
+	populateFn              func(t *testing.T) testCase
+}
+
+// populate creates per-run test data for cases that require it.
+func (tc testCase) populate(t *testing.T) testCase {
+	t.Helper() // Keeps error lines pointing to the actual test table
+
+	if tc.populateFn == nil {
+		return tc
+	}
+	return tc.populateFn(t)
 }
 
 // TestKitchenSink tests specific kitchensink features across SDKs.
@@ -953,52 +958,8 @@ func TestKitchenSink(t *testing.T) {
 			expectedUnsupportedErrs: nexusUnsupportedSDKs,
 		},
 		{
-			name: "NexusOperation/QueryWorkflow",
-			testInput: &TestInput{
-				WorkflowInput: &WorkflowInput{
-					InitialActions: ListActionSet(
-						// start nexus op and wait for signal as a scaffold to
-						// query a running workflow
-						&Action{
-							Variant: &Action_NexusOperation{
-								NexusOperation: &ExecuteNexusOperation{
-									Operation:                       "echo-async",
-									HandlerWorkflowId:               queryTargetWorkflowID,
-									HandlerWorkflowIdConflictPolicy: enums.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
-									BeforeActions: ListActionSet(
-										NewSetWorkflowStateAction("query-result", "query successful"),
-									),
-									WaitForSignal: true,
-									AwaitableChoice: &AwaitableChoice{
-										Condition: &AwaitableChoice_WaitStarted{
-											WaitStarted: &emptypb.Empty{},
-										},
-									},
-								},
-							},
-						},
-						&Action{
-							Variant: &Action_NexusOperation{
-								NexusOperation: &ExecuteNexusOperation{
-									Operation: "query-workflow",
-									Input: QueryWorkflowNexusInput(QueryWorkflowTarget{
-										WorkflowID: queryTargetWorkflowID,
-									}),
-									ExpectedOutput: "query successful",
-									AwaitableChoice: &AwaitableChoice{
-										Condition: &AwaitableChoice_WaitFinish{
-											WaitFinish: &emptypb.Empty{},
-										},
-									},
-								},
-							},
-						}),
-				},
-			},
-			historyMatcher: PartialHistoryMatcher(`
-				NexusOperationScheduled {"operation":"query-workflow"}
-				NexusOperationCompleted {"links":[{"workflow":{"workflowId":"nexus-query-target","reason":"Query processed"}}]}`),
-			expectedUnsupportedErrs: nexusQueryUnsupportedSDKs,
+			name:       "NexusOperation/QueryWorkflow",
+			populateFn: populateQueryWorkflowTestCase,
 		},
 		{
 			name: "NexusOperation/Async",
@@ -1190,13 +1151,6 @@ func TestKitchenSink(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			// Ensure the workflow completes by appending a return action at the end.
-			input := tc.testInput
-			if input.WorkflowInput == nil {
-				input.WorkflowInput = &WorkflowInput{}
-			}
-			input.WorkflowInput.InitialActions = append(input.WorkflowInput.InitialActions, ListActionSet(NewEmptyReturnResultAction())...)
-
 			for _, sdk := range sdks {
 				if onlySDK != "" && string(sdk) != onlySDK {
 					continue // not using t.Skip as it's too noisy
@@ -1207,6 +1161,62 @@ func TestKitchenSink(t *testing.T) {
 				})
 			}
 		})
+	}
+}
+
+func populateQueryWorkflowTestCase(t *testing.T) testCase {
+	t.Helper()
+
+	handlerWorkflowID := fmt.Sprintf("nexus-query-target-%s", uuid.NewString())
+	return testCase{
+		name: "NexusOperation/QueryWorkflow",
+		testInput: &TestInput{
+			WorkflowInput: &WorkflowInput{
+				InitialActions: ListActionSet(
+					// Start a Nexus operation and wait for its signal as a scaffold to
+					// query a running workflow.
+					&Action{
+						Variant: &Action_NexusOperation{
+							NexusOperation: &ExecuteNexusOperation{
+								Operation:                       "echo-async",
+								HandlerWorkflowId:               handlerWorkflowID,
+								HandlerWorkflowIdConflictPolicy: enums.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
+								BeforeActions: ListActionSet(
+									NewSetWorkflowStateAction("query-result", "query successful"),
+								),
+								WaitForSignal: true,
+								AwaitableChoice: &AwaitableChoice{
+									Condition: &AwaitableChoice_WaitStarted{
+										WaitStarted: &emptypb.Empty{},
+									},
+								},
+							},
+						},
+					},
+					&Action{
+						Variant: &Action_NexusOperation{
+							NexusOperation: &ExecuteNexusOperation{
+								Operation: "query-workflow",
+								Input: QueryWorkflowNexusInput(QueryWorkflowTarget{
+									WorkflowID: handlerWorkflowID,
+								}),
+								ExpectedOutput: "query successful",
+								AwaitableChoice: &AwaitableChoice{
+									Condition: &AwaitableChoice_WaitFinish{
+										WaitFinish: &emptypb.Empty{},
+									},
+								},
+							},
+						},
+					},
+				),
+			},
+		},
+		historyMatcher: PartialHistoryMatcher(fmt.Sprintf(`
+			NexusOperationScheduled {"operation":"query-workflow"}
+			NexusOperationCompleted {"links":[{"workflow":{"workflowId":%q,"reason":"Query processed"}}]}`,
+			handlerWorkflowID)),
+		expectedUnsupportedErrs: nexusQueryUnsupportedSDKs,
 	}
 }
 
@@ -1259,6 +1269,8 @@ func testForSDK(
 	env *TestEnvironment,
 	workflowTimeout time.Duration,
 ) {
+	tc = tc.populate(t)
+
 	// Use mutex to ensure only one Java test runs at a time/a Gradle limitation.
 	if sdk == clioptions.LangJava {
 		javaMutex.Lock()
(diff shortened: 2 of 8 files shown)


loadgen/kitchensink/helpers.go
```diff
@@ -13,6 +14,22 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
+// QueryWorkflowTarget identifies the workflow that the "query-workflow" Nexus operation queries.
+// Only WorkflowID is required- by default, the operation queries the latest run in its own
```
stephanos:
```suggestion
// Only WorkflowID is required - by default, the operation queries the latest run in its own
```

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
 	return input.Input, nil
 })
 
+// QueryWorkflowOperationName is the registered name of QueryWorkflowOperation.
+const QueryWorkflowOperationName = "query-workflow"
```
stephanos: nit: I'd move that into the top `const` block

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+	}
+	// query links are to workflow as theres no history event
+	workflowLink := resp.GetLink().GetWorkflow()
+	if workflowLink == nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest,
+			"query response did not contain a workflow link; the server must support query-backed Nexus operations")
+	}
+	var state kitchensink.WorkflowState
+	if err := clioptions.OmesDataConverter().FromPayloads(resp.GetQueryResult(), &state); err != nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "failed to decode query result: %v", err)
```

<details>
<summary><strong>med</strong> — Handler-side failures are reported as <code>BadRequest</code> while genuine caller errors are left unmapped and retried forever.</summary>

Two problems with the error classification in this operation:

**Handler-side conditions blamed on the caller.** Failing to decode the query result (line 696), a query result missing the `query-result` key (line 700), and a response missing the workflow link (line 691) are all `HandlerErrorTypeBadRequest`. None of these is a problem with the caller's request — the request was well-formed and validated at lines 661–664. `BadRequest` is non-retryable and misattributes the fault, which makes the failure harder to read in a caller's history. The missing-link case in particular is a server-capability problem, which the message itself says.

**Caller-side conditions left unmapped.** Every `QueryWorkflow` error other than `NamespaceNotActive` falls through to `return "", err` at line 686. `temporalnexus` converts an unrecognized error into a retryable 500, so a target workflow that doesn't exist (`serviceerror.NotFound`) or a query the target rejects (`serviceerror.QueryFailed`) retries until the caller workflow's execution timeout instead of failing fast. In this test that's 30s of retries ending in a timeout rather than a clear "workflow not found".

</details>

**Suggestion:** Map the terminal service errors explicitly and use `HandlerErrorTypeInternal` for handler-side faults:

```go
	if err != nil {
		switch {
		// Namespace handover is transient.
		case errors.Is[*serviceerror.NamespaceNotActive](err):
			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeUnavailable, "%s", err.Error())
		}
		if _, ok := errors.AsType[*serviceerror.NamespaceNotActive](err); ok {
			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeUnavailable, "%s", err.Error())
		}
		if _, ok := errors.AsType[*serviceerror.NotFound](err); ok {
			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeNotFound, "%s", err.Error())
		}
		if _, ok := errors.AsType[*serviceerror.QueryFailed](err); ok {
			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "%s", err.Error())
		}
		return "", err
	}
```
and switch lines 691, 696 and 700 to `nexus.HandlerErrorTypeInternal`.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -2,25 +2,33 @@ package kitchensink
...
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/temporalnexus"
 	"go.temporal.io/sdk/workflow"
 )
 
-const KitchenSinkServiceName = "kitchen-sink"
+const (
+	KitchenSinkServiceName = "kitchen-sink"
+	nexusQueryName         = "report_state"
```
stephanos: How about `reportStateQueryName = "report_state"` and use it in `workflow.SetQueryHandler(ctx, "report_state",` as well?

internal/workertest/historyrequire.go
```diff
@@ -278,6 +278,20 @@ func looselyEqual(x, y any) bool {
 			return mapIsSuperset(x, yMap)
 		}
 		return false
+	case []any:
+		// Match the expected list as a prefix, element by element, so a spec can
+		// assert only what it knows. Eg, when asserting links, if runID is
+		// not known, it should be possible to assert on just WID+NS
+		yList, ok := y.([]any)
+		if !ok || len(yList) > len(x) {
```
stephanos: How about?
```go
		yList, ok := y.([]any)
		if !ok || len(yList) != len(x) {
			return false
		}
```

I see that the map matches a superset, but that feels out of place for lists IMO. This change would also allow writing `{"links":[]}`.

If this becomes too strict later, we could add options to change this behavior, but for now I feel like stricter is better.

loadgen/kitchensink/helpers.go
```diff
@@ -13,6 +14,22 @@ import (
...
+type QueryWorkflowTarget struct {
+	WorkflowID string `json:"workflow_id"`
+	Namespace  string `json:"namespace,omitempty"`
+	RunID      string `json:"run_id,omitempty"`
+	QueryType  string `json:"query_type,omitempty"`
+}
+
+// QueryWorkflowNexusInput encodes a target as the input of a "query-workflow" Nexus operation.
+func QueryWorkflowNexusInput(target QueryWorkflowTarget) string {
+	encoded, _ := json.Marshal(target)
```
stephanos: How about a panic here instead if there's an error?

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+		Query: &query.WorkflowQuery{QueryType: target.QueryType},
+	})
+	if err != nil {
+		// Treat namespace handover as retryable.
+		if _, ok := errors.AsType[*serviceerror.NamespaceNotActive](err); ok {
+			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeUnavailable, "%s", err.Error())
+		}
+		return "", err
+	}
+	// query links are to workflow as theres no history event
```
stephanos: nit: writing comments as full sentences is a good habit IMO.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+	workflowLink := resp.GetLink().GetWorkflow()
+	if workflowLink == nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest,
+			"query response did not contain a workflow link; the server must support query-backed Nexus operations")
+	}
+	var state kitchensink.WorkflowState
+	if err := clioptions.OmesDataConverter().FromPayloads(resp.GetQueryResult(), &state); err != nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "failed to decode query result: %v", err)
+	}
+	result, ok := state.GetKvs()["query-result"]
```
stephanos, line 693: WDYT about making `"query-result"` a const too?

loadgen/kitchen_sink_executor_test.go
```diff
@@ -19,11 +19,16 @@ import (
...
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
-const namespace = "default"
+const (
+	namespace = "default"
```
stephanos: I realize this was here before, but could we remove this and instead add

```go
func (env *TestEnvironment) Namespace() string {
     return testNamespace
}
```
and then use `env.Namespace()` instead of the constant?

Then we only have this defined in one place.

internal/workertest/historyrequire.go
```diff
@@ -278,6 +278,20 @@ func looselyEqual(x, y any) bool {
 			return mapIsSuperset(x, yMap)
 		}
 		return false
+	case []any:
```
stephanos, line 283: nit: it took me a bit to understand why this change was made; a little PR comment would have been useful here.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -939,6 +952,54 @@ func TestKitchenSink(t *testing.T) {
...
+			testInput: &TestInput{
+				WorkflowInput: &WorkflowInput{
+					InitialActions: ListActionSet(
+						// start nexus op and wait for signal as a scaffold to
+						// query a running workflow
+						&Action{
+							Variant: &Action_NexusOperation{
+								NexusOperation: &ExecuteNexusOperation{
+									Operation:                       "echo-async",
+									HandlerWorkflowId:               queryTargetWorkflowID,
```
stephanos: I think the per-SDK runs here are racing and use the same `queryTargetWorkflowID` across runs. Since we use a singular server namespace.

There are a few ways we could go about this. I think the best would be to have sth like func: `testInputFn: func(si ScenarioInfo) *TestInput {...}` - then we can use ` si.RunID` to make unique workflow IDs.

We could have an invariant to either require `testInput` or `testInputFn` for each test case.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -19,11 +19,16 @@ import (
...
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
-const namespace = "default"
+const (
+	namespace = "default"
+
+	queryTargetWorkflowID = "nexus-query-target"
```
stephanos: nit: mind if we inline this? For tests I prefer inlined strings over consts as tests are meant to be self-contained and short.

internal/workertest/historyrequire.go
```diff
@@ -278,6 +278,20 @@ func looselyEqual(x, y any) bool {
...
+		// assert only what it knows. Eg, when asserting links, if runID is
+		// not known, it should be possible to assert on just WID+NS
+		yList, ok := y.([]any)
+		if !ok || len(yList) > len(x) {
+			return false
+		}
+		for i, yv := range yList {
+			if !looselyEqual(x[i], yv) {
+				return false
+			}
```
stephanos, line 294: WDYT about matching elements in any order? We only have a single link right now; but I can see this fail if there are 2 links and their order isn't deterministic.

loadgen/kitchensink/helpers.go
```diff
@@ -13,6 +14,22 @@ import (
...
 )
 
+// QueryWorkflowTarget identifies the workflow that the "query-workflow" Nexus operation queries.
+// Only WorkflowID is required- by default, the operation queries the latest run in its own
+// namespace with the "report_state" query
+type QueryWorkflowTarget struct {
+	WorkflowID string `json:"workflow_id"`
+	Namespace  string `json:"namespace,omitempty"`
+	RunID      string `json:"run_id,omitempty"`
+	QueryType  string `json:"query_type,omitempty"`
```
stephanos: It looks like these are unused?

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1220,8 +1281,13 @@ func testForSDK(
 	// worker picks them up.
 	runTaskQueue := TaskQueueForRun(scenarioInfo.RunID)
 
+	// PrepareTestInput fills the per-run endpoint and task queue into the input in
+	// place, and the case's input is shared by every SDK subtest, so work on a copy.
+	testInput, ok := proto.Clone(tc.testInput).(*TestInput)
```
stephanos: Nice catch!

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1220,8 +1281,13 @@ func testForSDK(
...
+	require.True(t, ok, "failed to clone test input")
```
stephanos: nit: seems redundant, as I don't think that can fail?

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+	}
+	if target.WorkflowID == "" {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "query target must include a workflow ID")
+	}
+	if target.Namespace == "" {
+		target.Namespace = temporalnexus.GetOperationInfo(ctx).Namespace
+	}
+	if target.QueryType == "" {
+		target.QueryType = nexusQueryName
+	}
```
stephanos: nit: as mentioned elsewhere, this seems redundant. But I'll still mention that this can be expressed better with "modern Go" as `target.QueryType = cmp.Or(target.QueryType, nexusQueryName)`

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+		if _, ok := errors.AsType[*serviceerror.NamespaceNotActive](err); ok {
+			return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeUnavailable, "%s", err.Error())
+		}
+		return "", err
+	}
+	// query links are to workflow as theres no history event
+	workflowLink := resp.GetLink().GetWorkflow()
+	if workflowLink == nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest,
+			"query response did not contain a workflow link; the server must support query-backed Nexus operations")
```
stephanos: I think `BadRequest` isn't quite right here; as the error seems to be server-side and non-retryable. Maybe:

```
return "", &nexus.HandlerError{
    Type:          nexus.HandlerErrorTypeInternal,
    RetryBehavior: nexus.HandlerErrorRetryBehaviorNonRetryable,
    Cause:         errors.New("query response did not contain a workflow link; the server must support query-backed Nexus operations"),
}
```

workers/go/workerlib/kitchensink/kitchen_sink.go
stephanos: This should maybe also be a `HandlerErrorTypeInternal` as above IMO.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -633,6 +650,60 @@ var EchoSyncOperation = nexus.NewSyncOperation("echo-sync", func(ctx context.Con
...
+	}
+	var state kitchensink.WorkflowState
+	if err := clioptions.OmesDataConverter().FromPayloads(resp.GetQueryResult(), &state); err != nil {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "failed to decode query result: %v", err)
+	}
+	result, ok := state.GetKvs()["query-result"]
+	if !ok {
+		return "", nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "query result did not contain query-result state")
```
stephanos: Same here. Or can this fail intermittently?

workers/go/workerlib/kitchensink/kitchen_sink.go
mavemuri: Was hesitant as that is in an unrelated test case that just happens to have same query/handler - just in case one changes in the future? Will rename this one

workers/go/workerlib/kitchensink/kitchen_sink.go
mavemuri: I think this one might be ok as-is because its used once in test and once in the worker - constant would mean adding package/etc but there isnt any other usage currently

internal/workertest/historyrequire.go
mavemuri: Im not sure when that could happen - multiple links on same event in non-deterministic order. Maybe server should always return in some order - chronological or alphabetical? But not clear on this- could we revisit when its required?

loadgen/kitchensink/helpers.go
```diff
@@ -13,6 +14,21 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
+// QueryWorkflowTarget identifies the workflow that the "query-workflow" Nexus operation queries.
+// Only WorkflowID is required
```
stephanos, line 18: nit
```suggestion
// WorkflowID is required.
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -67,6 +74,17 @@ type testCase struct {
 	historyMatcher          HistoryMatcher
 	expectedUnsupportedErrs map[clioptions.Language]string
 	expectedWorkflowError   string
+	populateFn              func(t *testing.T) testCase
```
stephanos, line 77: Hm, I don't love it. I think maybe this was the wrong direction; instead we should isolate the SDK runs by using separate namespaces. I noodled a bit one it and came up with this: https://github.com/temporalio/omes/pull/462

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1149,6 +1164,62 @@ func TestKitchenSink(t *testing.T) {
...
+			WorkflowInput: &WorkflowInput{
+				InitialActions: ListActionSet(
+					// Start a Nexus operation and wait for its signal as a scaffold to
+					// query a running workflow.
+					&Action{
+						Variant: &Action_NexusOperation{
+							NexusOperation: &ExecuteNexusOperation{
+								Operation:                       "echo-async",
+								HandlerWorkflowId:               handlerWorkflowID,
+								HandlerWorkflowIdConflictPolicy: enums.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
```

<details>
<summary><strong>nit</strong> — `USE_EXISTING` conflict policy is dead weight given the ID is a fresh UUID.</summary>

`handlerWorkflowID` is `nexus-query-target-<uuid>` generated per SDK run, so there is never an existing execution to reuse. This is the only test case that sets a conflict policy, which makes it read as though it matters here.

</details>

**Suggestion:** Drop the line, or add a short comment if it's there deliberately to exercise the field.

internal/workertest/historyrequire.go
stephanos: That's fine with me; can we tighten the comment there?

> // Match the unordered expected list so a spec can
		// assert only what it knows. Eg, when asserting links, if runID is
		// not known, it should be possible to assert on just WID+NS

I don't quite understand what it's saying tbh. Esp. since they are not "unordered".

I'd omit it or add a shorter comment maybe on `if !looselyEqual(x[i], yv) {` to explain that this recurses into list elements (if that's what it means to say). Or drop the comment.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1149,6 +1164,62 @@ func TestKitchenSink(t *testing.T) {
 	}
 }
 
+func populateQueryWorkflowTestCase(t *testing.T) testCase {
```
stephanos: Mind if we inline that? I don't quite see the benefit for it being a named function. The downside here is added navigation effort.

loadgen/kitchen_sink_executor_test.go
mavemuri: Agree, per-SDK ns is cleaner - thanks for raising the pr

PR #462, opened 2026-09-03
Run kitchen sink SDK tests in isolated namespaces

stephanos:
## What changed?

Runs each SDK's kitchen sink tests in an isolated namespace while sharing one dev server.

## Why?

This prevents cross-test contamination and allows using static identifiers for each test as they are guaranteed to be unique.


internal/workertest/env.go
```diff
@@ -64,6 +66,21 @@ func WithDynamicConfig(values map[string]any) TestEnvOption {
...
+// WithNamespace configures the namespace used by the test environment.
+func WithNamespace(namespace string) TestEnvOption {
+	return func(c *testEnvConfig) {
+		c.namespace = namespace
+	}
+}
+
+// WithDevServerGroup configures the test environment to share a dev server
+// with other environments created by the same test using the same group.
+func WithDevServerGroup(group string) TestEnvOption {
```
stephanos: I considered moving the devserver start into its own method and pass that into `SetupTestEnvironment` but I didn't want to update all callers (there are many).

loadgen/kitchen_sink_executor_test.go
```diff
@@ -75,15 +75,26 @@ func TestKitchenSink(t *testing.T) {
...
-	}))
+	}
+	testEnvironments := make(map[clioptions.Language]*TestEnvironment)
+	for _, sdk := range sdks {
+		if onlySDK != "" && string(sdk) != onlySDK {
+			continue // not using t.Skip as it's too noisy
+		}
+		testEnvironments[sdk] = SetupTestEnvironment(t,
+			WithDevServerGroup("kitchen-sink"),
+			WithNamespace(fmt.Sprintf("default.%s", sdk)),
```

<details>
<summary><strong>high</strong> — History lookup still queries the hardcoded <code>"default"</code> namespace, which no longer exists on the shared dev server.</summary>

Workflows now run in `default.<sdk>`, but `getWorkflowHistory` still sets `Namespace: namespace` explicitly on the `ListWorkflowExecutionsRequest`, where `namespace` is the package-level `const namespace = "default"` (line 26, used at line 1354). An explicit `Namespace` on the request overrides the client's default, so the query never sees the per-SDK namespace.

Worse, `"default"` is never registered anymore: the pooled `devserver.Start` is called with `opts.Namespace` set to the *first* SDK's namespace (`default.go`, or whatever `SDK=` selects), and `acquire` only registers the remaining `default.<sdk>` names. So the call fails with `NamespaceNotFound` rather than just returning an empty list.

`getWorkflowHistory` is on the success path of every `testSupportedFeature` subtest (line 1309), and `historyErr` is asserted with `require.NoError(t, historyErr, "failed to get workflow history")` — so every non-`expectedUnsupportedErrs` case fails. This only reproduces when the test actually runs (CI skips it unless `SDK` is set), which is why it can slip through.

As an aside: the `default.` prefix reads like it's scoping under a `default` namespace that doesn't exist. Something like `kitchen-sink-<sdk>` would be clearer.

</details>

**Suggestion:** Thread the environment's namespace through instead of the constant, and drop the now-dead `const namespace`:

```go
// line 1309
historyEvents, historyErr := getWorkflowHistory(t, taskQueueName, env.TemporalClient(), env.Namespace())

// line 1351
func getWorkflowHistory(t *testing.T, taskQueueName string, temporalClient client.Client, namespace string) ([]*history.HistoryEvent, error) {
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1136,10 +1147,7 @@ func TestKitchenSink(t *testing.T) {
 			}
 			input.WorkflowInput.InitialActions = append(input.WorkflowInput.InitialActions, ListActionSet(NewEmptyReturnResultAction())...)
 
-			for _, sdk := range sdks {
-				if onlySDK != "" && string(sdk) != onlySDK {
-					continue // not using t.Skip as it's too noisy
-				}
+			for sdk, env := range testEnvironments {
```

<details>
<summary><strong>small</strong> — Ranging over the map makes subtest registration order nondeterministic.</summary>

The previous `for _, sdk := range sdks` gave a stable subtest order; Go randomizes map iteration, so `-run` output and log interleaving now differ run to run. `sdks` is already the canonical order and is still in scope.

</details>

**Suggestion:**

```suggestion
			for _, sdk := range sdks {
				env, ok := testEnvironments[sdk]
				if !ok {
					continue
				}
```

internal/workertest/serverpool.go
```diff
@@ -0,0 +1,80 @@
...
+	servers: make(map[testDevServerGroupKey]*pooledTestDevServer),
+}
+
+func (p *testDevServerPool) acquire(
+	t *testing.T,
+	group string,
+	namespace string,
+	opts devserver.Options,
+) (*devserver.Server, error) {
+	p.Lock()
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — Package-global pool mutex is held across the dev server clone+build and namespace registration.</summary>

`p.Lock()` is held for the whole function, including `devserver.Start` (line 64), which clones and builds `temporalio/temporal` on a cold cache and can take minutes, and `RegisterNamespace` (line 57), which does network I/O with a 30s deadline. Because `groupedTestDevServers` is a package-level singleton, any other top-level test in the package acquiring a *different* group blocks behind it, even though the two servers are unrelated.

Today the only caller is `TestKitchenSink`, which acquires sequentially, so this is latent rather than active — but the singleton plus `*testing.T`-keyed map exists precisely to support concurrent/independent callers, and it doesn't.

</details>

**Suggestion:** Store a per-key entry under the lock with a `sync.Once` (or a `chan struct{}` ready signal), release the lock, then do `Start`/`RegisterNamespace` outside it. Alternatively, if only "one test wants N namespaces on one server" is needed, a plain `SetupSharedTestEnvironments(t, namespaces ...string) map[string]*TestEnvironment` would drop the global state, the `*testing.T` map key, and the group-mismatch error paths entirely.

internal/workertest/env.go
```diff
@@ -120,18 +142,33 @@ func SetupTestEnvironment(t *testing.T, opts ...TestEnvOption) *TestEnvironment
...
+	}
+
+	var server *devserver.Server
+	if cfg.devServerGroup == "" {
+		server, err = devserver.Start(t.Context(), serverOpts)
+		if err == nil {
+			t.Cleanup(func() { _ = server.Stop() })
+		}
+	} else {
+		server, err = groupedTestDevServers.acquire(
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — "Failed to start dev server" mislabels the group-conflict errors from <code>acquire</code>.</summary>

`acquire` can return `dev server group %q already uses different dynamic config` / `... already uses namespace %q` — misuse of the API, not a startup failure. The assertion message on line 167 sends the reader looking at the server rather than at the conflicting `SetupTestEnvironment` call.

</details>

**Suggestion:** Reword line 167 to `require.NoError(t, err, "Failed to acquire dev server")`.

devserver/devserver.go
```diff
@@ -219,6 +219,23 @@ func (s *Server) FrontendHostPort() string {
...
 
+// RegisterNamespace registers a namespace if it does not already exist.
+func (s *Server) RegisterNamespace(ctx context.Context, namespace string) error {
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
+
+	go func() {
+		select {
+		case procErr := <-s.done:
+			s.done <- procErr
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — Exporting <code>RegisterNamespace</code> makes the "process exited" fast-path unreliable: <code>Stop()</code> can consume <code>s.done</code> first, so the caller blocks for the full 30s retry window instead of failing fast.</summary>

The steal-and-requeue trick on `s.done` was safe when this was private and only reachable from `Start`, where `Stop` cannot run concurrently. Now that it is exported and called mid-test from `serverpool.acquireNamespace`, `Stop()` and this watcher goroutine race for the single buffered value:

- `Stop()` does `s.cancel(); err := <-s.done`. If it wins the receive, this goroutine stays parked in `select` and `cancel(...)` never fires.
- `registerNamespace` then keeps retrying against a dead frontend — `codes.Unavailable` is retryable — until its own 30s `context.WithTimeout` expires.

So a dev server that dies while a namespace is being registered turns a fast, well-described failure into a 30s stall with a generic deadline error. In the pooled path that stall happens while `pooledTestDevServer.mu` is held, so every other environment in the group waits on it too.

A single-value channel is the wrong primitive for a lifetime signal that now has multiple observers.

</details>

**Suggestion:** Give `Server` a broadcast lifetime signal instead of re-sending on `done` — e.g. a `exited chan struct{}` closed by the `cmd.Wait()` goroutine (alongside storing the error), and have both `Stop()` and `RegisterNamespace` observe that. Then any number of concurrent watchers see the exit, and `RegisterNamespace` can cancel immediately with the real process error.

internal/workertest/serverpool.go
```diff
@@ -0,0 +1,108 @@
...
+	test  *testing.T
+	group string
+}
+
+type testDevServerPool struct {
+	sync.Mutex
+	servers map[testDevServerGroupKey]*pooledTestDevServer
+}
+
+var groupedTestDevServers = testDevServerPool{
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — A package-global pool keyed by <code>*testing.T</code> is a lot of concurrency machinery for one call site that creates all its environments serially; an explicit shared-server option would be simpler and equally non-breaking.</summary>

The sole consumer is `TestKitchenSink`, which builds every environment in one loop on one goroutine before any subtest starts. To serve that, this file adds a process-global map, a `*testing.T` map key, a `sync.Once`, a second mutex, cleanup-driven map eviction, and a `reflect.DeepEqual` dynamic-config equality check — all of which exist only to re-derive "these environments want the same server".

The `*testing.T` key also makes sharing quietly positional: the group is shared only among `SetupTestEnvironment` calls passing the *identical* `T`. A subtest that passes its own `t` with the same group string silently gets a second dev server rather than joining the group, and nothing reports it. That is a subtle failure mode for a helper whose whole purpose is "share one server".

Your inline note says you avoided passing the server in because there are many callers. That concern is addressable without the pool: a new option is opt-in, so all ten existing `SetupTestEnvironment` call sites in `scenarios/` stay untouched.

</details>

**Suggestion:** Export a `workertest.StartDevServer(t, opts...) *devserver.Server` that holds the `versions.Get` / logger / `NewLogWriter` plumbing currently inlined in `SetupTestEnvironment`, plus a `WithDevServer(*devserver.Server)` option that skips the start path. `TestKitchenSink` then reads:

```go
server := StartDevServer(t, WithDynamicConfig(dynamicConfig))
for _, sdk := range enabledSDKs {
    testEnvironments[sdk] = SetupTestEnvironment(t,
        WithDevServer(server),
        WithNamespace(...),
    )
}
```

`SetupTestEnvironment` registers the namespace when `WithDevServer` is set, and `serverpool.go` goes away entirely along with the global, the `T` key, and the config-equality check.

internal/workertest/serverpool.go
```diff
@@ -0,0 +1,108 @@
...
+	servers: make(map[testDevServerGroupKey]*pooledTestDevServer),
+}
+
+func (p *testDevServerPool) acquire(
+	t *testing.T,
+	group string,
+	namespace string,
+	opts devserver.Options,
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — <code>acquire</code> takes a full <code>devserver.Options</code> but only <code>DynamicConfigValues</code> is validated; every other field from non-first callers is silently discarded.</summary>

The first caller into `start.Do` wins, and `acquireNamespace` compares only `dynamicConfig`. `Ref`, `Persistence`, `ClusterEndpoint`, `Output`, and `Logger` from later callers are dropped without a word. Today `Ref` always comes from `versions.Get("SERVER_VERSION")` so it happens to match, but a future test that passes `WithDevServerGroup` plus a different persistence driver or cluster endpoint will get a server that ignores it and no diagnostic.

Relatedly, the two error messages say *that* something differs but not *what*, which is unhelpful when a caller has a large dynamic-config map.

</details>

**Suggestion:** Compare the whole struct rather than one field — zero out the caller-scoped fields (`Namespace`, `Output`, `Logger`) and `reflect.DeepEqual` the rest, or accept only the fields that legitimately vary per environment so incompatible options are unrepresentable. Either way, include the diverging value in the error text.

internal/workertest/env.go
```diff
@@ -64,6 +66,21 @@ func WithDynamicConfig(values map[string]any) TestEnvOption {
...
 
+// WithNamespace configures the namespace used by the test environment.
+func WithNamespace(namespace string) TestEnvOption {
+	return func(c *testEnvConfig) {
+		c.namespace = namespace
+	}
+}
+
+// WithDevServerGroup configures the test environment to share a dev server
+// with other environments created by the same test using the same group.
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — <code>WithDevServerGroup</code>'s doc omits the two constraints that make it fail: every environment in a group needs a distinct <code>WithNamespace</code>, and dynamic configs must match exactly.</summary>

Both are hard errors in `acquireNamespace`, and both are easy to trip: using `WithDevServerGroup` without `WithNamespace` leaves every environment on `defaultTestNamespace`, so the second `SetupTestEnvironment` fails with `already uses namespace "default"` — a confusing message for a caller who never mentioned namespaces. Callers shouldn't have to read `serverpool.go` to learn that.

</details>

**Suggestion:**

```go
// WithDevServerGroup configures the test environment to share a dev server
// with other environments created by the same test using the same group.
// Each environment in a group must use a distinct WithNamespace, and all of
// them must pass the same WithDynamicConfig values; the dev server is
// configured from the first environment in the group.
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -75,15 +73,26 @@ func TestKitchenSink(t *testing.T) {
...
 		// Enable StartNexusOperationExecution for the standalone-nexus subtests.
 		"nexusoperation.enableStandalone": true,
 		// Standalone Nexus system callbacks require CHASM callbacks.
 		"history.enableCHASMCallbacks": true,
 		// Enable StartActivityExecution for the standalone-activity subtest.
 		"activity.enableStandalone":                        true,
 		"history.enableStandaloneActivityOperatorCommands": true,
-	}))
+	}
+	testEnvironments := make(map[clioptions.Language]*TestEnvironment)
```

<details>
<summary><strong>small</strong> — The <code>onlySDK</code> filter is now applied in two places, and a typo'd <code>SDK</code> value makes the whole test pass vacuously with zero subtests.</summary>

`SDK=pyhton` (or a language that was renamed) yields an empty `testEnvironments`, the loop at line 1148 skips every SDK via `!ok`, and `TestKitchenSink` reports PASS having asserted nothing. CI runs this per-SDK via the env var, so a silent green is exactly the wrong outcome. The pre-existing single-loop version had the same hole, but now that the filter is duplicated the two copies can also drift.

</details>

**Suggestion:** Filter once into an ordered slice, then guard it:

```go
var enabledSDKs []clioptions.Language
for _, sdk := range sdks {
    if onlySDK != "" && string(sdk) != onlySDK {
        continue // not using t.Skip as it's too noisy
    }
    enabledSDKs = append(enabledSDKs, sdk)
}
require.NotEmpty(t, enabledSDKs, "SDK=%q matches no known SDK", onlySDK)
```

Then both loops range over `enabledSDKs` and the `env, ok :=` lookup at line 1149 becomes a plain indexed access.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -75,15 +73,26 @@ func TestKitchenSink(t *testing.T) {
...
+	for _, sdk := range sdks {
+		if onlySDK != "" && string(sdk) != onlySDK {
+			continue // not using t.Skip as it's too noisy
+		}
+		testEnvironments[sdk] = SetupTestEnvironment(t,
+			WithDevServerGroup("kitchen-sink"),
+			WithNamespace(fmt.Sprintf("default.%s", sdk)),
```

<details>
<summary><strong>nit</strong> — <code>default.<sdk></code> reads like a qualified form of the default namespace, but the shared dev server has no <code>default</code> namespace at all.</summary>

The server is started with the first environment's namespace, so on this pool the only namespaces that exist are `default.go`, `default.java`, and so on. A name like `kitchensink-go` says what it is and won't be mistaken for a sub-scope of `default` when it shows up in server logs or `tctl` output.

</details>

**Suggestion:** `WithNamespace(fmt.Sprintf("kitchensink-%s", sdk))`

internal/workertest/env.go
```diff
@@ -122,16 +162,29 @@ func SetupTestEnvironment(t *testing.T, opts ...TestEnvOption) *TestEnvironment
 	serverLogger := testLogger.Named("devserver").Sugar()
 	server, err := devserver.Start(t.Context(), devserver.Options{
 		Ref:                 serverRef,
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — Server-scoped options such as <code>WithDynamicConfig</code> are silently dropped when <code>WithDevServer</code> is set.</summary>

`TestEnvOption` now feeds two entry points with disjoint relevant subsets: `StartDevServer` only reads `namespace` and `dynamicConfig`, and `SetupTestEnvironment` only forwards them when it has to start its own server. So

```go
SetupTestEnvironment(t, WithDevServer(server), WithDynamicConfig(...))
```

compiles, reads correctly, and silently runs against whatever dynamic config the shared server happened to be started with. The reverse holds too — `StartDevServer(t, WithExecutorTimeout(...), WithNexusEndpoint(...))` ignores both.

That is a slow failure mode: the symptom is a kitchen sink assertion failing deep in a feature-gated subtest, not an error at setup.

</details>

**Suggestion:** Fail fast in the shared-server branch:

```go
} else {
	require.Nil(t, cfg.dynamicConfig,
		"WithDynamicConfig is ignored when WithDevServer is set; pass it to StartDevServer instead")
	err := server.RegisterNamespace(t.Context(), cfg.namespace)
	require.NoError(t, err, "Failed to register namespace")
}
```

Alternatively, split the server-only knobs into their own `DevServerOption` type so the compiler catches the mistake instead of a runtime assertion.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -75,7 +73,17 @@ func TestKitchenSink(t *testing.T) {
...
-	env := SetupTestEnvironment(t, WithDynamicConfig(map[string]any{
+
+	var enabledSDKs []clioptions.Language
+	for _, sdk := range sdks {
+		if onlySDK != "" && string(sdk) != onlySDK {
+			continue // not using t.Skip as it's too noisy
+		}
+		enabledSDKs = append(enabledSDKs, sdk)
+	}
+	require.NotEmpty(t, enabledSDKs, "SDK=%q matches no known SDK", onlySDK)
```

<details>
<summary><strong>small</strong> — <code>SDK=ts</code> and the other accepted aliases now fail the test with a message claiming they are not known SDKs.</summary>

`enabledSDKs` is filtered with an exact `string(sdk) != onlySDK` comparison, but `clioptions.Language.Set` (`clioptions/sdk.go:24`) accepts `py`, `ts`, `cs`, and `rb` as aliases, and those are what the CLI accepts everywhere else in the repo.

Before this change `SDK=ts go test ./loadgen -run TestKitchenSink` silently ran zero subtests. Now it hard-fails with `SDK="ts" matches no known SDK`, which is factually wrong and sends the reader looking for a typo that isn't there.

The guard itself is a good addition — turning a silent no-op into a failure is the right call. It just needs to compare against the normalized value.

</details>

**Suggestion:**

```go
var only clioptions.Language
if onlySDK != "" {
	require.NoError(t, only.Set(onlySDK), "invalid SDK env var")
}

var enabledSDKs []clioptions.Language
for _, sdk := range sdks {
	if only != "" && sdk != only {
		continue // not using t.Skip as it's too noisy
	}
	enabledSDKs = append(enabledSDKs, sdk)
}
require.NotEmpty(t, enabledSDKs, "SDK=%q matches no known SDK", onlySDK)
```

internal/workertest/env.go
```diff
@@ -106,13 +128,31 @@ func (env *TestEnvironment) NexusEndpointName() string {
...
+}
+
+func startDevServer(t *testing.T, namespace string, opts ...DevServerOption) *devserver.Server {
+	var cfg devServerConfig
+	for _, opt := range opts {
+		opt(&cfg)
+	}
 
 	serverRef, err := versions.Get("SERVER_VERSION")
 	require.NoError(t, err)
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — Dev server exit errors are discarded, so an abnormal server exit is never reported.</summary>

`devserver.classifyExitErr` exists specifically to distinguish a clean shutdown or our own SIGTERM from a crash, a non-zero exit, or the SIGKILL that fires when the server misses `WaitDelay`. Dropping `Stop()`'s return value means none of that reaches the test.

With a shared server this matters more than it did when each `TestEnvironment` owned its own: if the server dies mid-run, every SDK's subtests fail with unrelated gRPC errors and nothing in the output points at the server as the cause.

</details>

**Suggestion:**

```go
t.Cleanup(func() { assert.NoError(t, server.Stop(), "dev server exited abnormally") })
```

`github.com/stretchr/testify/assert` is already used in this package's tests; prefer it over `require` here so the remaining cleanups still run.

devserver/devserver.go
```diff
@@ -219,6 +221,22 @@ func (s *Server) FrontendHostPort() string {
...
 
+// RegisterNamespace registers a namespace if it does not already exist.
+func (s *Server) RegisterNamespace(ctx context.Context, namespace string) error {
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
+
+	go func() {
+		select {
+		case <-s.exited:
+			cancel(fmt.Errorf("process exited before namespace registration completed: %w", s.exitErr))
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — <code>%w</code> renders as <code>%!w(&lt;nil&gt;)</code> when the server exited cleanly.</summary>

`s.exitErr` is `cmd.Wait()`'s result, which is `nil` for a zero exit status — exactly what the Temporal server produces when it shuts down gracefully on SIGTERM. Calling `RegisterNamespace` on an already-stopped server then produces `process exited before namespace registration completed: %!w(<nil>)`, and the returned error isn't unwrappable.

This is reachable now that `RegisterNamespace` is exported and called from `SetupTestEnvironment`'s shared-server path: a `t.Context()` cancellation tears the server down gracefully before a later `SetupTestEnvironment` call.

</details>

**Suggestion:**

```go
case <-s.exited:
	err := errors.New("process exited before namespace registration completed")
	if s.exitErr != nil {
		err = fmt.Errorf("%w: %w", err, s.exitErr)
	}
	cancel(err)
```

internal/workertest/env.go
```diff
@@ -122,16 +162,29 @@ func SetupTestEnvironment(t *testing.T, opts ...TestEnvOption) *TestEnvironment
 	serverLogger := testLogger.Named("devserver").Sugar()
 	server, err := devserver.Start(t.Context(), devserver.Options{
 		Ref:                 serverRef,
-		Namespace:           testNamespace,
+		Namespace:           namespace,
 		DynamicConfigValues: cfg.dynamicConfig,
 		Output:              workerctl.NewLogWriter(serverLogger),
 		Logger:              serverLogger,
 	})
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — Registration failure doesn't say which namespace failed.</summary>

`TestKitchenSink` now registers one namespace per SDK in a loop, and neither `registerNamespace` nor this assertion mentions the namespace name, so a failure gives no hint about which iteration broke. `devserver.Start` wraps with `register namespace %q` for exactly this reason; the newly exported path lost that context.

</details>

**Suggestion:**

```go
require.NoErrorf(t, err, "Failed to register namespace %q", cfg.namespace)
```

THardy98 (approved):

PR #463, opened 2026-09-04
Make Nexus attach signal resilient to active/passive transitions

michaely520:
## What changed

- Send the Nexus attach-handler unblock signal without awaiting its result.
- Continue awaiting all Nexus operation futures as the correctness gate.
- Add regression coverage for the action sequence.

## Why

During an active/passive cluster transition, the unblock signal can be replayed after the handler workflow has already completed. Awaiting that duplicate signal incorrectly fails the load workflow even when all Nexus operations succeeded.

## Testing

Focused scenario and Nexus throughput tests pass.


THardy98 (approved):
change lgtm, just curious about the context behind why we want this

michaely520:
> change lgtm, just curious about the context behind why we want this

it is possible after failover depending on our luck that certain parts of the workflow replay, meaning there can be a duplicate signal sent causing omes to break

PR #464, opened 2026-09-04
Add Nexus workflow messaging actions

stephanos:
Extend `NexusWorkflowAction` with signal and update variants using the existing `DoSignal` and `DoUpdate` action types.


loadgen/kitchensink/helpers.go
```diff
@@ -423,3 +424,108 @@ func ConvertToPayload(newInput any) *common.Payload {
...
+		// this should never happen; but we don't want to swallow the error
+		panic(fmt.Sprintf("failed to marshal Nexus operation target %T: %v", target, err))
+	}
+	return &Action{
+		Variant: &Action_NexusOperation{
+			NexusOperation: &ExecuteNexusOperation{
+				Endpoint:        endpoint,
+				Operation:       operation,
+				Input:           string(input),
+				ExpectedOutput:  expectedOutput,
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — <code>ExpectedOutput</code> is set to the literal <code>"$self"</code>, so the output assertion is wrong for every self-targeted action (and silently never runs today).</summary>

`newNexusTargetAction` uses `target.WorkflowID` as `ExpectedOutput`. For `SignalWorkflowTarget{WorkflowID: SelfWorkflowIDToken}` and `UpdateWorkflowTarget{WorkflowID: SelfWorkflowIDToken}` — the shape used by both `scenarios/throughput_stress.go:954`/`:971` and the two `$self` executor test cases — that stores `"$self"`.

The substitution in `workers/go/workerlib/kitchensink/kitchen_sink.go:482` only rewrites `nexusOp.Input`, not `nexusOp.ExpectedOutput`. The handler then echoes the *substituted* ID (`temporalnexus.NewSyncResult(target.WorkflowID)`), so the comparison at `kitchen_sink.go:500` would be `"$self" != "kitchensinktest-…"` → the action fails with `expected output "$self", got "<real id>"`.

The reason this doesn't show up in the executor tests is a second problem: in `withAwaitableChoiceCustom` (`kitchen_sink.go:429-475`) the `afterCompletedWaiter` — the only closure that reads `ExpectedOutput` — is invoked exclusively on the `CancelAfterCompleted` branch. `WaitFinish` (and a nil `AwaitableChoice`) falls through to `err = fut.Get(ctx, nil)`, which discards the result. So all three new actions ship an assertion that never executes, and that is incorrect the moment it does. The pre-existing `echo-sync`/`echo-async` `ExpectedOutput: "hello"` in `throughput_stress.go` is dead for the same reason.

</details>

**Suggestion:** Either substitute the token in `ExpectedOutput` too, alongside `Input`:

```go
opInput := strings.ReplaceAll(
    nexusOp.Input, kitchensink.SelfWorkflowIDToken, workflow.GetInfo(ctx).WorkflowExecution.ID)
expOutput := strings.ReplaceAll(
    nexusOp.GetExpectedOutput(), kitchensink.SelfWorkflowIDToken, workflow.GetInfo(ctx).WorkflowExecution.ID)
```

…or drop the `expectedOutput` argument from `newNexusTargetAction` rather than carrying a value that is never checked. Separately, worth wiring `afterCompletedWaiter` into the `WaitFinish` branch so `ExpectedOutput` actually asserts something — otherwise the "every operation echoes back the workflow ID it acted on" contract in the doc comment above is unverified.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -465,8 +478,11 @@ func handleNexusOperation(ctx workflow.Context, nexusOp *kitchensink.ExecuteNexu
 	return withAwaitableChoiceCustom(ctx, state, func(ctx workflow.Context) workflow.NexusOperationFuture {
 		client := workflow.NewNexusClient(nexusOp.Endpoint, KitchenSinkServiceName)
 		nexusOptions := workflow.NexusOperationOptions{}
+		// Let a spec target the calling workflow without knowing its ID.
+		opInput := strings.ReplaceAll(
```

<details>
<summary><strong>small</strong> — Blanket <code>strings.ReplaceAll</code> rewrites <code>$self</code> in the input of <em>every</em> Nexus operation, not just the three target-shaped ones.</summary>

`nexusOp.Input` is an opaque string shared by all operations — `echo-sync`/`echo-async` pass arbitrary payloads through it. Any spec whose echo input happens to contain `$self` now gets it silently rewritten to a generated workflow ID, which is a confusing failure to debug because the substitution is invisible in the spec.

It also substitutes inside the JSON string value rather than at the field level, so a real workflow ID containing `$self` (unlikely, but IDs are user-controlled in this repo — e.g. `nexus-sws-` prefixes are already hand-built) would be mangled on the way out.

</details>

**Suggestion:** Scope the substitution to the operations that define the token — e.g. only rewrite when `nexusOp.Operation` is one of `kitchensink.SignalWorkflowOperationName`, `SignalWithStartWorkflowOperationName`, or `UpdateWorkflowOperationName` — or resolve `$self` on the target struct after unmarshalling rather than on the raw input string.

scenarios/throughput_stress.go
```diff
@@ -898,6 +946,38 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
 	}
 }
 
+// createNexusSignalAction signals the caller's own workflow through a Nexus
+// operation. Self-signalling keeps the action self-contained: the caller is a
+// kitchen sink workflow, so it already handles "do_actions_signal".
+func (t *tpsExecutor) createNexusSignalAction() *Action {
```

<details>
<summary><strong>small</strong> — The three new throughput_stress flags have no scenario-level test, unlike the flag they were modelled on.</summary>

`IncludeNexusStandaloneActivityFlag` ships with both an end-to-end run (`TestThroughputStressNexusStandaloneActivity`, `scenarios/throughput_stress_test.go:150`) and an action-shape assertion (`TestThroughputStressNexusStandaloneActivityActions`, `:190`). The three flags added here have neither, so nothing checks that `IncludeNexusSignal`/`IncludeNexusSignalWithStart`/`IncludeNexusUpdate` actually reach `createActionsChunk`, that the new `Configure` gate loop wires each flag to the right field, or that `Run` clears them when the endpoint can't be resolved.

The gate loop in particular is easy to get wrong silently — a copy/paste that pointed two entries at the same `*bool` would compile and pass every existing test, and the failure would only surface as a nightly quietly not exercising a primitive, which is exactly the scenario the per-primitive flags exist to prevent.

The executor tests cover the kitchen-sink side well; this is about the scenario wiring only.

</details>

**Suggestion:** Add an actions-shape test in the style of `TestThroughputStressNexusStandaloneActivityActions` that sets all three config fields and asserts one action per operation name appears in the generated chunk, plus a table case asserting `include-nexus-*` without `nexus-enabled` returns the gating error.

internal/workertest/historyrequire.go
```diff
@@ -278,6 +278,19 @@ func looselyEqual(x, y any) bool {
 			return mapIsSuperset(x, yMap)
 		}
 		return false
+	case []any:
+		// Compare element-wise so a spec can assert only the fields it knows.
+		// Asserting a link, for instance, should not require knowing the run ID.
+		yList, ok := y.([]any)
+		if !ok || len(yList) != len(x) {
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — The list case requires an exact length match, which contradicts the "assert only what you know" comment above it and makes link assertions brittle.</summary>

Every other branch of `looselyEqual` is subset-flavoured (`mapIsSuperset`), but lists must match element count exactly. So the new `{"links":[…]}` assertions break the moment the server attaches a second link to `NexusOperationCompleted` — a change unrelated to what the test is actually asserting.

The comment ("so a spec can assert only the fields it knows") describes the recursion into the elements, not the list itself; a reader could easily take it to mean partial lists work.

</details>

**Suggestion:** Either reword to say the list length must match exactly, or make it a prefix/subset match so a spec can assert the links it cares about:

```go
	case []any:
		// Element-wise, so an element can be a partial map. The lengths must match:
		// a spec asserts the whole list, just not every field of each element.
		yList, ok := y.([]any)
```

scenarios/throughput_stress.go
```diff
@@ -898,6 +946,38 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
...
+// createNexusUpdateAction updates the caller's own workflow through a Nexus
+// operation. The caller is a kitchen sink workflow, so it handles the update this
+// operation sends.
+func (t *tpsExecutor) createNexusUpdateAction() *Action {
+	return NewNexusUpdateWorkflowAction(t.config.NexusEndpoint,
+		UpdateWorkflowTarget{WorkflowID: SelfWorkflowIDToken}, waitFinishChoice())
+}
+
+// waitFinishChoice awaits an operation through to completion.
+func waitFinishChoice() *AwaitableChoice {
```

<details>
<summary><strong>nit</strong> — <code>waitFinishChoice</code> lives in <code>scenarios</code>, so the executor tests can't use it and spell the literal out three times.</summary>

`loadgen/kitchen_sink_executor_test.go:1157`, `:1176` and `:1194` each repeat `&AwaitableChoice{Condition: &AwaitableChoice_WaitFinish{WaitFinish: &emptypb.Empty{}}}` next to the new `NewNexus*Action` constructors that this helper was written for.

</details>

**Suggestion:** Move it to `loadgen/kitchensink/helpers.go` alongside the new action constructors (exported, e.g. `WaitFinishChoice()`), and use it from both the scenario and the executor tests.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1417,11 +1487,19 @@ func (w *kitchenSinkTestWrapper) Run(ctx context.Context, info ScenarioInfo) err
...
+	taskQueueName string,
+	executionID string,
+) ([]*history.HistoryEvent, error) {
 	executions, err := temporalClient.ListWorkflow(t.Context(),
 		&workflowservice.ListWorkflowExecutionsRequest{
 			Namespace: namespace,
-			Query:     fmt.Sprintf("TaskQueue = '%s' AND WorkflowType = 'kitchenSink'", taskQueueName),
+			Query: fmt.Sprintf(
+				"TaskQueue = '%s' AND WorkflowType = 'kitchenSink' AND %s = '%s'",
+				taskQueueName, OmesExecutionIDSearchAttribute, executionID),
```

<details>
<summary><strong>small</strong> — The new <code>OmesExecutionID</code> filter always queries for the empty string, because <code>ScenarioInfo.ExecutionID</code> is never set in this test.</summary>

`testForSDK` builds `scenarioInfo` with only `ScenarioName`, `RunID` and `Configuration` (`loadgen/kitchen_sink_executor_test.go:1342`), and `RunExecutorTest` only fills in `Logger`/`MetricsHandler`/`Client`/`Namespace` (`internal/workertest/env.go:272-276`). So `scenarioInfo.ExecutionID` is `""` and the query is literally `... AND OmesExecutionID = ''`.

It happens to do the right thing today: `Run.DefaultStartWorkflowOptions` sets the search attribute to `r.ExecutionID` (also `""`) on the caller workflow, while the kitchenSink workflows started *through Nexus* carry no such attribute at all — so the empty-string comparison excludes them. But that is an accident of both sides being empty rather than the identity filter the code reads as, and it breaks the moment the server normalizes an empty keyword to "unset" (every case would then fail with `no workflow executions found`).

</details>

**Suggestion:** Set a real execution ID in `testForSDK` so the filter is meaningful:

```go
	scenarioInfo := ScenarioInfo{
		ScenarioName: "kitchenSinkTest",
		RunID:        fmt.Sprintf("%s-%d", strings.ReplaceAll(t.Name(), "/", "-"), time.Now().Unix()),
		ExecutionID:  uuid.NewString(),
		Configuration: RunConfiguration{
			Iterations: 1,
		},
	}
```

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -629,24 +629,153 @@ func startNexusOperation(
...
+		ctx, input.GetWorkflowId(), input.GetRunId(), signalName, signalArg)
+	if err != nil {
+		return result, nexusOutboundError("SignalWorkflow", err)
+	}
+	return temporalnexus.NewSyncResult(kitchensink.ConvertToPayload(input.GetWorkflowId())), nil
+}
+
+// nexusSignalArg defaults to an empty action set so an operation can exercise
+// signal delivery without requiring the target workflow to run another action.
+func nexusSignalArg(signal *kitchensink.DoSignal_DoSignalActions) *kitchensink.DoSignal_DoSignalActions {
```

<details>
<summary><strong>small</strong> — <code>nexusSignalArg</code>'s parameter is dead: the only call site passes <code>nil</code>, so the non-nil branch is unreachable.</summary>

`signalWorkflowNexusOperation` calls it exactly once, as `nexusSignalArg(nil)` (line 666), and the `DoSignalActions` case a few lines below assigns `signalArg = doActions` directly rather than going through the helper. The `if signal != nil` branch can never be taken, and the eager call means the default is built even when it is immediately overwritten by the custom or do-actions branch.

</details>

**Suggestion:** Drop the parameter and inline the default at the point where it's needed:

```go
	signal := input.GetSignal()
	signalName := "do_actions_signal"
	// Default to an empty action set so an operation can exercise signal
	// delivery without requiring the target workflow to run another action.
	var signalArg any = &kitchensink.DoSignal_DoSignalActions{
		Variant: &kitchensink.DoSignal_DoSignalActions_DoActions{
			DoActions: kitchensink.SingleActionSet(),
		},
	}
	if custom := signal.GetCustom(); custom != nil {
		signalName = custom.GetName()
		signalArg = custom.GetArgs()
	} else if doActions := signal.GetDoSignalActions(); doActions != nil {
		signalArg = doActions
	}
```

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -564,6 +564,8 @@ message NexusWorkflowAction {
   NexusWorkflowStartOptions start_options = 3;
   oneof action {
     google.protobuf.Empty start = 4;
+    DoSignal signal = 5;
```

<details>
<summary><strong>small</strong> — The new <code>signal</code>/<code>update</code> oneof variants carry semantics that only exist in the Go handler, and this proto is the contract the other five SDK workers implement against.</summary>

Sibling messages in this file document each oneof arm (see `NexusOperationRequest` right above, and every field of `NexusWorkflowStartOptions` below). The behaviour a Python/Java/TS/.NET/Ruby implementer would have to reverse-engineer from `workers/go/workerlib/kitchensink/kitchen_sink.go` today:

- `signal` honours `DoSignal.with_start`, and in that case `start_options` supplies the workflow input while the conflict policy is forced to `USE_EXISTING`.
- `update` does *not* support `DoUpdate.with_start` — it is rejected as a bad request.
- `run_id` applies to the non-with-start signal and to the update.
- An unset `DoSignal` variant is not an error; it delivers an empty `do_actions_signal`.

</details>

**Suggestion:**

```suggestion
    // Signal the target workflow. Honors DoSignal.with_start, in which case
    // start_options supplies the workflow input.
    DoSignal signal = 5;
    // Update the target workflow. DoUpdate.with_start is not supported.
    DoUpdate update = 6;
```

loadgen/kitchensink/helpers.go
```diff
@@ -101,6 +101,95 @@ func NexusStartWorkflowRequest(workflowID string, options *NexusWorkflowStartOpt
...
+			WorkflowId: workflowID,
+			RunId:      runID,
+			Action:     &NexusWorkflowAction_Update{Update: update},
+		}},
+	}
+}
+
+// NewNexusUpdateResultAction encodes a Payload as the workflow update result.
+// The server forwards that result directly to the Nexus completion callback.
+func NewNexusUpdateResultAction(value any) *Action {
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — The double <code>ConvertToPayload</code> is the surprising part of this helper and the doc comment doesn't explain it.</summary>

Every other producer of a Nexus operation result in this PR wraps exactly once — e.g. `signalWorkflowNexusOperation` returns `temporalnexus.NewSyncResult(kitchensink.ConvertToPayload(input.GetWorkflowId()))`, and the signal test cases assert against a single `ConvertToPayload("nexus-signal-target")`. Here the update test also asserts a single-wrapped `ConvertToPayload("nexus-update-target")`, yet `ReturnThis` is wrapped twice, so one layer must be peeled somewhere in the update-result → Nexus-completion → caller path.

That extra encode/decode hop is exactly the thing a future reader (or another SDK's implementer) will get wrong, and the current comment ("encodes a Payload as the workflow update result") reads as if a single wrap would do.

</details>

**Suggestion:** Spell out where the extra layer goes, e.g.:

```go
// NewNexusUpdateResultAction returns a Payload as the workflow update result.
// The update handler's return value is itself encoded by the data converter
// before the server forwards it to the Nexus completion callback, so the value
// is wrapped twice here: the caller decodes the outer layer and compares the
// inner Payload against ExecuteNexusOperation.expected_output.
func NewNexusUpdateResultAction(value any) *Action {
	return NewReturnResultAction(ConvertToPayload(ConvertToPayload(value)))
}
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1319,30 +1396,37 @@ func testForSDK(
...
-									}
+			var prepareActions func([]*Action)
+			prepareActions = func(actions []*Action) {
+				for _, action := range actions {
+					if nexusOp := action.GetNexusOperation(); nexusOp != nil && nexusOp.Endpoint == "" {
+						nexusOp.Endpoint = nexusEndpoint
+					}
+					if nested := action.GetNestedActionSet(); nested != nil {
+						prepareActions(nested.GetActions())
+					}
```
stephanos: New addition here.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1202,6 +1213,108 @@ func TestKitchenSink(t *testing.T) {
 			DoStandaloneActivityOperatorCommands_COMMAND_TYPE_RESET),
 		standaloneActivityOperatorCommandsTestCase("Update",
 			DoStandaloneActivityOperatorCommands_COMMAND_TYPE_UPDATE),
+		// These tests start one kitchenSink workflow, apply an action to it, then
+		// await pending actions.
```
stephanos: fix: drop this comment

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1436,24 +1538,12 @@ func (w *kitchenSinkTestWrapper) Run(ctx context.Context, info ScenarioInfo) err
...
-	if len(executions.Executions) == 0 {
-		return nil, fmt.Errorf("no workflow executions found for task queue %s", taskQueueName)
-	}
-	if len(executions.Executions) > 1 {
-		t.Logf("Warning: found %d kitchenSink workflow executions for task queue %s, using the first one", len(executions.Executions), taskQueueName)
-	}
-
-	execution := executions.Executions[0]
-	historyIter := temporalClient.GetWorkflowHistory(t.Context(), execution.Execution.WorkflowId, execution.Execution.RunId, false, enums.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+func getWorkflowHistory(
```
stephanos, line 1544: Simpler!

loadgen/kitchensink/helpers.go
```diff
@@ -51,6 +51,24 @@ func ActivityNameAndArgs(act *ExecuteActivityAction) (string, []any) {
...
+
+func UpdateNameAndArgs(update *DoUpdate) (string, []any, error) {
+	if actions := update.GetDoActions(); actions != nil {
+		return "do_actions_update", []any{actions}, nil
+	} else if handler := update.GetCustom(); handler != nil {
+		return handler.GetName(), []any{handler.GetArgs()}, nil
+	}
+	return "", nil, fmt.Errorf("do_update must recognizable variant")
+}
+
```
stephanos, line 71: ^ new helpers for re-use

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -623,22 +623,89 @@ func startNexusOperation(
 		return temporalnexus.NewSyncResult(kitchensink.ConvertToPayload(action.Echo)), nil
 	case *kitchensink.NexusOperationRequest_WorkflowAction:
 		workflowAction := cmp.Or(action.WorkflowAction, &kitchensink.NexusWorkflowAction{})
-		if workflowAction.GetStart() == nil {
-			break
+		switch workflowAction.GetAction().(type) {
```
stephanos, line 630: This file is the main change:

internal/workertest/historyrequire.go
```diff
@@ -278,6 +279,11 @@ func looselyEqual(x, y any) bool {
 			return mapIsSuperset(x, yMap)
 		}
 		return false
+	case []any:
```
mavemuri, line 282: 👍

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -623,27 +623,124 @@ func startNexusOperation(
...
+			ID:                       cmp.Or(input.GetWorkflowId(), opts.RequestID),
+			TaskQueue:                startOptions.GetTaskQueue(),
+			WorkflowExecutionTimeout: 60 * time.Minute,
+			WorkflowIDConflictPolicy: startOptions.GetWorkflowIdConflictPolicy(),
+		},
+		KitchenSinkWorkflow,
+		cmp.Or(startOptions.GetWorkflowInput(), &kitchensink.WorkflowInput{}))
+}
+
+func signalWorkflowNexusOperation(
```
mavemuri, line 666: nit: maybe a small comment that this covers both signal and signal-with-start (may not be obvious because unlike signal, we dont support update-with-start yet)

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -623,27 +623,124 @@ func startNexusOperation(
...
+			signalName,
+			signalArg,
 			client.StartWorkflowOptions{
-				ID:                       cmp.Or(workflowAction.GetWorkflowId(), opts.RequestID),
-				TaskQueue:                startOptions.GetTaskQueue(),
+				ID:                       input.GetWorkflowId(),
+				TaskQueue:                cmp.Or(input.GetStartOptions().GetTaskQueue(), temporalnexus.GetOperationInfo(ctx).TaskQueue),
 				WorkflowExecutionTimeout: 60 * time.Minute,
-				WorkflowIDConflictPolicy: startOptions.GetWorkflowIdConflictPolicy(),
+				WorkflowIDConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
```
mavemuri: Just checking if this is intentional - in the base pr, looks like this is in the [proto](https://github.com/temporalio/omes/pull/361/changes#diff-9be522fb1d9855c524d20e5c4a25d914e21a5970da60a00fdf5b15269a110fbbR569) - why hardcode it now? maybe a similar cmp.Or like above instead?

mavemuri (approved):

workers/go/workerlib/kitchensink/kitchen_sink.go
stephanos: Good point 👍

workers/go/workerlib/kitchensink/kitchen_sink.go
stephanos: Good question! It was because Signal-with-Start I _thought_ required `USE_EXISTING`, but I checked and it also supports `TERMINATE_EXISTING`. I'll make the change to use the start options 👍

stephanos: (server default is `USE_EXISTING` if it's not defined)

PR #466, opened 2026-09-05
Exercise Nexus workflow messaging in throughput stress

stephanos:
Add opt-in `throughput_stress` actions for Nexus signal, signal-with-start, and update.


scenarios/throughput_stress.go
```diff
@@ -521,6 +552,9 @@ func (t *tpsExecutor) createActionsChunk(
 
 	// Create actions for the current chunk
 	for i := 0; i < itersPerChunk; i++ {
+		nexusWorkflowID := fmt.Sprintf("%s-nexus-target-%d",
```

<details>
<summary><strong>med</strong> — Deterministic target workflow IDs collide with leftovers from a previous iteration attempt, and the plain start path has no conflict policy to survive it.</summary>

`nexusWorkflowID` is a pure function of `run.DefaultStartWorkflowOptions().ID` (`w-<runID>-<execID>-<iteration>`) and `internalIterationIndex`, so it is identical across attempts of the same iteration. `GenericExecutor`'s `retryLoop` (loadgen/generic_executor.go:179) re-invokes `Execute` for the same `*Run`, which re-runs `UpdateWorkflowOptions` → `t.createActions(run)` → the same target IDs.

If a previous attempt died before the target sequence reached its tail `SendSignal`, the target workflow is still open — `nexusWorkflowOptions` gives it a 60-minute `WorkflowExecutionTimeout`. The retry's start operation then hits `WorkflowExecutionAlreadyStarted`: `NewNexusWorkflowTargetSequence`'s default start action passes no `WorkflowIdConflictPolicy`, so `nexusWorkflowOptions` leaves it UNSPECIFIED (fail), and `nexusOutboundError` doesn't classify already-started as a bad request, so it becomes a retryable handler error that keeps retrying until the operation times out.

Note the asymmetry: the signal-with-start path is immune because the handler forces `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING`, and the neighbouring `createNexusAttachCallbacksAction` sidesteps this entirely with `uuid.NewString()`.

Separately, this is computed for every internal iteration even when all three flags are off, ~85 lines above its only use at line 643.

</details>

**Suggestion:** Mirror `createNexusAttachCallbacksAction` and make the ID unique per attempt, and move it next to its use:

```go
			if t.config.IncludeNexusSignal || t.config.IncludeNexusSignalWithStart || t.config.IncludeNexusUpdate {
				nexusWorkflowID := fmt.Sprintf("%s-nexus-target-%d-%s",
					run.DefaultStartWorkflowOptions().ID,
					t.internalIterationIndex(run, remainingInternalIters, i),
					uuid.NewString())
				asyncActions = append(asyncActions, t.createNexusWorkflowTargetSequence(nexusWorkflowID, rng))
			}
```

`TestThroughputStressNexusWorkflowTargetsAreDistinctAcrossChunks` would then assert distinctness/count rather than exact IDs. If you want to keep IDs deterministic for debuggability, pass an explicit start action with `WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING` instead of relying on the helper's default.

scenarios/throughput_stress.go
```diff
@@ -605,6 +639,9 @@ func (t *tpsExecutor) createActionsChunk(
 					)
 				}
 			}
+			if t.config.IncludeNexusSignal || t.config.IncludeNexusSignalWithStart || t.config.IncludeNexusUpdate {
+				asyncActions = append(asyncActions, t.createNexusWorkflowTargetSequence(nexusWorkflowID, rng))
```

<details>
<summary><strong>med</strong> — Two <code>AwaitPendingActions</code> now run concurrently in the same action set and drain the same workflow-global future list.</summary>

`NewNexusWorkflowTargetSequence` ends with `Action_AwaitPendingActions`, and this sequence is appended to `asyncActions`, which is executed as a `Concurrent: true` set. `createNexusAttachCallbacksAction` (line 890) — always present when `NexusEnabled` — already ends with its own `AwaitPendingActions` in that same concurrent set.

`handleAwaitPendingActions` (workers/go/workerlib/kitchensink/kitchen_sink.go:506) is not scoped to the enclosing action set: it snapshots and clears `ws.pendingActions`, which is workflow-global state that every `WaitStarted` action appends to. So the two branches steal each other's futures.

Failure scenario: the target sequence's default start action uses `WaitStarted`, appending its Nexus future to the shared list. The attach-callbacks branch reaches its `AwaitPendingActions` first, drains that future along with its own three, and blocks until the *target* workflow completes; if the target's start operation failed, the error surfaces as `pending action 3 failed` from the attach-callbacks branch, pointing at an unrelated operation. The target sequence's own `AwaitPendingActions` then finds an empty list and verifies nothing.

Today the union of futures still gets awaited by *someone*, so nothing is silently dropped — but the guarantee is accidental and depends on coroutine interleaving. Any future branch that appends a `WaitStarted` future after the last drain would have its failure dropped entirely, which is exactly the kind of miss this stress harness exists to catch.

</details>

**Suggestion:** Either append the target sequence to `chunkActions` (sequential) instead of `asyncActions`, so only one `AwaitPendingActions` is ever in flight, or make the await scoped in the worker — have `handleActionSet` push/pop a `pendingActions` frame so `AwaitPendingActions` only drains futures created within its own set.

scenarios/throughput_stress.go
```diff
@@ -248,6 +263,19 @@ func (t *tpsExecutor) Configure(info loadgen.ScenarioInfo) error {
 	if config.IncludeNexusStandaloneActivity && !config.NexusEnabled {
 		return fmt.Errorf("%s requires %s", IncludeNexusStandaloneActivityFlag, NexusEnabledFlag)
 	}
+	for _, nexusWorkflowAction := range []struct {
```

<details>
<summary><strong>small</strong> — Anonymous-struct/<code>*bool</code> loop is heavier than the surrounding style and its flag→field wiring is untested.</summary>

Every other option in `Configure` (including `IncludeNexusStandaloneActivity` on the two lines directly above, which has the identical `requires nexus-enabled` check) is read with a plain assignment plus an `if`. The loop is the same line count but adds a pointer indirection, and a copy-paste slip in the table — e.g. `{IncludeNexusUpdateFlag, &config.IncludeNexusSignal}` — would compile and silently mis-map a flag. `TestThroughputStressNexusWorkflowTargetSequence` sets all three flags to `true`, so it wouldn't catch that, and unlike `IncludeStandaloneNexusFlag` (see `TestThroughputStressConfigureExplicitStandaloneNexusRequiresNexusEnabled`) there is no test for the new validation errors.

</details>

**Suggestion:**

```go
	config.IncludeNexusSignal = info.OptionBool(IncludeNexusSignalFlag)
	config.IncludeNexusSignalWithStart = info.OptionBool(IncludeNexusSignalWithStartFlag)
	config.IncludeNexusUpdate = info.OptionBool(IncludeNexusUpdateFlag)
	for _, flag := range []struct {
		name    string
		enabled bool
	}{
		{IncludeNexusSignalFlag, config.IncludeNexusSignal},
		{IncludeNexusSignalWithStartFlag, config.IncludeNexusSignalWithStart},
		{IncludeNexusUpdateFlag, config.IncludeNexusUpdate},
	} {
		if flag.enabled && !config.NexusEnabled {
			return fmt.Errorf("%s requires %s", flag.name, NexusEnabledFlag)
		}
	}
```

and add a table test that enables one flag at a time with `nexus-enabled=false`, asserting both the error and that only the matching config field is set.

docs/throughput-stress.md
```diff
@@ -90,15 +90,24 @@ is any.
...
+support for standalone activities and activity completion callbacks (dynamic config
+`activity.enableStandalone` and `activity.enableCallbacks`) and a Nexus callback URL; if those are
+off, the operation fails clearly rather than being skipped.
+
+The workflow actions use one target kitchenSink workflow per iteration for all selected actions.
+With signal-with-start enabled, exactly one signal-with-start request is made: it either creates the
+target or messages a target created by a regular Nexus workflow start.
+
+All four options are off by default, require `nexus-enabled`, and are currently supported by Go
+workers.
```
stephanos: fix: let's drop this disclaimer

docs/throughput-stress.md
```diff
@@ -90,15 +90,24 @@ is any.
...
-Currently only supported and run by Go workers.
+The standalone activity action is driven two ways each iteration: as an in-workflow Nexus operation
+and, when standalone Nexus is part of the run, as a standalone Nexus operation. It also needs server
+support for standalone activities and activity completion callbacks (dynamic config
+`activity.enableStandalone` and `activity.enableCallbacks`) and a Nexus callback URL; if those are
+off, the operation fails clearly rather than being skipped.
+
+The workflow actions use one target kitchenSink workflow per iteration for all selected actions.
+With signal-with-start enabled, exactly one signal-with-start request is made: it either creates the
+target or messages a target created by a regular Nexus workflow start.
```
stephanos: fix: let's drop this paragraph

scenarios/throughput_stress.go
```diff
@@ -620,6 +653,13 @@ func (t *tpsExecutor) createActionsChunk(
 					)
 				}
 			}
+			if t.config.IncludeNexusSignal || t.config.IncludeNexusSignalWithStart || t.config.IncludeNexusUpdate {
+				nexusWorkflowID := fmt.Sprintf("%s-nexus-target-%d-%s",
+					run.DefaultStartWorkflowOptions().ID,
+					t.internalIterationIndex(run, remainingInternalIters, i),
+					uuid.NewString())
```
stephanos: fix: is it safe to use `uuid`? what if it failed, will it end up abandoned?

scenarios/throughput_stress_test.go
```diff
@@ -241,6 +242,131 @@ func TestThroughputStressNexusStandaloneActivityActions(t *testing.T) {
...
+			IncludeNexusSignalFlag:          "true",
+			IncludeNexusSignalWithStartFlag: "true",
+			IncludeNexusUpdateFlag:          "true",
+		}),
+	}))
+
+	const workflowID = "nexus-target"
+	seenSignalWithStartCreator := map[bool]bool{}
+	for seed := int64(0); seed < 100; seed++ {
+		actions := executor.createNexusWorkflowTargetSequence(
```
stephanos: fix: why not mimic `TestThroughputStressNexusStandaloneActivityActions` and not reach into `createNexusWorkflowTargetSequence` but just run the full test via `newThroughputStressExecutor`

scenarios/throughput_stress_test.go
```diff
@@ -353,6 +479,50 @@ func TestThroughputStressConfigureExplicitStandaloneNexusRequiresNexusEnabled(t
 	require.Contains(t, err.Error(), NexusEnabledFlag)
 }
 
+func TestThroughputStressConfigureNexusWorkflowActionsRequireNexusEnabled(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name    string
+		flag    string
+		enabled [3]bool
```
stephanos: fix: this is awkward

scenarios/throughput_stress.go
```diff
@@ -930,6 +969,104 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
...
+	}
+	if startAction == nil {
+		startAction = NexusOperation(&ExecuteNexusOperation{
+			Endpoint: t.config.NexusEndpoint,
+			Input: &NexusOperationRequest{
+				Action: &NexusOperationRequest_WorkflowAction{WorkflowAction: &NexusWorkflowAction{
+					WorkflowId: workflowID,
+					StartOptions: &NexusWorkflowStartOptions{
+						WorkflowInput: &WorkflowInput{InitialActions: ListActionSet(
+							NewAwaitWorkflowStateAction("status", "done"),
```

<details>
<summary><strong>high</strong> — Nothing in the sequence ever completes the target workflow, so the trailing <code>await_pending_actions</code> blocks until the target hits its 60-minute execution timeout.</summary>

The start operation launches the target with `AwaitWorkflowState("status", "done")` and registers the future as pending (`wait_started` → `ws.pendingActions`, `workers/go/workerlib/kitchensink/kitchen_sink.go:447-451`). But nothing that follows ever satisfies that await:

- `createNexusSignalAction` / `createNexusSignalWithStartAction` send `do_actions_signal` with `DoActions: SingleActionSet()` — an *empty* action set, so the target's signal goroutine runs no actions and never sets state or returns a result.
- `createNexusUpdateAction` only returns a value from the update handler (`kitchen_sink.go:77-83`); it does not complete the workflow.

So the target stays parked on `workflow.Await` (`kitchen_sink.go:316-323`) until the handler-imposed `WorkflowExecutionTimeout: 60 * time.Minute` (`kitchen_sink.go:648`) fires; the pending start future never resolves, and `Action_AwaitPendingActions` blocks the whole caller iteration until then. This hits every iteration that takes the plain-start path — i.e. all of them unless `include-nexus-signal-with-start` is on and wins the coin flip.

The signal-with-start creator path has the mirror problem: `StartOptions` there is `&WorkflowInput{}` (no initial actions), so that target also never completes — it just leaks one open workflow per internal iteration instead of blocking the caller.

Earlier revisions of this sequence ended with a `SendSignal` carrying `NewReturnResultSignal()`; cb32178 replaced it with the `status=done` await but never added anything that sets `status`. The existing `createNexusAttachCallbacksAction` (`scenarios/throughput_stress.go:900-956`) still uses the working shape: `wait_started` ops → signal the handler with `SingleActionSet(NewEmptyReturnResultAction())` → `await_pending_actions`.

A terminating action appended by the sequence itself (rather than folded into one of the Nexus signals) also keeps the update-only and signal-only flag combinations working, and must stay last so the update still targets a running workflow.

</details>

**Suggestion:** Unblock the target explicitly before awaiting pending actions, and drop the now-unneeded `status`/`done` await from the start input:

```go
	actions := append([]*Action{startAction}, targetActions...)
	actions = append(actions,
		// Complete the target so the wait_started start operation can finish.
		&Action{Variant: &Action_SendSignal{SendSignal: &SendSignalAction{
			WorkflowId: workflowID,
			SignalName: "do_actions_signal",
			Args: []*common.Payload{ConvertToPayload(&DoSignal_DoSignalActions{
				Variant: &DoSignal_DoSignalActions_DoActions{
					DoActions: SingleActionSet(NewEmptyReturnResultAction()),
				},
			})},
			AwaitableChoice: &AwaitableChoice{Condition: &AwaitableChoice_WaitFinish{WaitFinish: &emptypb.Empty{}}},
		}}},
		&Action{Variant: &Action_AwaitPendingActions{AwaitPendingActions: &AwaitPendingActions{}}},
	)
```

Worth asserting in `TestThroughputStressNexusWorkflowActions` that the sequence contains a completing action, since the current shape-only assertions pass with a target that can never finish.

scenarios/throughput_stress.go
```diff
@@ -930,6 +969,104 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
...
+	if t.config.IncludeNexusUpdate {
+		targetActions = append(targetActions, t.createNexusUpdateAction(workflowID))
+	}
+	if startAction == nil {
+		startAction = NexusOperation(&ExecuteNexusOperation{
+			Endpoint: t.config.NexusEndpoint,
+			Input: &NexusOperationRequest{
+				Action: &NexusOperationRequest_WorkflowAction{WorkflowAction: &NexusWorkflowAction{
+					WorkflowId: workflowID,
+					StartOptions: &NexusWorkflowStartOptions{
```

<details>
<summary><strong>med</strong> — Deterministic target workflow ID with no <code>workflow_id_conflict_policy</code> makes iteration retries and resumes fail with <code>WorkflowExecutionAlreadyStarted</code>.</summary>

The target ID is now fully deterministic (`<iteration workflow ID>-nexus-target-<internal index>`), which is what makes retries safe — but only if the start can attach to an existing execution. `startWorkflowNexusOperation` passes `WorkflowIDConflictPolicy: startOptions.GetWorkflowIdConflictPolicy()` straight through (`workers/go/workerlib/kitchensink/kitchen_sink.go:653`), and this call site leaves it unset (`UNSPECIFIED` → fail).

Failure scenario: internal iteration 5 fails anywhere after the start operation (a failing update, the hang described above, a worker restart). `generic_executor.go:178-196` retries the same iteration up to `MaxIterationAttempts`, and the iteration workflow reattaches by ID, so `createActionsChunk` rebuilds the identical `…-nexus-target-N` ID. The first attempt's target is still open (it has a 60-minute execution timeout), so the retry's start operation fails with `WorkflowExecutionAlreadyStarted` and every subsequent attempt fails the same way. The same applies when the scenario resumes, since `RunID`/`ExecutionID` are unchanged.

Both neighbouring uses of a fixed workflow ID already guard against this: `createNexusAttachCallbacksAction` sets `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING` (`throughput_stress.go:915`), and the signal-with-start handler path hardcodes `USE_EXISTING` (`kitchen_sink.go:686`).

</details>

**Suggestion:**

```go
					StartOptions: &NexusWorkflowStartOptions{
						WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
						WorkflowInput: &WorkflowInput{ /* ... */ },
					},
```

scenarios/throughput_stress.go
```diff
@@ -616,6 +649,12 @@ func (t *tpsExecutor) createActionsChunk(
 					)
 				}
 			}
+			if t.config.IncludeNexusSignal || t.config.IncludeNexusSignalWithStart || t.config.IncludeNexusUpdate {
+				nexusWorkflowID := fmt.Sprintf("%s-nexus-target-%d",
+					run.DefaultStartWorkflowOptions().ID,
+					t.internalIterationIndex(run, remainingInternalIters, i))
+				syncActions = append(syncActions, t.createNexusWorkflowTargetSequence(nexusWorkflowID, rng))
```

<details>
<summary><strong>small</strong> — The sequence's correctness depends on being in <code>syncActions</code>, but nothing records that.</summary>

`await_pending_actions` drains a workflow-global list (`ws.pendingActions`, `kitchen_sink.go:500-508`), so two sequences using `wait_started` must not overlap. Sitting in `syncActions` guarantees this sequence finishes before the concurrent `asyncActions` set (which contains `createNexusAttachCallbacksAction`, also `wait_started` + `await_pending_actions`) starts. The first revision of this PR appended it to `asyncActions`, where every other Nexus action lives — moving it back there would let the two `await_pending_actions` calls swallow each other's futures, so the attach-callbacks assertion would end up awaiting this target's start operation.

</details>

**Suggestion:** Add a one-liner so the placement survives future edits, e.g. `// Kept in the sequential phase: await_pending_actions drains a workflow-global list, so this must not overlap with the wait_started ops in asyncActions.`

scenarios/throughput_stress.go
```diff
@@ -248,6 +263,21 @@ func (t *tpsExecutor) Configure(info loadgen.ScenarioInfo) error {
 	if config.IncludeNexusStandaloneActivity && !config.NexusEnabled {
 		return fmt.Errorf("%s requires %s", IncludeNexusStandaloneActivityFlag, NexusEnabledFlag)
 	}
+	config.IncludeNexusSignal = info.OptionBool(IncludeNexusSignalFlag)
+	config.IncludeNexusSignalWithStart = info.OptionBool(IncludeNexusSignalWithStartFlag)
+	config.IncludeNexusUpdate = info.OptionBool(IncludeNexusUpdateFlag)
+	for _, nexusWorkflowAction := range []struct {
```

<details>
<summary><strong>small</strong> — Twelve-line table loop to do what the surrounding code does in three <code>if</code>s.</summary>

The values are already assigned above the loop, so the anonymous struct slice only carries flag-name/bool pairs for validation. `Configure` validates every other dependent option with a plain `if` (e.g. line 263 right above), and matching that reads better than a table with three static rows.

</details>

**Suggestion:**

```go
	if config.IncludeNexusSignal && !config.NexusEnabled {
		return fmt.Errorf("%s requires %s", IncludeNexusSignalFlag, NexusEnabledFlag)
	}
	if config.IncludeNexusSignalWithStart && !config.NexusEnabled {
		return fmt.Errorf("%s requires %s", IncludeNexusSignalWithStartFlag, NexusEnabledFlag)
	}
	if config.IncludeNexusUpdate && !config.NexusEnabled {
		return fmt.Errorf("%s requires %s", IncludeNexusUpdateFlag, NexusEnabledFlag)
	}
```

scenarios/throughput_stress.go
```diff
@@ -930,6 +969,104 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
 	})
 }
 
+func (t *tpsExecutor) createNexusWorkflowTargetSequence(workflowID string, rng *rand.Rand) *Action {
+	var startAction *Action
+	var targetActions []*Action
+	if t.config.IncludeNexusSignalWithStart && rng.Intn(2) == 0 {
```

<details>
<summary><strong>nit</strong> — <code>IncludeNexusSignalWithStart</code> is tested twice to place the same action in one of two slots.</summary>

The `else if` repeating the flag hides that the only real choice is *where* the signal-with-start action goes.

</details>

**Suggestion:**

```go
	if t.config.IncludeNexusSignalWithStart {
		signalWithStart := t.createNexusSignalWithStartAction(workflowID)
		if rng.Intn(2) == 0 {
			startAction = signalWithStart // creates the target
		} else {
			targetActions = append(targetActions, signalWithStart)
		}
	}
```

scenarios/throughput_stress.go
```diff
@@ -109,16 +118,20 @@ type tpsConfig struct {
...
-	isResuming bool
-	runID      string
-	rng        *rand.Rand
+	lock             sync.Mutex
+	state            *tpsState
+	config           *tpsConfig
+	isResuming       bool
+	runID            string
+	rng              *rand.Rand
+	onActionsCreated func(*loadgen.Run, []*ActionSet)
```
stephanos: New addition to be able to observe generated actions.

scenarios/throughput_stress.go
```diff
@@ -364,7 +398,9 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 				//
 				// NOTE: No client actions (e.g. Signal) are defined; however, client action activities are.
 				// That means these client actions are sent from the activity worker instead of Omes.
-				options.Params.WorkflowInput.InitialActions = t.createActions(run)
+				actions := t.createActions(run)
+				options.Params.WorkflowInput.InitialActions = actions
+				t.onActionsCreated(run, actions)
```
stephanos: Calling the new callback.

scenarios/throughput_stress.go
```diff
@@ -930,6 +975,123 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
 	})
 }
 
+// createNexusWorkflowActionSequence starts a workflow, sends the configured actions, and waits for completion.
+func (t *tpsExecutor) createNexusWorkflowActionSequence(workflowID string, rng *rand.Rand) *Action {
+	var startAction *Action
+	var targetActions []*Action
+	if t.config.IncludeNexusSignalWithStart {
+		signalWithStart := t.createNexusSignalWithStartAction(workflowID)
+		if rng.Intn(2) == 0 {
```
stephanos: I wanted to keep the combinations/complexity at bay, so I figured doing a 50/50 here is a good approach.

scenarios/throughput_stress_test.go
```diff
@@ -147,11 +147,9 @@ func TestThroughputStressFeatureAutoEnable(t *testing.T) {
 	require.Equal(t, 1, executor.Snapshot().(tpsState).CompletedIterations)
 }
 
-func TestThroughputStressNexusStandaloneActivity(t *testing.T) {
+func TestThroughputStressNexusStandaloneActivityActions(t *testing.T) {
```
stephanos, line 152: Updated this, too, now to use the new callback (it wasn't actually running the scenario before which I think is not ideal; it should do both: run it and verify its actions IMO). cc @tekkaya (just FYI)

scenarios/throughput_stress.go
```diff
@@ -109,16 +118,20 @@ type tpsConfig struct {
...
-	isResuming bool
-	runID      string
-	rng        *rand.Rand
+	lock             sync.Mutex
+	state            *tpsState
+	config           *tpsConfig
+	isResuming       bool
+	runID            string
+	rng              *rand.Rand
+	onActionsCreated func([]*ActionSet)
```
stephanos, line 127: New addition to be able to observe generated actions.

scenarios/throughput_stress.go
```diff
@@ -67,11 +67,21 @@ const (
 	// Opt-in and off by default (only the Go worker implements the operation); requires Nexus load
 	// (nexus-enabled) and server support for standalone activities + activity completion callbacks.
 	IncludeNexusStandaloneActivityFlag = "include-nexus-standalone-activity"
+	// IncludeNexusWorkflowActionsFlag enables Nexus operations that signal and update a workflow.
+	// Opt-in and off by default; requires Nexus load (nexus-enabled).
+	IncludeNexusWorkflowActionsFlag = "include-nexus-workflow-actions"
```
stephanos, line 72: At first I had 3 flags, but I think 1 is perfectly fine.

scenarios/throughput_stress.go
```diff
@@ -932,6 +964,93 @@ func (t *tpsExecutor) createNexusStandaloneActivityAction() *Action {
 	})
 }
 
+// createNexusWorkflowActionSequence starts a workflow that waits for the configured actions before completing.
+func (t *tpsExecutor) createNexusWorkflowActionSequence(workflowID string) *Action {
+	targetWorkflowActions := []*Action{
+		NewAwaitWorkflowStateAction(nexusSignalWithStartStateKey, nexusActionComplete),
+		NewAwaitWorkflowStateAction(nexusSignalStateKey, nexusActionComplete),
```

<details>
<summary><strong>high</strong> — Three-key state gate deadlocks the target workflow whenever two of the messages land in the same workflow task, because `SetWorkflowState` replaces the whole KV map.</summary>

`workers/go/workerlib/kitchensink/kitchen_sink.go:315` implements the action as `ws.workflowState = setWfState` — a full replacement, not a merge. Each of the three actions here writes a single-key map, so at most one of `nexus-signal-with-start` / `nexus-signal` / `nexus-update` is ever present at a time.

That only works if the target processes each message in a *separate* workflow task, so the main coroutine can observe each key before the next write clobbers it. Nothing guarantees that:

- The caller only waits for the Nexus signal operation to complete, which means `SignalWorkflow` returned — not that the target has processed the signal. It then schedules the update. Under a backlogged task queue (the normal state in a throughput-stress run) the target's next WFT can carry both the signal event and the update.
- Within one WFT the writes are not even ordered by history order: the signal actions run in a coroutine spawned from the signal-receive loop (`workflow.Go` at kitchen_sink.go:117), while the update handler runs in its own coroutine. The dispatcher runs newly created coroutines in the same pass, so the signal's write can land *after* the update's.

Either interleaving hangs the target:

- update writes last → main is still on `Await(nexus-signal)`, key gone, blocked forever.
- signal writes last → main clears `Await(nexus-signal)` on the next pass, then blocks on `Await(nexus-update)`, which will never be written again.

The caller does not notice: the update operation completes normally (the handler ran), so the iteration passes while the target workflow hangs until its 60-minute `WorkflowExecutionTimeout`. At stress-level concurrency that leaks a growing pool of stuck workflows, and the "target completes after every action" invariant silently stops holding.

</details>

**Suggestion:** Gate the target on a key that exactly one message writes — the last one — and let the Nexus operation results (which already assert success and `ExpectedOutput`) be the proof that the signal and signal-with-start were delivered:

```go
	targetWorkflowActions := []*Action{
		// Only the update writes state: SetWorkflowState replaces the whole KV map, so a
		// multi-key gate breaks as soon as two messages land in the same workflow task.
		NewAwaitWorkflowStateAction(nexusUpdateStateKey, nexusActionComplete),
		// Yield a workflow task so update completion is recorded before the target closes.
		NewTimerAction(time.Millisecond),
		NewEmptyReturnResultAction(),
	}
```

with the two signal actions carrying a harmless action set (e.g. `SingleActionSet(NewTimerAction(time.Millisecond))`) instead of `NewSetWorkflowStateAction(...)`. If you would rather keep the three-key gate, `SetWorkflowState` needs to merge into `ws.workflowState.Kvs` in every worker, which is a broader change.

scenarios/throughput_stress.go
```diff
@@ -616,18 +640,22 @@ func (t *tpsExecutor) createActionsChunk(
 					)
 				}
 			}
+			if t.config.IncludeNexusWorkflowActions {
+				nexusWorkflowID := fmt.Sprintf("%s/nexus-workflow-%d", run.DefaultStartWorkflowOptions().ID, iterationIndex)
+				// Keep this sequence sequential so signal-with-start creates the target before the remaining actions.
+				syncActions = append(syncActions, t.createNexusWorkflowActionSequence(nexusWorkflowID))
```

<details>
<summary><strong>med</strong> — Appending the sequence to <code>syncActions</code> serializes three Nexus round-trips ahead of the whole iteration, and is not needed for the ordering the comment cites.</summary>

`createNexusWorkflowActionSequence` already returns a nested `ActionSet` with `Concurrent` unset, and `handleActionSet` runs a non-concurrent nested set strictly in order even when its parent set is concurrent (`workers/go/workerlib/kitchensink/kitchen_sink.go:198-205`, `:339`). So signal-with-start is guaranteed to precede the signal and update regardless of which bucket the nested set goes in.

Putting it in `syncActions` instead means every internal iteration blocks on signal-with-start → signal → update (each a caller WFT plus a Nexus handler round-trip, and the update additionally waits for an async completion callback) *before* any of the concurrent load starts. In a scenario whose purpose is measuring throughput, that adds serial latency per iteration and reduces concurrency, and it is inconsistent with every other Nexus action here (`createNexusEchoSyncAction`, `createNexusStartWorkflowAction`, `createNexusAttachCallbacksAction`, the standalone ones) which all go to `asyncActions`.

</details>

**Suggestion:**

```go
			if t.config.IncludeNexusWorkflowActions {
				nexusWorkflowID := fmt.Sprintf("%s/nexus-workflow-%d", run.DefaultStartWorkflowOptions().ID, iterationIndex)
				// The nested set is non-concurrent, so signal-with-start still creates the
				// target before the remaining actions run.
				asyncActions = append(asyncActions, t.createNexusWorkflowActionSequence(nexusWorkflowID))
			}
```

scenarios/throughput_stress_test.go
```diff
@@ -160,93 +159,128 @@ func TestThroughputStressNexusStandaloneActivity(t *testing.T) {
...
-				for _, set := range seq.GetActionSets() {
-					for _, ca := range set.GetActions() {
-						if sn := ca.GetDoStandaloneNexusOperation().GetOperation(); sn.GetInput().GetStartActivity() != nil {
-							standalone = true
-						}
+	type actionCounts struct {
+		starts           int
+		signals          int
+		signalWithStarts int
+		updates          int
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — <code>onActionsCreated</code> is invoked from the per-iteration goroutines, so the callbacks race on <code>counts</code> and call <code>require</code> off the test goroutine.</summary>

`GenericExecutor` runs each iteration in its own goroutine (`loadgen/generic_executor.go:132`), and the hook fires from `UpdateWorkflowOptions` inside that goroutine. Two consequences for both new callbacks (this one and the one at line 171):

- `counts` is mutated without synchronization. It happens to be safe today only because both tests use `Iterations: 1`; bumping iterations — or anyone copying this pattern — makes it a `-race` failure.
- `require.NotNil` / `require.Equal` here call `t.FailNow()` from a non-test goroutine, which testify explicitly documents as unsupported: it calls `runtime.Goexit()` on the wrong goroutine, so the test either hangs or reports a confusing failure instead of failing cleanly.

</details>

**Suggestion:** Guard `counts` with a `sync.Mutex` (read it back under the lock before the final assert) and switch the in-callback checks to `assert.NotNil` / `assert.Equal`, which record the failure and let the test goroutine fail at the end. Documenting the contract on the field itself would help too:

```go
	// onActionsCreated observes generated actions; called concurrently from iteration goroutines.
	onActionsCreated func([]*ActionSet)
```

docs/throughput-stress.md
```diff
@@ -90,15 +90,18 @@ is any.
...
-This is **opt-in and off by default**; pass `--option include-nexus-standalone-activity=true`.
-It requires `nexus-enabled` and also needs server support for standalone activities and activity
-completion callbacks (dynamic config `activity.enableStandalone` and `activity.enableCallbacks`) and a
-Nexus callback URL; if those are off the operation fails clearly rather than being skipped.
+- `include-nexus-standalone-activity`
+- `include-nexus-workflow-actions`
 
-Currently only supported and run by Go workers.
+The workflow actions start their target workflow with signal-with-start, then send an ordinary
+signal and an update to that target.
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — The new section documents the server requirements for the standalone-activity option but not for <code>include-nexus-workflow-actions</code>, which also needs opt-in dynamic config.</summary>

`TestThroughputStressNexusWorkflowActions` has to start the dev server with `history.enableChasm`, `history.enableCHASMCallbacks`, `history.enableCHASMSignalBacklinks` and `history.enableUpdateCallbacks` for this path to work — the Nexus update operation goes async at `WorkflowUpdateStageAccepted` and relies on the update completion callback. None of that is mentioned here, while the paragraph right below carefully spells out `activity.enableStandalone` / `activity.enableCallbacks` for the sibling option.

Worth documenting: unlike the standalone-activity operation, the update path has no fail-fast error mapping in the handler, so on a server without update callbacks the operation fails in a way that is not obviously a configuration problem.

</details>

**Suggestion:**

```markdown
The workflow actions start their target workflow with signal-with-start, then send an ordinary
signal and an update to that target. The update completes asynchronously through a completion
callback, so this needs server support for CHASM callbacks and update callbacks (dynamic config
`history.enableChasm`, `history.enableCHASMCallbacks`, `history.enableCHASMSignalBacklinks` and
`history.enableUpdateCallbacks`).
```

scenarios/throughput_stress_test.go
```diff
@@ -160,93 +159,128 @@ func TestThroughputStressNexusStandaloneActivity(t *testing.T) {
...
 	}
-	for _, set := range exec.createActions(run) {
-		walk(set.GetActions())
+	scenarioInfo := loadgen.ScenarioInfo{
+		RunID:         "nexus-workflow-actions-signal-with-start",
+		Configuration: loadgen.RunConfiguration{Iterations: 1},
+		Options: loadgen.MustResolveScenarioOptions("throughput_stress", map[string]string{
+			IterFlag:                          "1",
+			NexusEnabledFlag:                  "true",
+			"include-nexus-workflow-actions":  "true",
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — The new tests hardcode flag and state-key strings that exist as constants in the same package.</summary>

`IncludeNexusWorkflowActionsFlag`, `nexusSignalWithStartStateKey`, `nexusSignalStateKey`, `nexusUpdateStateKey` and `nexusActionComplete` are all reachable here, and the surrounding tests use the flag constants (`IterFlag`, `NexusEnabledFlag`, `IncludeStandaloneNexusFlag`). Same for `const flag = "include-nexus-workflow-actions"` at line 401. As written, renaming a state key leaves the assertion at lines 256-258 passing against stale strings only until the counts drift.

</details>

**Suggestion:**

```go
			IncludeNexusWorkflowActionsFlag:   "true",
```

and in the sequence assertion:

```go
						ks.NewAwaitWorkflowStateAction(nexusSignalWithStartStateKey, nexusActionComplete),
						ks.NewAwaitWorkflowStateAction(nexusSignalStateKey, nexusActionComplete),
						ks.NewAwaitWorkflowStateAction(nexusUpdateStateKey, nexusActionComplete),
```

scenarios/throughput_stress.go
```diff
@@ -364,7 +379,9 @@ func (t *tpsExecutor) Run(ctx context.Context, info loadgen.ScenarioInfo) error
 				//
 				// NOTE: No client actions (e.g. Signal) are defined; however, client action activities are.
 				// That means these client actions are sent from the activity worker instead of Omes.
-				options.Params.WorkflowInput.InitialActions = t.createActions(run)
+				actions := t.createActions(run)
+				options.Params.WorkflowInput.InitialActions = actions
+				t.onActionsCreated(actions)
```
THardy98, line 384: is this only populated on scenario start ?

THardy98 (approved):
Nice! left a clarifying question

scenarios/throughput_stress.go
stephanos: It's inside of `UpdateWorkflowOptions`, which is called for each iteration. The tests only have 1 iteration; but if we had 2, we'd see it being called twice.

`onActionsCreated` defaults to an empty function; only some of the tests override it.

PR #468, opened 2026-09-05
Stop pinning the initial autoscaling poller count to the max in omes worker

veeral-patel:
## What was changed

`--worker-{activity,workflow}-poller-autoscale-max=N` set `InitialNumberOfPollers` as well as `MaximumNumberOfPollers`, so the flag quietly fixed the *starting* poller count at the ceiling. Dropping the two assignments lets the SDK apply its own initial default.

```diff
  options.ActivityTaskPollerBehavior = sdkworker.NewPollerBehaviorAutoscaling(sdkworker.PollerBehaviorAutoscalingOptions{
-     InitialNumberOfPollers: args.ActivityPollerAutoscaleMax,
      MaximumNumberOfPollers: args.ActivityPollerAutoscaleMax,
  })
```

and the same in the `workflow-poller-autoscale-max` block.

Previous code was a workaround to a SDK bug that's been since fixed.
* https://github.com/temporalio/omes/pull/253
* https://github.com/temporalio/sdk-go/pull/2105

### Will this cause a inconsistency with the omes workers in other languages?

No - I checked the other workers: Go is the only one coupling initial to max.

| Harness | Call | Sets initial? |
|---|---|---|
| Python | `PollerBehaviorAutoscaling(maximum=…)` (`worker.py:233`) | no |
| TypeScript | `{type: 'autoscaling', maximum: …}` (`worker.ts:250`) | no |
| Java | `new PollerBehaviorAutoscaling(null, max, null)` (`WorkerHarness.java:149`) | no — positional `(min, max, initial)` |
| .NET | `new PollerBehavior.Autoscaling(maximum: …)` (`Worker.cs:346`) | no |
| Ruby | `PollerBehavior::Autoscaling.new(maximum: …)` (`worker.rb:129`) | no |
| **Go (before)** | `{InitialNumberOfPollers: max, MaximumNumberOfPollers: max}` | **yes** |

## ⚠️ Behavior change

Below `-max=5` you get `initial=5, max=3`: the SDK doesn't validate `initial <= max` and stores the initial target unclamped (`internal_worker_base.go:956`), so the worker transiently exceeds its own maximum. Verified not to error.

Nothing in this repo passes these flags. External command lines might be affected, if they are relying on initial being set to max. However, this should be rare as this is unexpected and incorrect behavior.

## How was this tested

- Build, vet and tests pass


docs/running.md
```diff
@@ -111,6 +111,10 @@ scenario (its action tree / executor), not by combining scenarios on the command
   - `resource-based-default` — the SDK resource-based worker tuner.
   - `throughput-stress-baseline` — a fixed config for `throughput_stress` runs (workflow cache 50, 8
     workflow-task slots, 32 activity/local-activity slots, 2 workflow-task pollers, 4 activity pollers).
+- `--worker-max-concurrent-{activity,workflow}-pollers` pin a fixed poller count.
```
rkannan82: Prefer to explicitly call out the flags instead of shortcut like -{activity,workflow}

docs/running.md
```diff
@@ -111,6 +111,10 @@ scenario (its action tree / executor), not by combining scenarios on the command
...
+- `--worker-{activity,workflow}-poller-autoscale-{min,max,initial}` autoscale it instead. Any one of the
+  three enables autoscaling; unset values use the SDK's own defaults. `-min` and `-initial` are
```
rkannan82: - I would suggest introducing a new bool flag instead - enable-poller-auto-scaling
- To avoid breaking existing usage, we continue to honor the max flag. We don't need to document this here. Just put a comment in the worker.go.

rkannan82 (approved):

THardy98 (approved):

PR #469, opened 2026-09-07
Forward kitchen sink signal arguments

stephanos:
Forward the existing `SendSignalAction.args` payloads from the Go, Java and Python kitchen sink workers instead of always delivering an empty signal.


stephanos:
Closing because the Nexus tests will make the Nexus workflow action itself complete the target, removing this PR from the stack.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -389,8 +389,8 @@ message SendSignalAction {
   string run_id = 2;
   // Name of the signal handler
   string signal_name = 3;
-  // Arguments for the handler
-  repeated temporal.api.common.v1.Payload args = 4;
+  // Argument for the handler
+  temporal.api.common.v1.Payload arg = 4;
```
stephanos, line 391: We only need one; a payload can contain lists inside if needed. Simpler.

THardy98: nit: `Argument for the handler, uses lists to represent multi-arg input`. 

No strong feelings either way if you feel this is redundant.

THardy98 (commented):
I don't see a corresponding change for the Typescript worker

stephanos:
> I don't see a corresponding change for the Typescript worker

~Oh good catch; adding that now.~ Actually, TS doesn't support this action at all! This PR just fixes the existing 3 that do.

PR #474, opened 2026-09-07
Fix TypeScript protobuf resolution

stephanos:
Bump `github.com/temporalio/features` and TS SDK.

Alternatives considered:
>  - SDK 1.23 without the direct, pinned dependency: fails with "type must be a Type".
>  - SDK 1.22: fails because features overrides protobufjs to 8.x, while SDK 1.22 was built around the older protobuf stack.


internal/workerctl/build.go
```diff
@@ -231,6 +231,7 @@ func (b *Builder) buildTypeScript(ctx context.Context, baseDir string) (sdkbuild
 		},
 		MoreDependencies: map[string]string{
 			"@grpc/proto-loader": "^0.8.0",
+			"protobufjs":         "^8.7.1",
```
stephanos: Fails without this.

stephanos:
Closed in favor of https://github.com/temporalio/omes/pull/475

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.11.1
-	github.com/temporalio/features v0.0.0-20260427223549-86e4c0deedd7
+	github.com/temporalio/features v0.0.0-20260908141410-6f64678c5b39
```
stephanos, line 17: ie latest main

internal/workerctl/build.go
stephanos: > The direct dependency is necessary because generated json-module.js performs require("protobufjs/light"). With pnpm, a transitive SDK dependency is not reliably visible there. Without a direct declaration, Node either cannot resolve it or finds the worker fixture’s protobufjs 7.x copy, producing incompatible runtime type identities.

THardy98 (approved):

mjameswh (approved):

mjameswh:
@stephanos You are missing dependencies in "workers/typescript/package.json". Should be something like this:

```
  "dependencies": {
    ...
    "protobufjs": "^8.7.1",
    "protobufjs-cli": "^2.6.1",
    ...
  },
```

workers/typescript/package.json
```diff
@@ -35,6 +35,8 @@
     "@temporalio/workflow": "^1.23.0",
     "commander": "^11.1.0",
     "long": "^5.2.3",
+    "protobufjs": "^8.8.0",
```
mjameswh, line 38: Makes sense.

workers/typescript/apps/worker/app.ts
```diff
@@ -39,6 +40,15 @@ async function buildWorker(client: Client, context: WorkerContext): Promise<Work
     activities: createActivities(client, context.errOnUnimplemented),
     taskQueue: context.taskQueue,
     ...context.workerOptions,
+    bundlerOptions: {
+      webpackConfigHook(config) {
+        config.resolve!.alias = {
+          ...config.resolve!.alias,
+          protobufjs: dirname(require.resolve('protobufjs/package.json')),
```
mjameswh, line 47: This change is unexpected (i.e. you should not need that), and quite surprising (i.e. even assuming you need it, it should not point to a package.json file).

stephanos: Yes, it's odd eh?

I'll leave this claim from my agent here for posterity:

> Without the app.ts alias, source-SDK builds can bundle two physical protobufjs 8.8.0 copies:
> 
>   - Generated kitchen-sink protobuf types use the copy under prepared/node_modules.
>   - The SDK payload converter uses the copy resolved through @temporalio/common.
> 
>   Even though both are version 8.8.0, their Type constructors have different identities. The SDK’s instanceof check then throws:
> 
>   TypeError: type must be a Type
> 
>   Workflow activations repeatedly fail, so the five-noop smoke test appears stuck indefinitely. Released SDK layouts may pass due to package-manager hoisting, but the source-build CI layout
>   does not reliably deduplicate them. The build.go entry installs the dependency; the app.ts hook enforces one physical copy in the bundle.

PR #478, opened 2026-09-08
Add CI job timeouts

stephanos:
Ensure jobs don't spin forever.


THardy98 (approved):
would be ok if these were even shorter (i.e. 15-20mins)

stephanos:
👍 ~changing it to 20min; I think 15min is a bit too tight~ changed it to +10min of what `main` runtimes are on case by case basis.

PR #483, opened 2026-09-10
Preserve Nexus operation payloads with RawValue

stephanos:
## Summary

The fix establishes one consistent rule: kitchen-sink Nexus results are opaque payloads and cross every result boundary as `RawValue`.

## Why

The kitchen-sink APIs use `common.Payload` as an already-encoded, opaque value. Passing that value through the SDK as an ordinary `*common.Payload` causes the data converter to serialize it again, ie adding another payload wrapper/envelope.

Nexus exposed this because synchronous and asynchronous completions take different routes:
  - A synchronous result is returned directly by the Nexus handler.
  - An asynchronous workflow/update result is first persisted as the workflow or update result and later forwarded through a completion callback.

Using `RawValue` only in the Nexus caller was insufficient. It stopped decoding the outer envelope, but producers such as `KitchenSinkWorkflow` etc were still creating that envelope.

`RawValue` is the SDK abstraction specifically intended for forwarding an already-encoded payload, so this avoids custom converter special cases and makes results independent of whether the operation completed synchronously or asynchronously.


loadgen/kitchen_sink_executor_test.go
```diff
@@ -1009,14 +1010,15 @@ func TestKitchenSink(t *testing.T) {
...
-													NewEmptyReturnResultAction(),
+													NewReturnResultAction(ConvertToPayload("nexus-workflow-result")),
 												),
 											},
 										},
 										Action: &NexusWorkflowAction_Start{Start: &emptypb.Empty{}},
 									},
 								},
 							},
+							ExpectedOutput: ConvertToPayload("nexus-workflow-result"),
```
stephanos: Regression test 1/2

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1262,13 +1264,73 @@ func TestKitchenSink(t *testing.T) {
...
+					},
+					ExpectedOutput: ConvertToPayload("nexus-sync-update-target"),
+				}),
+			)}},
+			historyMatcher: PartialHistoryMatcher(`
+				NexusOperationStarted {"links":[{"workflowEvent":{"workflowId":"nexus-sync-update-target","requestIdRef":{"eventType":"EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED"}}}]}
+				NexusOperationCompleted
+				NexusOperationCompleted {"links":[{"workflowEvent":{"workflowId":"nexus-sync-update-target","requestIdRef":{"eventType":"EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED"}}}]}`),
+			expectedUnsupportedErrs: nexusWorkflowActionUnsupportedSDKs,
+		},
```
stephanos, line 1340: This is the main regression we're fixing.

The other ones I added here are also valuable, ofc. They increase confidence in the overall change here.

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -47,7 +48,7 @@ type KSWorkflowState struct {
 	pendingActions []workflow.Future
 }
 
-func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (*common.Payload, error) {
+func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput) (converter.RawValue, error) {
```

<details>
<summary><strong>med</strong> — The new "results cross every boundary as RawValue" rule is only implemented in the Go and Python workers; the other four SDK workers still double-encode.</summary>

`ReturnResultAction.return_this` is an already-encoded `common.Payload` in every worker, but only Go and Python now pass it through as a raw value. The remaining workers still hand the `Payload` proto straight to the data converter, which re-encodes it:

- `workers/java/io/temporal/omes/workerlib/kitchensink/KitchenSinkWorkflowImpl.java:152` — `return result.getReturnThis();`
- `workers/typescript/workerlib/kitchensink/workflows/kitchen_sink.ts:131` — `return action.returnResult.returnThis;`
- `workers/ruby/workerlib/kitchensink/kitchen_sink.rb:107` — `return action.return_result.return_this`
- .NET equivalent

Nothing breaks today because `nexusUnsupportedSDKs` excludes those four from every Nexus test and no test asserts on a plain workflow result. But the kitchen-sink workflow result encoding is now SDK-dependent: the same `TestInput` produces a `json/plain` payload on Go/Python and a `json/protobuf`-wrapped `Payload` on Java/TS/Ruby/.NET. Whoever adds `ExecuteNexusOperation` support to those SDKs next will hit exactly the bug this PR fixes, and the double-wrapping comments that were just deleted are no longer there to explain it.

</details>

**Suggestion:** Either port the same `RawValue`/`Payload`-passthrough handling to the Java, TypeScript, Ruby, and .NET workers in this PR, or leave a short note on `ReturnResultAction` in `workers/proto/kitchen_sink/kitchen_sink.proto` stating that `return_this` must be returned as an opaque payload (Go `converter.RawValue`, Python `RawValue`, etc.) and that only Go/Python currently comply.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1262,13 +1264,73 @@ func TestKitchenSink(t *testing.T) {
...
+							Action: &NexusWorkflowAction_Update{Update: &DoUpdate{
+								Variant: &DoUpdate_DoActions{DoActions: &DoActionsUpdate{
+									Variant: &DoActionsUpdate_DoActions{DoActions: SingleActionSet()},
+								}},
+								UpdateId: "nexus-sync-update",
+							}},
+						}},
+					},
+					ExpectedOutput: ConvertToPayload("nexus-sync-update-target"),
+				}),
```

<details>
<summary><strong>small</strong> — `NexusOperation/Sync/Update` never awaits the `WaitStarted` operation, so a failure of the first (start-workflow) operation is silently swallowed.</summary>

The first operation uses `AwaitableChoice_WaitStarted`, which in `withAwaitableChoiceCustom` appends the future to `ws.pendingActions` instead of waiting for it (`workers/go/workerlib/kitchensink/kitchen_sink.go:452`). Unlike `NexusOperation/Sync/Signal`, `NexusOperation/Sync/SignalWithStart` and `NexusOperation/Async/Update`, this test set has no trailing `AwaitPendingActions` action, so that future is dropped: if `nexus-sync-update-target` fails or its Nexus operation errors after start, the caller workflow completes green and the test still passes because `PartialHistoryMatcher` only requires a subsequence.

</details>

**Suggestion:**

```suggestion
				}),
				&Action{Variant: &Action_AwaitPendingActions{AwaitPendingActions: &AwaitPendingActions{}}},
```

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -184,11 +185,11 @@ func KitchenSinkWorkflow(ctx workflow.Context, params *kitchensink.WorkflowInput
 			if !ok {
 				err = fmt.Errorf("timeout waiting for all signals before deadline, missing signals: %v, err: %w", missingSignals, err)
 			}
-			return nil, fmt.Errorf("failed waiting for signals, missing signals: %v, err: %w", missingSignals, err)
+			return converter.RawValue{}, fmt.Errorf("failed waiting for signals, missing signals: %v, err: %w", missingSignals, err)
 		}
 	}
 
-	return retOrErr.retme, nil
+	return converter.NewRawValue(retOrErr.retme), nil
```

<details>
<summary><strong>small</strong> — With RawValue passthrough, <code>NewEmptyReturnResultAction</code> now writes a payload with no <code>encoding</code> metadata into history.</summary>

`NewEmptyReturnResultAction()` is `NewReturnResultAction(&common.Payload{})` (`loadgen/kitchensink/helpers.go:177`), i.e. nil metadata and nil data. Previously the data converter encoded that proto, so the workflow result was always a well-formed payload. Now it is forwarded verbatim, so `nexus-signal-target`, `nexus-sws-target` and `nexus-sync-update-target` all complete with `Payload{}`.

Anything that decodes it into a concrete type (rather than `RawValue`) gets "no encoding metadata"; the UI, `temporal workflow show`, and payload codecs all treat metadata-less payloads as undecodable. It also means the async-completion result those Nexus operations deliver is a metadata-less payload, so an `ExpectedOutput` on one of them would have to be spelled `&common.Payload{}`.

</details>

**Suggestion:** Make the empty case a real encoded payload, e.g. `func NewEmptyReturnResultAction() *Action { return NewReturnResultAction(ConvertToPayload(nil)) }`, which yields a `binary/null` payload that round-trips through `RawValue` and stays readable in history.

loadgen/kitchensink/client_action_executor.go
```diff
@@ -218,12 +218,12 @@ func (e *ClientActionsExecutor) executeStandaloneNexusOperation(ctx context.Cont
...
 
 	if expectedOutput := nexusOp.GetExpectedOutput(); expectedOutput != nil {
-		var result commonpb.Payload
+		var result converter.RawValue
 		if err := handle.Get(ctx, &result); err != nil {
 			return fmt.Errorf("get standalone Nexus operation: %w", err)
 		}
-		if !expectedOutput.Equal(&result) {
-			return fmt.Errorf("expected standalone Nexus operation output %v, got %v", expectedOutput, &result)
+		if !expectedOutput.Equal(result.Payload()) {
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — The expected-output check is duplicated verbatim between the client executor and the workflow worker.</summary>

This block is now byte-for-byte the same logic as `workers/go/workerlib/kitchensink/kitchen_sink.go:475-482` (decode into `converter.RawValue`, compare `expectedOutput.Equal(result.Payload())`, format the same error). Both sites had to be edited in this PR for the same reason, and they will have to be edited together again if the comparison ever needs to tolerate, say, differing metadata.

</details>

**Suggestion:** Extract a helper next to `ConvertToPayload` in `loadgen/kitchensink/helpers.go`, e.g. `func CheckExpectedOutput(expected *common.Payload, actual converter.RawValue) error`, and call it from both places.

workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -16,6 +16,8 @@ message TestInput {
   WorkflowInput workflow_input = 1;
   ClientSequence client_sequence = 2;
   WithStartClientAction with_start_action = 3;
+  // If set, the workflow's output is compared against this value after completion.
+  temporal.api.common.v1.Payload expected_output = 4;
```
stephanos: Added this to be able to test the results from workflows (useful for our regression tests). This mirrors the same pattern we already have in `ExecuteNexusOperation`).

workers/typescript/workerlib/kitchensink/payload-converter.ts
```diff
@@ -1,4 +1,4 @@
-import root, { temporal } from './protos/root';
+import root from './protos/root';
```
stephanos: 🤖

<details>
<summary><strong>high</strong> — Removing <code>PassThroughPayload</code> makes the TypeScript worker re-encode raw <code>Payload</code> <em>arguments</em> (child-workflow input, continue-as-new args, custom handler args).</summary>

`PassThroughPayload` was the only converter that let an already-encoded `temporal.api.common.v1.Payload` reach the server unchanged. This PR migrated the *result* paths to `RawValue`, but the *argument* paths still hand bare `Payload` messages to the SDK:

- `workflows/kitchen_sink.ts:152` — `startChild(..., { args: execChild.input ?? [] })`
- `workflows/kitchen_sink.ts:136` — `continueAsNew(action.continueAsNew.arguments![0])`
- `client-action-executor.ts:102/138/196` — `signal.custom.args`, `update.custom.args`, `query.custom.args`

`protos/root.js` is a `patchProtobufRoot`-wrapped reflection root, so those decoded `Payload` values carry `$type`. With the passthrough gone they now fall through to `ProtobufJsonPayloadConverter(root)`, which wraps them in a second `json/protobuf` envelope (`messageType=temporal.api.common.v1.Payload`) — exactly the double-encoding this PR removes elsewhere.

Failure scenario: `ExecActivity/ExecChildWorkflow` (`loadgen/kitchen_sink_executor_test.go:253`) runs on TypeScript — it is absent from every `*UnsupportedSDKs` map — and passes `ConvertToPayload(&WorkflowInput{InitialActions: ...})` as child input. The child would decode a `Payload` instead of a `WorkflowInput`, find no initial actions, and block waiting for signals until the run times out, so `ChildWorkflowExecutionCompleted` never appears.

The other SDKs already wrap these argument sites: Python `RawValue(i)`, Ruby `Temporalio::Converters::RawValue.new(a)`, .NET `new RawValue(p)`. TypeScript is the only one that was relying on the converter instead.

</details>

**Suggestion:** Wrap the argument sites the same way the return sites now are, e.g. `args: (execChild.input ?? []).map((p) => RawValue.fromPayload(Payload.create(p)))` and `continueAsNew(RawValue.fromPayload(Payload.create(action.continueAsNew.arguments![0])))`, or keep `PassThroughPayload` in the composite until those call sites are migrated.

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1509,7 +1593,9 @@ func testSupportedFeature(
 	}
 
 	// Check if workflow failure is expected
-	if tc.expectedWorkflowError != "" {
+	if tc.expectedRunError != "" {
+		require.ErrorContains(t, execErr, tc.expectedRunError, "run should fail")
```

<details>
<summary><strong>small</strong> — <code>expectedRunError</code> is matched case-sensitively while the sibling <code>expectedWorkflowError</code> branch lowercases both sides.</summary>

The next branch does `strings.Contains(strings.ToLower(execErr.Error()), strings.ToLower(tc.expectedWorkflowError))`, but `require.ErrorContains` here is an exact substring match. Two assertions with near-identical intent behaving differently is an easy trap for the next test author: a case difference in an SDK's error text (the non-Go workers phrase these errors themselves) fails only in the `expectedRunError` branch, with no hint as to why.

The two fields also overlap in name without any doc. `expectedRunError` means "the scenario run fails, the workflow itself succeeds"; `expectedWorkflowError` means "the workflow execution fails". That distinction is only discoverable by reading this function.

</details>

**Suggestion:** Match case-insensitively for consistency, and document the split on the struct fields (`loadgen/kitchen_sink_executor_test.go:75-76`):

```go
	require.Errorf(t, execErr, "SDK %s should fail with run error", sdk)
	require.Containsf(t, strings.ToLower(execErr.Error()), strings.ToLower(tc.expectedRunError),
		"SDK %s run error should contain '%s'", sdk, tc.expectedRunError)
```

workers/proto/kitchen_sink/kitchen_sink.proto

<details>
<summary><strong>nit</strong> — Inserting <code>expected_output</code> before the trailing note orphans that comment and strips <code>with_start_action</code>'s docs in every generated language.</summary>

The "Technically worker options should be known as well…" block used to be `with_start_action`'s trailing comment, so protoc emitted it as that field's doc. Now it sits after `expected_output` at the end of the message and is attached to nothing — visible in this PR as the deleted javadoc on `getWithStartAction` (`workers/java/io/temporal/omes/KitchenSink.java`) and the deleted `<summary>` on `WithStartAction` (`workers/dotnet/.../KitchenSink.cs`). The note was always about `TestInput` as a whole rather than `with_start_action`, so the docs loss is incidental churn rather than intended.

</details>

**Suggestion:** Move the note above `message TestInput` (where it actually applies) so field insertions stop shuffling generated docs:

```proto
// Technically worker options should be known as well. We don't have any common format for that
// and creating one feels overkill to start with. Requiring the harness to print the config at
// startup seems good enough for now.
message TestInput {
  WorkflowInput workflow_input = 1;
  ClientSequence client_sequence = 2;
  WithStartClientAction with_start_action = 3;
  // If set, the workflow's output is compared against this value after completion.
  temporal.api.common.v1.Payload expected_output = 4;
}
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -1334,7 +1416,9 @@ func TestKitchenSink(t *testing.T) {
 			if input.WorkflowInput == nil {
 				input.WorkflowInput = &WorkflowInput{}
 			}
-			input.WorkflowInput.InitialActions = append(input.WorkflowInput.InitialActions, ListActionSet(NewEmptyReturnResultAction())...)
+			if input.GetExpectedOutput() == nil {
```

<details>
<summary><strong>nit</strong> — Setting <code>ExpectedOutput</code> now silently makes the test responsible for terminating the workflow.</summary>

The comment two lines up still says "Ensure the workflow completes by appending a return action at the end", but that guarantee now only holds when `ExpectedOutput` is unset. A future case that sets `ExpectedOutput` without its own `NewReturnResultAction` gets a workflow that blocks on the signal loop until `defaultWorkflowTimeout`, surfacing as an opaque timeout rather than "you forgot the return action".

</details>

**Suggestion:** Reword the comment so the coupling is explicit, e.g.:

```go
			// Ensure the workflow completes by appending a return action at the end.
			// Cases that assert on the output must supply their own ReturnResult action,
			// since the value returned here has to match ExpectedOutput.
			if input.GetExpectedOutput() == nil {
```

THardy98 (approved):
thanks for doing this across all workers!

Not blocking, but I believe there's an identity converter in the code somewhere (i.e payloadA -> identity converter just returns ... -> payloadA) that should no longer be necessary.

If so, would you mind filing an issue to remove it?

mavemuri (approved):

PR #484, opened 2026-09-10
Add explicit kitchen sink update IDs

stephanos:
## Why?

Without this, we cannot test a Workflow Update retry; that requires the same update ID.


workers/proto/kitchen_sink/kitchen_sink.proto
```diff
@@ -140,6 +140,8 @@ message DoUpdate {
   }
   // If set, the Update is an Update-with-Start.
   bool with_start = 3;
+  // If set, use this Update ID instead of an SDK-generated ID.
+  string update_id = 4;
```
stephanos, line 144: This is the main change.

workers/java/io/temporal/omes/workerlib/kitchensink/ClientActionExecutor.java
```diff
@@ -139,6 +140,12 @@ private void executeUpdateAction(KitchenSink.DoUpdate update) {
     }
 
     try {
+      UpdateOptions.Builder<KitchenSink.WorkflowState> updateOptions =
+          UpdateOptions.newBuilder(KitchenSink.WorkflowState.class)
+              .setUpdateName(updateName)
+              .setWaitForStage(WorkflowUpdateStage.COMPLETED)
+              .setUpdateId(update.getUpdateId());
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — Java worker sets the update ID to the proto's empty-string default instead of leaving it unset, which likely sends an empty update ID for every update that doesn't specify one.</summary>

`update.getUpdateId()` is `""` for every existing kitchen sink update (proto3 scalar default), so this unconditionally calls `setUpdateId("")`. The Java SDK's documented fallback applies only when the option is *not set*, and "not set" means `null`, not empty:

- `UpdateOptions.getUpdateId()` is annotated `@Nullable`, and its javadoc reads *"Default value if not set: Random UUID"*.
- `UpdateOptions.Builder.build()` passes `updateId` through verbatim — only `firstExecutionRunId` gets an empty-string default.
- `WorkflowClientCallsInterceptor.StartUpdateInput`'s constructor assigns `this.updateId = updateId;` with no null/empty fallback.
- `RootWorkflowClientInvoker` writes it straight into the request: `setMeta(Meta.newBuilder().setUpdateId(input.getUpdateId())...)`.

So the only remaining place a UUID can be generated is a null check in `WorkflowStubImpl`, which `""` would slip past. This is a runtime-only difference — `String` is `String`, so the compiler and CI type checks won't catch it.

Note that the other workers all guard explicitly, which makes Java (and Python, see the other comment) the odd ones out: .NET uses `string.IsNullOrEmpty` (`ClientActionsExecutor.cs:182,196`), Ruby uses `update_id.empty?` (`client_action_executor.rb:129`), and TypeScript uses `update.updateId || undefined` (`client-action-executor.ts:154,160`). Go is safe because the Go SDK explicitly treats an empty `UpdateID` as unset.

</details>

**Suggestion:** Map the empty default back to "unset", matching the `@Nullable` contract:

```suggestion
              .setUpdateId(update.getUpdateId().isEmpty() ? null : update.getUpdateId());
```

workers/python/client_action_executor.py
```diff
@@ -126,12 +126,15 @@ async def _execute_update_action(self, update: DoUpdate):
                     update_name,
                     update_args,
                     start_workflow_operation=start_op,
+                    id=update.update_id,
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — Python worker forwards the proto's empty-string default as the update ID rather than <code>None</code>, relying on falsy-coercion the API signature doesn't promise.</summary>

`update.update_id` is `""` whenever a scenario doesn't set one, which is every existing kitchen sink update. Both call sites forward that empty string: here and at line 136 (`handle.execute_update(..., id=update.update_id)`).

The SDK declares this parameter as `id: Optional[str] = None` and documents the default as a freshly generated UUID — i.e. the sentinel for "generate one" is `None`, not `""`. Whether `""` happens to be coerced is an implementation detail of the SDK's internal defaulting, not part of its contract, and `str` vs `str` means neither mypy nor CI will flag it if that behavior changes.

The other workers avoid the question entirely: .NET checks `string.IsNullOrEmpty` (`ClientActionsExecutor.cs:182,196`), Ruby checks `update_id.empty?` (`client_action_executor.rb:129`), TypeScript uses `update.updateId || undefined` (`client-action-executor.ts:154,160`). Only Java has the same gap (commented separately).

</details>

**Suggestion:** Normalize the empty default to `None` at both call sites:

```suggestion
                    id=update.update_id or None,
```

workers/go/workerlib/kitchensink/kitchen_sink.go
```diff
@@ -733,12 +733,12 @@ func updateWorkflowNexusOperation(
 		return result, nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "%s", err.Error())
 	}
 
-	// UpdateID is deliberately left unset: StartUpdateWorkflow derives it from the
-	// Nexus request ID, so a retried Nexus task attaches to the original update
-	// rather than starting a second one.
 	return temporalnexus.StartUpdateWorkflow[*common.Payload](ctx, nc, client.UpdateWorkflowOptions{
```

<details>
<summary><strong>small</strong> — Reworded comment now says <code>StartUpdateWorkflow</code> "derives an empty UpdateID from the Nexus request ID", which states the opposite of what happens.

</summary>

The original comment explained why `UpdateID` was omitted. Now that it is populated from the proto, the rewrite reads as if the derived ID is empty. The actual behavior is the inverse: the request-ID-derived ID is used *because* `UpdateID` is empty, and the retry-idempotency property described here only holds when a scenario leaves `update_id` unset.

It's also worth stating the consequence of the new opt-in path, since it's the whole point of the field: a scenario that pins `update_id` gets ID-based idempotency instead of request-ID-based, so two distinct invocations of this operation against the same workflow will dedupe into one update and the second will silently return the first result.

</details>

**Suggestion:**

```go
	// When UpdateID is empty, StartUpdateWorkflow derives it from the Nexus request
	// ID, so a retried Nexus task attaches to the original update. A scenario that
	// pins update_id opts into ID-based idempotency instead: two invocations against
	// the same workflow then dedupe into a single update.
```

loadgen/kitchen_sink_executor_test.go
```diff
@@ -418,7 +419,7 @@ func TestKitchenSink(t *testing.T) {
 			historyMatcher: PartialHistoryMatcher(`
 				ActivityTaskScheduled {"activityType":{"name":"client"}}
 				...
-				WorkflowExecutionUpdateAccepted {"acceptedRequest":{"input":{"name":"do_actions_update"}}}`),
+				WorkflowExecutionUpdateAccepted {"acceptedRequest":{"meta":{"updateId":"exec-activity-client-update"},"input":{"name":"do_actions_update"}}}`),
```

<details>
<summary><strong>small</strong> — Only the plain client-update path asserts the new update ID; the update-with-start and Nexus-update paths carry the same new plumbing untested.</summary>

`update_id` now flows through three distinct Go call sites, but this is the only case that asserts it reaches the server:

- `client.UpdateWorkflow` — covered here.
- `client.UpdateWithStartWorkflow` (`loadgen/kitchensink/client_action_executor.go:175`) — `ExecActivity/Client/Update/WithStart` (line 466) leaves `UpdateId` unset and only matches `WorkflowExecutionUpdateCompleted`.
- `temporalnexus.StartUpdateWorkflow` (`workers/go/workerlib/kitchensink/kitchen_sink.go:741`) — the Nexus update case (line 1287) leaves it unset too.

Both existing cases already assert on update history events, so extending them is a two-line change each and would pin down the paths where an explicit ID actually matters — the Nexus one especially, since setting `UpdateID` there changes which idempotency key the operation uses.

</details>

**Suggestion:** Set a distinct `UpdateId` on the `ExecActivity/Client/Update/WithStart` case and assert `{"acceptedRequest":{"meta":{"updateId":...}}}` in its matcher; do the same for the Nexus update case.

loadgen/kitchen-sink-gen/src/main.rs
```diff
@@ -557,6 +557,7 @@ impl<'a> Arbitrary<'a> for DoUpdate {
             variant: Some(variant),
             failure_expected,
             with_start: u.arbitrary()?,
+            update_id: String::new(),
```
stephanos: 🤖

<details>
<summary><strong>nit</strong> — Hardcoded empty <code>update_id</code> is the only non-arbitrary field in this <code>Arbitrary</code> impl and carries no explanation.</summary>

Every other field here is fuzzer-driven (`variant`, `failure_expected`, `with_start`), so a bare `String::new()` reads as an oversight rather than a decision. Leaving it empty is the right default — arbitrary IDs would collide across generated updates and dedupe them — but that reasoning isn't recoverable from the code.

</details>

**Suggestion:**

```suggestion
            // Left empty so each generated update gets a fresh SDK-generated ID;
            // arbitrary IDs could collide and silently dedupe updates.
            update_id: String::new(),
```

workers/java/io/temporal/omes/workerlib/kitchensink/ClientActionExecutor.java
```diff
@@ -139,6 +140,12 @@ private void executeUpdateAction(KitchenSink.DoUpdate update) {
     }
 
     try {
+      UpdateOptions.Builder<KitchenSink.WorkflowState> updateOptions =
+          UpdateOptions.newBuilder(KitchenSink.WorkflowState.class)
+              .setUpdateName(updateName)
+              .setWaitForStage(WorkflowUpdateStage.COMPLETED)
```
stephanos, line 146: This is explicitly specified now as we're using `stub.startUpdate` now.

mavemuri (approved):
LGTM!

THardy98 (approved):

PR #494, opened 2026-09-17
Read the mTLS credentials from a single combined PEM

gcristea-temporal:
## What was changed

Added ability for serverless workers to read the client key and certificate as one secret. That secret is pushed from cert-manager's `tls-combined.pem` to AWS/GCP. That means both TLS loading paths (lambda & regular) need to parse a blob holding both halves.

- Lambda: `TLS_COMBINED`, one `GetSecretValue` calls. `TLS_CERT` and `TLS_KEY` still work, for workers already deployed with them.
- CLI and worker: `--tls-combined-path`, for a GCP Cloud Run mount.
- Misconfigurations now name themselves instead of failing later as a parse or handshake error: the two forms cannot be mixed, and a half-set pair is rejected rather than silently skipping mTLS.

## Why?

The External Secrets Operator (ESO) is used to push the secrets from k8s to AWS/GCP SecretsManager. If there are two secrets (key & cert) and the push for one of them fails, the reconcile stops there. That results in a state mismatch that is not surfaced in k8s. The next cold start then fails in `tls.X509KeyPair` with "private key does not match public key", and stays broken until a subsequent reconciliation succeeds.
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


clioptions/client.go
```diff
@@ -51,20 +53,45 @@ type ClientOptions struct {
 	fs *pflag.FlagSet
 }
 
+// X509KeyPairFromCombinedPEM parses cert-manager's tls-combined.pem. X509KeyPair
```
THardy98, line 56: `parses cert-manager's tls-combined.pem`
there is no relation between this comment the function below
in any case, the comment is describing specific usage of a general function

THardy98 (approved):
left a nit on a comment, but otherwise, lgtm

